PR #814, opened 2014-08-27
Pull Request for Issue631: SGM Visible Light Valves

davidChevrier:
Update to how visible light movement works on SGM. Requires beamline testing.


davidChevrier:
With the SGM going down for a major upgrade starting in a couple of weeks I'm going to close this. We're going to do such a major overhaul that I think we'll likely need to start a lot of things from scratch.

PR #925, opened 2014-10-24
Pull Request for Issue912: Add the wire stage motors to VESPERSSpatialLineScan and Add big beam stages to new VESPERS application

dretrex:
Added the motors to VESPERS application. They work as the did before but now in the new scheme. I have tested all the motors and the wire stage does not behave well, but that is more due to tuning more than anything.


davidChevrier:
I removed the fact this also handles Issue899 from the title to get around a problem. All of the time is accounted for.

Issue #1110, opened 2014-12-22
SXRMB: Acquaman crashed when change "Multiple Spectra" for Bruker detector


XuetaoLiu-CLS:
Acquaman on SXRMB crashed consistently when changing (unchecking "Raw Spectrum") of "Show Multiple Spectrum" window.

PR #1112, opened 2014-12-23
Pull Request for Issue1110: add flag for AMSelectionDialog to control whether allows empty selected items

XuetaoLiu-CLS:
-- Added: emptySelectedItemsAllowed feature for AMSelectionDialog. With this flag, if empty selected items is not allowed and no items are selected, then the Okay button will be greyed out. 
-- add compile setting to AMCommon.pri, so that all the files (.obj, moc_, ui_) generated by compiler will be stored in corresponding folders instead of the root folder

https://github.com/acquaman/acquaman/issues/1110

@dretrex @davidChevrier

### DIFF
--- a/acquaman_suite.pro
+++ b/acquaman_suite.pro
@@ -7,34 +7,34 @@ SUBDIRS += \
 	Initialize.pro \
 	pluginProjects/FileLoaders/AMFileLoaderPlugins.pro \
 	# pluginProjects/AnalysisBlocks/AMAnalysisBlockPlugins.pro \
-	REIXSTest.pro \
-	REIXSAcquaman.pro \
-	SGMAcquaman.pro \
-	BareBonesAcquaman.pro \
-	acquamanTest.pro \
-	VESPERSAcquaman.pro \
-	MidIRBPM.pro \
-	CLSPGTDwellTimeCoordinator.pro \
-	SGMAddOnsCoordinator.pro \
-	SGMSSAAcquaman.pro \
 	EnsureHeaderNotice.pro \
-	VESPERSBendingMirrors.pro \
-	SGMAmptekCoordinator.pro \
-	SGMLookupTableCoordinator.pro \
 	AMCrashReporter.pro \
-	StripToolProject.pro \
-	IDEASAcquaman.pro \
-	BioXASSideAcquaman.pro \
+	CLSNetworkDirectorySynchronizer.pro \
 	BioXASMainAcquaman.pro \
+	BioXASSideAcquaman.pro \
 	BioXASImagingAcquaman.pro \
 	BioXASToolSuite.pro \
-	VESPERSDatabaseDuplicateEntryPatch.pro \
-	StripTool2.pro \
-	CLSNetworkDirectorySynchronizer.pro \
-	buildTests/AMBuildTest.pro \
-	XRDAnalysis.pro \
-	CLSSR570Coordinator.pro \
+	MidIRBPM.pro \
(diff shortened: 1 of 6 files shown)


dretrex:
@XuetaoLiu-CLS, what was wrong with the AMCommon.pri stuff? I see you reverted all those changes.

As to the selection dialog, I would say if you want to make it so you can enable selections with no items, then you should also make it so you can disable it later.  Conversely, I think the better option is just not allow someone to hit "Okay" if you there have been no items selected.

XuetaoLiu-CLS:
@dretrex with that changes, it works when I build with a single beamline project, such as SXRMB, BioXAS. But it failed when building the acquaman_suite.pro. I'm trying to figure out the root cause of this issue.

Regarding the selection dialog, I believe that "enable/disable selection with no items" will only happen when you create the dialog, since this should be a nature of the dialog we are going to create. There should be no requirement of changing  "enable/disable selection with no items" after a selection dialog is created.

I did a search when I made the changes. So far this is the only place we are using the selection dialog. Originally I also planned to change the codes as you suggested. But after a second thought, in the future, we may have the usage of the "no items selection". That's why I added this flag, but set it to "false" by default.

dretrex:
Even if it's only when creating the dialog, if you are going to provide an API to change the state of it in some way, you should implement the way to change it back.  Can you please add a "disable selection with no items" to the API for the class?

davidChevrier:
@XuetaoLiu-CLS what was the problem with the build on the suite? I built the whole suite on my mac and it was good to go. I'll check what happens if I build it on my OPI tomorrow.

XuetaoLiu-CLS:
@davidChevrier there are some linking errors.

@dretrex the API is changed to `setEmptySelectedItemsAllowed(bool allow)`

dretrex:
Looks good as it is now.  Will wait on @davidChevrier to merge and close it out.

XuetaoLiu-CLS:
@davidChevrier the building issue is solved now.

All compiled file will be under folder build_files

XuetaoLiu-CLS:
@dretrex I noticed that the -r will make the script failed for the first time. We have the run the script twice to get a success build.

XuetaoLiu-CLS:
@dretrex I just took a quick look at the qmake manual. maybe we should also add a `-project`?

```
http://qt-project.org/doc/qt-4.8/qmake-running.html

Project Mode Options

 qmake -project [options] files

In project mode, qmake will generate a project file. Additionally, you may supply the following options in this mode:

    -r
    qmake will look through supplied directories recursively
    -nopwd
    qmake will not look in your current working directory for source code and only use the specified files
```

XuetaoLiu-CLS:
@dretrex the `-r` is only mentioned in the project mode, which will generate a single project file instead of make file. It doesn't mention (or did I miss it?) the usage of `-r` in makefile mode.

http://qt-project.org/doc/qt-4.8/qmake-running.html

davidChevrier:
Sadly the qmake documentation is not good. You can run 
qmake acquaman_suite.pro
vs
qmake -r acquaman_suite.pro
And see the difference even at the command line level.

davidChevrier:
Hey @XuetaoLiu-CLS 

All of this build stuff seems a little confusing. I'm down at SXRMB at the moment. Do you want to stop by so we can discuss?

XuetaoLiu-CLS:
@davidChevrier @dretrex 
Dave is correct. the failure is caused by the `make -j 3` instead of 'qmake -r'. This looks good to go now.

Thanks.

PR #1135, opened 2015-01-12
Pull Request for Issue1113: AMExporter File Names

davidChevrier:
Fix to make sure that exported file names don't have non-printable characters in them.

### DIFF
--- a/source/dataman/export/AMExporter.h
+++ b/source/dataman/export/AMExporter.h
@@ -200,6 +200,9 @@ protected:
 
 	QString krFileSystemAutoIncrement(const QString &arg = QString());
 
+	/// Helper function to remove non-printable characters from a file name
+	QString removeNonPrintableCharacters(const QString &fileName);
+
 };
 
 #endif // AMEXPORTER_H
--- a/source/dataman/export/AMExporter.h
+++ b/source/dataman/export/AMExporter.h
@@ -201,7 +201,7 @@ protected:
 	QString krFileSystemAutoIncrement(const QString &arg = QString());
 
 	/// Helper function to remove non-printable characters from a file name
-	QString removeNonPrintableCharacters(const QString &fileName);
+	QString removeNonPrintableCharacters(const QString &fileName) const;
 
 };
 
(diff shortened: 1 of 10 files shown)


source/dataman/export/AMExporterGeneralAscii.cpp
```diff
@@ -93,6 +93,20 @@ QString AMExporterGeneralAscii::exportScan(const AMScan *scan, const QString &de
 
 	// prepare export file
 	mainFileName_ = parseKeywordString( destinationFolderPath % "/" % option->fileName() );
+	mainFileName_ = removeNonPrintableCharacters(mainFileName_);
+
+//	bool detectedNonprintable = false;
```
dretrex: Unnecessary comments.  Probably should remove these.

source/dataman/export/AMExporter.h
```diff
@@ -200,6 +200,9 @@ public slots:
 
 	QString krFileSystemAutoIncrement(const QString &arg = QString());
 
+	/// Helper function to remove non-printable characters from a file name
+	QString removeNonPrintableCharacters(const QString &fileName);
```
dretrex: I think technically this should be a const function.

source/dataman/export/AMExporterGeneralAscii.cpp
```diff
@@ -516,6 +530,19 @@ bool AMExporterGeneralAscii::writeSeparateFiles(const QString& destinationFolder
 
 			QFile file;
 			QString separateFileName = parseKeywordString( destinationFolderPath % "/" % option_->separateSectionFileName() );
+			separateFileName = removeNonPrintableCharacters(separateFileName);
+//			bool detectedNonprintable = false;
```
dretrex: Commented code.

source/dataman/export/AMExporterGeneralAscii.cpp
```diff
@@ -599,6 +626,22 @@ bool AMExporterGeneralAscii::writeSeparateFiles(const QString& destinationFolder
 			QFile file;
 			QString separateFileName = parseKeywordString( destinationFolderPath % "/" % option_->separateSectionFileName() );
 
+			bool detectedNonprintable = false;
```
dretrex: Is there a reason why you're not just using the function you made here?  They look pretty similar to me.

davidChevrier: Thanks @dretrex good catch. I found a couple places I missed because of this.

davidChevrier:
Time Estimate: 8 hours
(with diagnostics and testing)

PR #1141, opened 2015-01-13
Pull Request for Issue1040: AMControlWaitAction Crash

davidChevrier:
Altered the AMControl[Move/Stop/Wait]ActionInfo classes to return the AMControlInfo by value rather than a const pointer. Tested on VESPERS and doesn't appear to introduce any problems.

### DIFF
--- a/source/actions3/actions/AMControlMoveActionInfo3.h
+++ b/source/actions3/actions/AMControlMoveActionInfo3.h
@@ -54,7 +54,8 @@ public:
 	//////////////////////////
 
 	/// Returns a pointer to our move destination setpoint
-	const AMControlInfo* controlInfo() const { return &controlInfo_; }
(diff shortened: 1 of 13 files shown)


source/actions3/actions/AMControlMoveActionInfo3.h
```diff
@@ -54,7 +54,7 @@ class AMControlMoveActionInfo3 : public AMActionInfo3
 	//////////////////////////
 
 	/// Returns a pointer to our move destination setpoint
-	const AMControlInfo* controlInfo() const { return &controlInfo_; }
+	AMControlInfo controlInfo() const { return controlInfo_; }
```
dretrex: Since this always a valid piece of memory, you could potentially make it a ```const AMControlInfo &```.  It's not a big deal though.

PR #1144, opened 2015-01-14
Pull Request for Issue1142: Add Internal Retry to AMControlMoveAction

davidChevrier:
Added and tested on VESPERS. Appears to work well. Default retries is set to 1.

### DIFF
--- a/source/actions3/actions/AMControlMoveAction3.cpp
+++ b/source/actions3/actions/AMControlMoveAction3.cpp
@@ -25,7 +25,6 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 #include "util/AMErrorMonitor.h"
 #include "acquaman/AMAgnosticDataAPI.h"
 
- AMControlMoveAction3::~AMControlMoveAction3(){}
 AMControlMoveAction3::AMControlMoveAction3(AMControlMoveActionInfo3 *info, AMControl *control, QObject *parent)
 	: AMAction3(info, parent)
 {
@@ -33,6 +32,10 @@ AMControlMoveAction3::AMControlMoveAction3(AMControlMoveActionInfo3 *info, AMCon
 		control_ = control;
 	else
 		control_ = AMBeamline::bl()->exposedControlByInfo(*(info->controlInfo()));
+
+	retries_ = 1;
+	attempts_ = 0;
+	destination_ = 0;
 }
 
 AMControlMoveAction3::AMControlMoveAction3(const AMControlMoveAction3 &other)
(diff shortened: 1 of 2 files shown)


source/actions3/actions/AMControlMoveAction3.cpp
```diff
@@ -44,6 +47,32 @@ AMControlMoveAction3::AMControlMoveAction3(const AMControlMoveAction3 &other)
 		control_ = AMBeamline::bl()->exposedControlByInfo(*(info->controlInfo()));
 	else
 		control_ = 0;
+
+	retries_ = 1;
```
dretrex: Do we not want to set the number of retries to the ```other.retries()```?  It is the copy constructor.

source/actions3/actions/AMControlMoveAction3.cpp
```diff
@@ -85,15 +114,21 @@ void AMControlMoveAction3::startImplementation()
...
 	startPosition_ = control_->toInfo();
 
-	// start the move:
-	int failureExplanation;
+	attempts_ = 1;
+
+	// determine the absolute destination
 	if(controlMoveInfo()->isRelativeMove() && controlMoveInfo()->isRelativeFromSetpoint())
-		failureExplanation = control_->moveRelative(setpoint.value(), AMControl::RelativeMoveFromSetpoint);
+		destination_ = control_->setpoint() + setpoint.value();
```
dretrex: You did this change after I left, right?  Does this still work properly if a control move action was inside of a loop doing a relative move 10 times?

source/actions3/actions/AMControlMoveAction3.cpp
```diff
@@ -151,10 +188,32 @@ void AMControlMoveAction3::onMoveFailed(int reason)
...
+	QString fundamentalFailureMessage = QString("There was an error moving the control '%1' into position (Retry %2, attempt %3 of %4). Target: %5 %6. Actual position: %7 %8. Tolerance %9. Reason: %10.").arg(control_->name()).arg(retryAvailableString).arg(attempts_).arg(retries_+1).arg(control_->setpoint()).arg(control_->units()).arg(control_->value()).arg(control_->units()).arg(control_->tolerance()).arg(AMControl::failureExplanation(reason));
 	// error message with reason
 	AMErrorMon::alert(this, definedFailureReason, QString("%1. Please report this problem to the beamline staff.").arg(fundamentalFailureMessage));
-	setFailed(fundamentalFailureMessage);
+
+	if(retryAvailable){
+		attempts_++;
+
+		// retry move
+		qDebug() << "Retrying move to " << destination_;
```
dretrex: Do you still want this qDebug() there?  Do you want to have an error monitor instead or is the error mon stuff handled entirely by the above error mon?

source/actions3/actions/AMControlMoveAction3.cpp
davidChevrier: Clever @dretrex

source/actions3/actions/AMControlMoveAction3.cpp
davidChevrier: Nope, this is what we tested yesterday. Seemed to work correctly as far as we could tell.

source/actions3/actions/AMControlMoveAction3.cpp
davidChevrier: Gone.

Issue #1131, opened 2015-01-09
SXRMB EXAFS Scan Failure


davidChevrier:
A user at SXRMB had a scan fail to cancel today. The reported error code was 262004. The user reported that the scan finished collecting data and then went into the clean up mode. The clean up seemed to hang and she left it alone for over 10 minutes. When she tried to cancel the error popped up.

I had to use the "Force Quit Acquaman" from the File menu so we could restart the program.

The only items in the SXRMBEXAFSScanActionController::createCleanupActions() call are dealing with the scaler. However, these are all controlMoveActions, so I really think they should have timed out. I'll try to look at the scanActions database to see if I can find any clues and I'll pass them along to @XuetaoLiu-CLS

davidChevrier:
A second failure in a row appears to shed new light on the problem. @XuetaoLiu-CLS could you give a quick report on what you saw when you were down there?

XuetaoLiu-CLS:
After the second failure, the computer is locked. We couldn't login as it keeps saying that the session is too short (10s) due to system failure or no disk space. We had to reboot the computer. After computer started, we were able to login as "SXRMB".

Both hanged scans have empty data files (all the user data files and the exported data files). 

Then we tried to rerun the scan without adding to workflow and it works. 

(maybe sth wrong with workflow? Will try it on Monday)

davidChevrier:
I really doubt it has anything to do with the workflow ... or AMActionRunner::workflow() anyway.

Whether you click the "Add to Workflow" button to the "Start Scan" button, the scan will be run in the workflow in either case.

dretrex:
Also, it doesn't seem weird to me that they could re-run the scan.  The scan meta data is in the user database, not in the generated ascii files.  To me, this seems like something eats file handles, never gives them back and if that OPI has been running for months without being rebooted, all kinds of quirky behaviour starts.  The only thing that is troubling to me is that we didn't seem to stop on file open failure.  Unless the computer was so messed up that it thought it had opened the file but couldn't do anything with the stream.

davidChevrier:
Options for checking on the file size are:
In the QFile itself using the fileWriter
Using QFileInfo to check on that file
Using a QProcess to check the output of an "ls" command

davidChevrier:
@XuetaoLiu-CLS Time Estimate?

XuetaoLiu-CLS:
4 hours?

PR #1145, opened 2015-01-14
Pull Request for Issue1131: Add protection when we failed to write to file:

XuetaoLiu-CLS:
1) check file size at a 2 mins interval. If size is not changed, raise FileWriteError to stop the scan
3) handle the new error message.

https://github.com/acquaman/acquaman/issues/1131

@dretrex @davidChevrier


source/acquaman/AMScanActionControllerBasicFileWriter.cpp
```diff
@@ -78,8 +78,17 @@ void AMScanActionControllerBasicFileWriter::writeToFile(int fileRank, const QStr
 		qDebug() << "Stream (1) writing differential " << startTime.msecsTo(endTime) << " for size " << textToWrite.size();
 		*/
 
+		QFileInfo dataFileInfo(filePath_+".dat");
+		int sizeBeforeWrite = dataFileInfo.size();
+
 		AMTextStream dataStream(dataFile_);
 		dataStream << textToWrite;
+		dataStream.flush();
```
XuetaoLiu-CLS: In QT documentation, it said:
```
void QTextStream::flush ()

Flushes any buffered data waiting to be written to the device.
If QTextStream operates on a string, this function does nothing.
```

Actually, QTextStream won't write to device, if `flush()` is not called. Therefore, in order to compare the size of the file, we have to force `flush`.

source/acquaman/AMTimedScanActionController.h
```diff
@@ -38,6 +38,7 @@ Q_DECLARE_METATYPE(AMScanActionControllerBasicFileWriter::FileWriterError)
 #define AMTIMEDSCANACTIONCONTROLLER_FILE_ALREADY_EXISTS 265003
 #define AMTIMEDSCANACTIONCONTROLLER_COULD_NOT_OPEN_FILE 265004
 #define AMTIMEDSCANACTIONCONTROLLER_UNKNOWN_FILE_ERROR 265005
+#define AMTIMEDSCANACTIONCONTROLLER_FAILED_TO_WRITE_FILE 265006
```
XuetaoLiu-CLS: @davidChevrier @dretrex 
I have a question here. We use the same error codes in `AMStepScanActionController`, `AMTimedScanActionController`, and `REIXSXESScanActionController`. But `SGMFastScanActionController` has different codes. 

I think we should use different codes. Thought?

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
dretrex: My understanding is the exact opposite.  Since QTextStream is specifically for handling strings, it automatically flushes the data to disk.  Hence why it says that calls to ```flush()``` are ignored.  Since you call ```QFileInfo::refresh()``` below, this shouldn't be necessary.

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
```diff
@@ -95,8 +104,17 @@ void AMScanActionControllerBasicFileWriter::writeToFile(int fileRank, const QStr
 			qDebug() << "Stream (2) writing differential " << startTime.msecsTo(endTime) << " for size " << textToWrite.size();
 			*/
 
+			QFileInfo spectraDataFileInfo(filePath_+"_spectra.dat");
+			int sizeBeforeWrite = spectraDataFileInfo.size();
+
 			AMTextStream spectraStream(spectraFile_);
 			spectraStream << textToWrite;
+			spectraStream.flush();
```
dretrex: See above discussion.

source/acquaman/AMStepScanActionController.cpp
```diff
@@ -204,6 +204,11 @@ void AMStepScanActionController::onFileWriterError(AMScanActionControllerBasicFi
 		userErrorString = "Your scan has been aborted because Acquaman was unable to open the desired file for writing (for internal storage). This is a serious problem and would have resulted in collecting data but not saving it. Please contact the Acquaman developers immediately.";
 		break;
 
+	case AMScanActionControllerBasicFileWriter::FailedToWriteFile:
+		AMErrorMon::alert(this, AMSTEPSCANACTIONCONTROLLER_FAILE_TO_WRITE_FILE, QString("Error, the %1 Scan Action Controller failed to write your data. This is a serious problem, please contact the Acquaman developers.").arg(stepConfiguration_->technique()));
```
dretrex: Honestly, I would say that this should be ```AMErrorMon::critical/serious()``` (I can't remember off the top of my head).  If data isn't being saved, users should stop right away and seek help.

source/acquaman/AMTimedScanActionController.h
dretrex: All of them should be different.  You are correct.  That is likely just laziness in copy paste that was never remedied.

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
davidChevrier: I think @XuetaoLiu-CLS is "right" here. The device in question is a QFile (dataFile_), so I think that the AMTextStream (QTextStream) is "operating" on that. Being said, it's plain that under normal circumstances we are writing things to the file, so we don't "need" to call flush() to write anything. Rather, it's most likely that calling flush() forces the data to be flushed so that when @XuetaoLiu-CLS is checking before and after he gets a different number. I imagine that under normal circumstances, the OS decides when to actually flush the device optimally (I hope that we're not going to significantly decrease performance by doing it this way). @dretrex @XuetaoLiu-CLS @iainworkman can you guys think of a way of refactoring this that doesn't rely on flushing the buffer in line?

source/acquaman/AMTimedScanActionController.h
davidChevrier: I agree, unique error codes have always been helpful.

source/acquaman/AMStepScanActionController.cpp
davidChevrier: @dretrex  is right, this is a cataclysmic problem. We need to escalate it as serious and actually have it cause scan failure.

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
dretrex: I still completely disagree.  The function says it does nothing in the only case we use it for.

source/acquaman/AMStepScanActionController.cpp
XuetaoLiu-CLS: I thought `AMErrorMon::alert` will abort the scan as well. At lease the other error handlers mentioned the scan `has been aborted`.

I changed to `AMErrorMon::serious` now. Will read the codes to figure out the difference. :)

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
```diff
@@ -120,6 +127,25 @@ void AMScanActionControllerBasicFileWriter::finishWriting()
 	emit fileWriterIsBusy(false);
 }
 
+void AMScanActionControllerBasicFileWriter::onDataFileSizeCheckTimerTimeout()
+{
+	QFileInfo dataFileInfo(filePath_+".dat");
+	if (dataFileInfo.size() == dataFileSizeCheckPoint_) {
```
XuetaoLiu-CLS: I assume that `writeToFile` will be called more than once in 2 mins.

davidChevrier:
Hey @XuetaoLiu-CLS 
I'd like to see a check to make sure that in the 2 minute window we have added at least one write to the file. I'm pretty sure I know how to do that easily, do you want me to code it up and send it your way for testing?

XuetaoLiu-CLS:
@davidChevrier I thought of adding a flag for writing operation in those two minutes window. If you have the codes ready, please go ahead.

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
```diff
@@ -62,6 +65,12 @@ AMScanActionControllerBasicFileWriter::AMScanActionControllerBasicFileWriter(con
 		if(!spectraFile_->open(QIODevice::WriteOnly | QIODevice::Text))
 			emitError(AMScanActionControllerBasicFileWriter::CouldNotOpenError);
 	}
+
+	// start the timer (2mins) to check whether we are writing to the files consistently
+	QTimer *dataFileSizeCheckTimer = new QTimer(this);
```
dretrex: This is a memory leak.

source/acquaman/AMStepScanActionController.cpp
```diff
@@ -204,6 +204,11 @@ void AMStepScanActionController::onFileWriterError(AMScanActionControllerBasicFi
 		userErrorString = "Your scan has been aborted because Acquaman was unable to open the desired file for writing (for internal storage). This is a serious problem and would have resulted in collecting data but not saving it. Please contact the Acquaman developers immediately.";
 		break;
 
+	case AMScanActionControllerBasicFileWriter::FailedToWriteFile:
```
dretrex: Should this also start setting things up to fail the entire scan?  If this happens, it is really serious.

davidChevrier: The set failed is right after this code block.

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
XuetaoLiu-CLS: @dretrex  The `dateFileSizeCheckTimer` is a child of `this`. Shouldn't it be destroyed by the parent's destructor? I have thought that in QT, QObject destructor will destroy its children and itself.

davidChevrier: I just checked to make sure, the file writer is also explicitly destroyed by calling deleteLater() some where on it. That was the file writer fix during the shutdown.

XuetaoLiu-CLS: @davidChevrier @dretrex Yes, we need to call deleteLater() for the fileWriter since it is NOT a child of the parent class. Therefore, it has to be destroyed explicitly.

Maybe I'm wrong. But QT documented that QObject can destroy its child automatically. That's the reason I use `new QTimer(this)` instead of `new QTimer()` here. I'm expecting the super class can destroy the timer for me.

davidChevrier:
Just thought that we likely need to stop that timer when we're done writing.

XuetaoLiu-CLS:
@davidChevrier good point. New changes will be pushed soon.

source/acquaman/AMScanActionControllerBasicFileWriter.cpp
dretrex: Yes, you are correct, @XuetaoLiu-CLS.  That was my mistake.  I didn't see the ```this``` pointer.

dretrex: You are also correct with your assumption of how QObjects are destroyed.

XuetaoLiu-CLS:
@davidChevrier @dretrex 

Test pass.
1) when no data written to the file, the scan will be stopped
2) every 2 mins, we will check the file status.

davidChevrier:
@XuetaoLiu-CLS @dretrex 
Is there any more testing to be done here? Is this something we want merged in now?

XuetaoLiu-CLS:
@davidChevrier The tested is done as I commented above. If @dretrex has no obligation, we can merge it .

dretrex:
I have no objections.

Issue #1149, opened 2015-01-15
Acquaman run script - make EnableCrashReport not mandantory


XuetaoLiu-CLS:
AMCrashReport caused bad behavior on SXRMB beamline. Therefore, we hope to make the enableCrashMonitor optional.

remove the --enableCrashMonitor option from the runAcquaman.sh script. Instead the beamline acquaman script should pass the option to runAcquamsh.sh

XuetaoLiu-CLS:
Time Estimate: 15mins

PR #1150, opened 2015-01-15
Pull Request for Issue1149: remove the --enableCrashMonitor option from the runAcquaman.sh script.

XuetaoLiu-CLS:
- removed the `--enableCrashMonitor` from the runAcquaman.sh script
- after this is merged in, each beamline should modify its own run script to pass the option to the runAcquaman.sh

https://github.com/acquaman/acquaman/issues/1149

@davidChevrier @dretrex


dretrex:
Can you make it so that you could pass an argument to the script for the people who use the crash reporter?  That way the boot script is the same for everyone and only the local launch script needs to be different.

XuetaoLiu-CLS:
@dretrex 
When using the `runAcquaman.sh` script, the option can be passed in as `runAcquaman.sh ./build/SXRMBAcquaman --enableCrashMonitor`. This will enable the `AMCrashReporter`.

dretrex:
Alright.

Issue #1151, opened 2015-01-19
BioXAS pseudo motor tool writes to motors on close


beaured0:
Sometimes when closing windows on the BioXAS pseudo motor tool very large values are written to the setpoint pvs sending motors to their limits.

The tool suite has been disabled until this is resolved.

davidChevrier:
Hey @beaured0 @XuetaoLiu-CLS 

What's the stuff in CLSPseudoMotorControl about "delayMove"? Can you guys explain that to me?

dretrex:
See the following commit:  https://github.com/acquaman/acquaman/commit/e434f9368064c68496fe963b493d7fa21cf2daec

It was something I had to add in to get the pseudo motors to work at all inside of AM apparently.

XuetaoLiu-CLS:
My understanding is that when the real motor is  moving, the pseudo motor won't move right away if we want it to move again.

________________________________________

dretrex:
It has to do with the fact that CLS pseudo motors push on every motor in the group regardless of whether all motors are part of the move.  This then causes some issues with move statuses coming in at wrong times and causing scans to hang up.

davidChevrier:
@XuetaoLiu-CLS @dretrex @beaured0 

A quick look at that code makes me think it's the culprit. The bool delayMove_ isn't initialized in the constructor. If we got a movingChanged at any point before we do a move() or stop() ourselves, there's a chance that it would just go somewhere since delayMoveSetpoint_ isn't initialized either.

dretrex:
And that's why we have code reviews now!

dretrex:
Done in pull request #1155.  Time estimate @XuetaoLiu-CLS?

XuetaoLiu-CLS:
Finished in minutes. Tested for half an hour

PR #1155, opened 2015-01-21
Pull Request for Issue1151: Initialize the delayMove_ for CLSPseudoMotorControl,

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1151

This might cause the random motor move when receiving status change.

Tested on BioXAS beamline. Didn't find random change.

@davidChevrier @dretrex


dretrex:
Can you initialize the delayed setpoint too?

XuetaoLiu-CLS:
@dretrex I have thought of doing that. But what's the proper value we should use? I checked the codes. We will set the value of the delayed setpoint when we set delayMove_ to true. I think it might be sufficient if we initialize delayMove_ to false.

dretrex:
Just initialize it to 0.0.  I agree that it won't be used until it is properly set, but it's still good practice to initialize all values.

XuetaoLiu-CLS:
@dretrex it is initialized now.

Issue #1167, opened 2015-01-27
AMCrashReport: prefer to make crash report file name ordered by date+time


XuetaoLiu-CLS:
It will be nicer to put Date in front of time for the crash report file, which will make the crash reports easier to search.

dretrex:
This was finished in #1168.  Time estimate?

XuetaoLiu-CLS:
Took about 30 mins to finish coding and 2 hours to test it

PR #1168, opened 2015-01-27
Pull Request for Issue1167: Make the report file date in front of time

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1167

1) the reports can be easier to be found since they are ordered by date and then time.
2) add year to the subfolder of the beamline folder, so that the history reports will be organized nicely

@dretrex @davidChevrier

### DIFF
--- a/source/application/AMCrashReporter.cpp
+++ b/source/application/AMCrashReporter.cpp
@@ -287,7 +287,7 @@ void AMCrashReporter::onOneSymbolProcessed(){
 }
 
 void AMCrashReporter::onAllSymbolsProcessed(){
-	QFile reportFile(QString("%1/report_%2_%3_%4_%5.txt").arg(errorFilePath_).arg(executableFullPath_.section('/', -1)).arg(QHostInfo::localHostName()).arg(QDateTime::currentDateTime().toString("ddMMyyyy_hhmmss")).arg(watchingPID_));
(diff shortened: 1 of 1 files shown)


source/application/AMCrashReporter.cpp
```diff
@@ -118,6 +118,12 @@ AMCrashMonitor::AMCrashMonitor(const QString &executableFullPath, const QString
 	errorFilePath_ = errorFilePath;
 	watchingPID_ = watchingPID;
 
+	//check the current year directory exist or not, if not create one.
+	QString targetPath = QString("%1/%2").arg(errorFilePath_).arg(QDateTime::currentDateTime().toString("yyyy"));
+	if (!QDir(targetPath).exists()) {
+		QDir().mkdir(targetPath);
```
davidChevrier: It's possible for this operation to fail. You might want to handle that situation.

XuetaoLiu-CLS: @davidChevrier I deleted these codes and move them down.

davidChevrier:
Looks much better. Have you tested this in the wild? Just checking before I pull it into master.

XuetaoLiu-CLS:
No, I didn't test it yet since SXRMB beamline is occupied now. But I do tested the codes for directory checking and creating code in a seperate QT project. and this is the major changes actually.

davidChevrier:
Okay. Let's get this tested next Monday on SXRMB. That way we can make sure that everything is good to go before it winds up in master.

XuetaoLiu-CLS:
Test is done on SXRMB beamline.
On SXRMB beamline, the AMCrashReporter is not launched by the AMCrashMonitor automatically when crash happened by some unknown reason. This is known issue and need to be investigated.

When launching AMCrashReporter manually, it will create the year subfolder under the beamline folder and create crash file as expected.

PR #1180, opened 2015-02-03
Pull Request for Issue348 - Adding options to plot points and or lines in Scan Plots

iainworkman:
Added widgets specifically designed for the selection of Line Pen Styles and Marker Shapes. Redesigned AMLinePropertyEditor to include options for changing the marker shapes and colour. Updated AMLinePropertyEditor so it emits signals on changes to these widgets which (eventually) make their way through to the series being drawn.


source/ui/AMLinePropertyEditor.cpp
```diff
@@ -81,112 +94,78 @@ void AMLinePropertyEditor::setupUi()
...
+	areaFillCheckBox_ = new QCheckBox();
+	areaFillCheckBox_->setEnabled(false);
+	mainLayout->addRow(QString("Area Fill:"), areaFillCheckBox_);
+	mainLayout->labelForField(
+				areaFillCheckBox_)->setStyleSheet(
+					QString("QLabel { color: gray }"));
+
+	// The color to fill the area of the plot with (not yet supported in MPlot,
+	// therefore the widget is disabled
+	areaFillColorPickerButton_ = new AMColorPickerButton();
```
davidChevrier: There are about 8 new'd elements here ... do they get cleaned up anywhere?

Issue #1183, opened 2015-02-03
Memory leak for AMCrashMonitor


XuetaoLiu-CLS:
There are memory leak for `AMCrashMonitor`. When `AMCrashMonitor` is running, the RSS (Resident Set Size) memory usage is growing slowly. 

The leak is in the following codes. For each 5 second, we will create two `AMProcessWatcher` instances, which are never released.
```
void AMCrashMonitor::onWatchingPIDTimerTimeout(){
	AMProcessWatcher *checkForWatchingPID = new AMProcessWatcher(watchingPID_);
	connect(checkForWatchingPID, SIGNAL(foundWatchPID(bool)), this, SLOT(onWatchingPIDFoundProcess(bool)));
	checkForWatchingPID->startWatching();
}

void AMCrashMonitor::onCrashReporterPIDTimerTimeout(){
	AMProcessWatcher *checkForWatchingPID = new AMProcessWatcher(crashReporterPID_);
	connect(checkForWatchingPID, SIGNAL(foundWatchPID(bool)), this, SLOT(onCrashReporterPIDFoundProcess(bool)));
	checkForWatchingPID->startWatching();
}
```

XuetaoLiu-CLS:
Took about 1.5 hours to code and test

PR #1184, opened 2015-02-04
Pull Request for Issue1183: release the AMProcessWatcher objects

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1183

AMCrashMonitor created instances of AMProcessWatcher, but never release them. These instances will stay in memory until AMCrashMonitor is terminated, which will cause memory leak since AMCrashMonitor will only stop if the beamline acquaman is closed.

@dretrex @davidChevrier


XuetaoLiu-CLS:
Tested on SXRMB. The RSS of AMCrashMonitor will grow a little bit at the beginning and then stable. 

We created two AMProcessWatcher every 5 seconds and will release them when it is finished. When the rate of creating new instances and the rate of releasing old instances are equal, the RSS will be stable.

PR #1188, opened 2015-02-09
Pull Request for Issue1178: Prototype the 32 element Ge detector on BioXAS

dretrex:
After an overall successful week, this is all the changes that were made for the 32 element Ge detector.  In an AM sense, I didn't get as far as I'd have liked, but the progress we did make allowed us to diagnose some potentially major issues with the detector/electronics.  The only thing left undone that should really be done is to get a delayed acquisition finished for the detector to wait on all the spectra channels to read their values back before saying the acquisition is done.  Because of that, the current scan controller that I built will put in the previous acquisition's data in it's raw data.  @davidChevrier, maybe we can figure that out quickly sometime so that BioXAS can use it as a debugging tool?


pluginProjects/FileLoaders/VESPERS2011XRFFileLoader/VESPERS2011XRFFileLoaderPlugin.cpp
```diff
@@ -69,6 +71,10 @@ bool VESPERS2011XRFFileLoaderPlugin::load(AMScan *scan, const QString &userDataF
 			// Note!  Not general!
 			QList<AMAxisInfo> axisInfo;
 			AMAxisInfo ai("Energy", 2048, "Energy", "eV");
+
+			if (elements == 32)
+				ai.size = 4096;
+
```
davidChevrier: Are you okay leaving this in here @dretrex ?

source/acquaman/BioXASSide/BioXASSideXRFScanController.cpp
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	QTextStream out(&file);
+
+	// There are two types of raw data sources.  The spectra of rank 1 and the dead time of rank 0.
+	for (int i = 0; i < scan_->rawDataSources()->count(); i++){
+
+		if (scan_->rawDataSources()->at(i)->rank() == 1){
+
+			qint32 counts = scan_->rawData()->value(AMnDIndex(), i, AMnDIndex(0));
```
davidChevrier: qint32 or quint32?

source/beamline/AMDetector.h
```diff
@@ -218,7 +218,7 @@ Q_OBJECT
 	/// Returns a string describing the given state
 	QString initializationStateDescription(AMDetector::InitializationState state);
 	/// Returns whether or not this detector currently requires initialization
-	bool requiresInitialization() const { return initializationState() == AMDetector::requiresInitialization(); }
+	bool requiresInitialization() const { return initializationState() == AMDetector::InitializationRequired; }
```
davidChevrier: Thanks @dretrex ... sometimes I'm an idiot :X

davidChevrier:
Aside from those comments it looks good ... and it's totally up to you if you want to do anything about them right now. Left me know.

pluginProjects/FileLoaders/VESPERS2011XRFFileLoader/VESPERS2011XRFFileLoaderPlugin.cpp
dretrex: Yeah.  What I should really do is make a new one since these XRF scans are actually incredibly general.

source/acquaman/BioXASSide/BioXASSideXRFScanController.cpp
dretrex: It probably should be quint, but I've been using qint32 for more than three years and it's been fine so far.  Should be fine for the time being.

dretrex:
I think it's fine as it is now.  We'll do more changes as it makes sense to do so.

davidChevrier:
Sure, we can leave those other pieces as discreet issues. Just to make sure, we're talking about:

- Generalizing the XRF scans
- Fixing up the detector state machine so that it's only finished after the data monitors
- Adding the element disable/enable to a lower level interface so that it can be used from the model and without any necessary relationship to dead time

Good list? More items I'm missing?

dretrex:
Nope, that sums it up concisely.

PR #1200, opened 2015-02-18
Pull Request for Issue857: Remove the database pointer and id Q_INVOKABLE constructors for AMDbObject

dretrex:
I removed all the old style ```AMDatabase *``` constructors since we don't use that schema anymore.  @davidChevrier could you make sure this doesn't seem too off it's rocker?  I also found that not all analysis blocks were setting up their default constructor properly and so I fixed those too.

Finally, I moved all the ```loadFromDb()``` implementations from all the analysis blocks and moved them to AMAnalysisBlock because there are no differences.


davidChevrier:
Looks good @dretrex 

I'll actually pull this one down and build it and see if I can pick up anything missed.

PR #1203, opened 2015-02-19
Issue1163 : Developing BioXAS Main.

helfrij:


source/acquaman/BioXASMain/BioXASMainXASScanConfiguration.cpp
```diff
@@ -0,0 +1,73 @@
...
+    AMScanAxisEXAFSRegion *region = new AMScanAxisEXAFSRegion;
+    AMScanAxis *axis = new AMScanAxis(AMScanAxis::StepAxis, region);
+    appendScanAxis(axis);
+}
+
+BioXASMainXASScanConfiguration::BioXASMainXASScanConfiguration(const BioXASMainXASScanConfiguration &original)
+    : AMStepScanConfiguration(original)
+{
+    setName(original.name());
+    setUserScanName(original.userScanName());
```
dretrex: You'll want to set the userScanName to name (```setUserScanName(original.name());```) because the user scan name is actually not used consistently.  Behaviour will be more what you'd expect with the above suggestion.

source/beamline/BioXAS/BioXASMainBeamline.cpp
```diff
@@ -23,91 +23,111 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
...
-	m2PseudoYaw_ = new BioXASPseudoMotorControl("BL1607-5-I21 Main M2 Yaw", "BL1607-5-I21:M2:Yaw", ":deg:fbk", ":deg", ":status", ":enabled", ":stop");
-	m2PseudoLateral_ = new BioXASPseudoMotorControl("BL1607-5-I21 Main M2 Lateral", "BL1607-5-I21:M2:Lateral", ":mm:fbk", ":mm", ":status", ":enabled", ":stop");
-
-	// BioXAS Mono Pseudo motors					   name,				   pvBaseName,				readPVname,	writePVname, movingPVname,	enabledPVname, stopPVname, tolerance, moveStartTimeoutSeconds, statusChecker, stopValue, description, parent = 0
-	monoPseudoEnergy_ = new BioXASPseudoMotorControl("BL1607-5-I21 Main Mono Energy", "BL1607-5-I21:Energy", ":EV:fbk", ":EV", ":status", ":enabled", ":stop");
+        if (!isConnected) {
+            actualConnected = isConnected;
+
+        } else {
+            actualConnected = (
```
dretrex: This is written really weird.

source/beamline/BioXAS/BioXASMainBeamline.h
```diff
@@ -57,12 +58,30 @@ class BioXASMainBeamline : public AMBeamline
 
 	/// Destructor.
 	virtual ~BioXASMainBeamline();
-
+    /// Returns true if all beamline components are connected, false otherwise.
```
dretrex: As a general comment, what are your tab spacings set to?  They don't seem to match your older code.

source/beamline/BioXAS/BioXASMainBeamline.cpp
helfrij: What is weird about it?

source/beamline/BioXAS/BioXASMainBeamline.h
helfrij: I have tab and indent sizes are set to 4. What should they be?

dretrex: That is what mine are set to as well.  Do you have "Insert spaces instead of tabs" enabled?

source/beamline/BioXAS/BioXASMainBeamline.cpp
dretrex: You have a single getter spread over three lines.  It reads awkwardly.

source/beamline/BioXAS/BioXASMainBeamline.h
davidChevrier: There are two places in creator that you need to enable that option.

Preferences->TextEditor->Behavior

and

Preferences->C++->CodeStyle->Edit->General

You might need to make a new profile for those C++ options.

source/acquaman/BioXASMain/BioXASMainXASScanConfiguration.cpp
helfrij: Okay, thanks!

source/beamline/BioXAS/BioXASMainBeamline.cpp
helfrij: It isn't a getter. It updates the beamline connected_ value in response to a change in a beamline component's connected state. I was hoping to take advantage of a shortcut. The beamline is always disconnected whenever any one component is disconnected, but it isn't always connected whenever any one component reports that it is. Does that make sense?

dretrex: Would it be reasonable to expect that as more systems come online that you'll have an ANDed combination of the various components there and not just the scaler?  If so, then I am happy with it.

helfrij: Yup, that's the idea.

dretrex:
Looks good @helfrij.

PR #1212, opened 2015-02-25
Pull Request for Issue1204: Create general mono class, ancestor to both Side and Main monos

helfrij:


source/beamline/BioXAS/BioXASSideMonochromator.cpp
```diff
@@ -78,124 +78,178 @@ BioXASSideMonochromator::~BioXASSideMonochromator()
...
+		return 0;
+
+	return AMActionSupport::buildControlMoveAction(crystalChangeMotorRel_, relDestination, true);
 }
 
-AMAction3* BioXASSideMonochromator::createMoveToCrystalChangePositionAction()
+AMAction3* BioXASSideMonochromator::createSetBraggMotorPowerOnAction()
 {
-	return createMoveStageAction(55);
+	return AMActionSupport::buildControlMoveAction(braggMotorPower_, 1);
```
XuetaoLiu-CLS: Just a question here: is it possible to make an enum definition for braggMotorPower status (the magic number "1" and "3" below) as what you did for the others (sucha as paddle, slit above)?

helfrij: @XuetaoLiu-CLS, yes! Good catch, I had missed that one.

dretrex:
Looking at how you've subclassed the mono, I think you could actually move most of the definitions of setters and getters to your super class and only have the subclass instantiate the controls with the correct PV's.  That will probably some code as well.

source/beamline/BioXAS/BioXASSSRLMonochromator.h
```diff
@@ -0,0 +1,104 @@
...
+#define BRAGG_MOTOR_CRYSTAL_CHANGE_POSITION 55.0
+
+
+class BioXASSSRLMonochromator : public QObject
+{
+	Q_OBJECT
+
+public:
+	/// Enumerates the different region states.
+	class Region { public: enum State { None = 0, A, B }; };
```
davidChevrier: @helfrij 
This is kinda cool. I've always had a problem with enum names in classes, because I would do something like
enum RegionState{ None, A, B} but then try to call it from another class and the name BioXASSSRLMonochromator::A wasn't that descriptive so I would change to enum RegionState{ RegionNone, RegionA, RegionB} which served the purpose, but seemed overly wordy.
With your way you can call BioXASSSRLMonochromator::Region::A from another class and be both descriptive and terse at the same time. Nice idea.

davidChevrier:
Hey @helfrij 
Nice job. I agree with @dretrex ... I think that once you've implemented the MainMono too you'll find that you can move a bunch of stuff from the subclasses into the parent class. I consider that to be a separate step, and otherwise everything looks good to me. Do you want me to merge this pull request and deal with the MainMono in another issue?

helfrij:
Moving some variables/method definitions to the superclass is a good idea, breaking it down into another issue might be the easiest way to deal with it. I'm good to merge now. Thanks for taking a look @XuetaoLiu-CLS @dretrex @davidChevrier!

PR #1219, opened 2015-03-04
Pull Request for Issue1027: Finish time based acquisitions

dretrex:
I have finished the time based scan acquisitions.  It operates in a ```<acquisition time> every <overall time period> <N> times```.  I'm sure there are other things that could be added, but as it stands now, it is finished to it's original spec.

### DIFF
--- a/source/acquaman/AMTimedScanActionControllerAssembler.cpp
+++ b/source/acquaman/AMTimedScanActionControllerAssembler.cpp
@@ -1,0 +1,5 @@
+#include "AMTimedScanActionControllerAssembler.h"
+
+AMTimedScanActionControllerAssembler::AMTimedScanActionControllerAssembler()
+{
+}
--- a/source/acquaman/AMTimedScanActionControllerAssembler.cpp
+++ b/source/acquaman/AMTimedScanActionControllerAssembler.cpp
@@ -24,13 +24,13 @@ AMTimedScanActionControllerAssembler::~AMTimedScanActionControllerAssembler()
 
 bool AMTimedScanActionControllerAssembler::generateActionTreeImplmentation()
 {
-	AMLoopAction3 *actionTree = new AMLoopAction3(iterations_);
-	actionTree->addSubAction(generateActionListForDetectorAcquisition());
-	actionTree->addSubAction(new AMWaitAction(new AMWaitActionInfo(timePerAcquisition_-acquisitionTime_)));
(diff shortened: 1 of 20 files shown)


source/acquaman/AMTimedScanActionControllerAssembler.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+AMTimedScanActionControllerAssembler::~AMTimedScanActionControllerAssembler()
+{
+
+}
+
+bool AMTimedScanActionControllerAssembler::generateActionTreeImplmentation()
+{
+	AMLoopAction3 *actionTree = new AMLoopAction3(iterations_);
```
davidChevrier: Naming is a little confusing here ... but I figured it out in the end

source/acquaman/AMTimedScanActionControllerAssembler.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+	actionTree_ = scanActions;
+
+	return true;
+}
+
+AMAction3* AMTimedScanActionControllerAssembler::generateActionListForDetectorAcquisition()
+{
+	AMListAction3 *retVal = new AMListAction3(new AMListActionInfo3(QString("Acquire All Detectors"), QString("Acquire %1 Detectors").arg(detectors_->count())), AMListAction3::Parallel);
+	qDebug() << detectors_->count();
```
davidChevrier: qDebug can be removed?

source/acquaman/AMTimedScanActionControllerAssembler.cpp
dretrex: If you have a better set of definitions I'll switch them.  They seemed the most appropriate at the time.

source/acquaman/AMTimedScanActionControllerAssembler.cpp
dretrex: Yeah, probably.

source/acquaman/AMTimedScanActionControllerAssembler.cpp
davidChevrier: Just that the "value loop" is called actionTree. It's a subaction of scanActions. And the actionTree_ member variable is set to scanActions. Maybe even renaming the first one mainLoop or something. Really, not that big a deal.

PR #1231, opened 2015-03-09
Issue1229: creating mono calibration widget.

helfrij:


source/beamline/BioXAS/BioXASSideMonochromator.cpp
```diff
@@ -31,6 +31,10 @@ BioXASSideMonochromator::BioXASSideMonochromator(QObject *parent) :
 	regionAStatus_ = new AMReadOnlyPVControl("RegionAStatus", "BL1607-5-I22:Mono:Region:A", this);
 	regionBStatus_ = new AMReadOnlyPVControl("RegionBStatus", "BL1607-5-I22:Mono:Region:B", this);
 
+	hcControl_ = new AMReadOnlyPVControl("hc Constant", "BL1607-5-I22:Energy:EV:fbk:tr.A", this);
```
dretrex: I just want to double check that the PV names are supposed to be named ```...:fbk:tr.A```.  That is a convention I have never seen before.  Do you know what the ```tr.A/B/K``` is supposed to imply?

dretrex:
Other than my PV comment, I think it looks fine.

source/beamline/BioXAS/BioXASSideMonochromator.cpp
helfrij: These are pv names that I got from a table of values in the BioXAS side mono energy application. The first value used in this application has extension .A, the next .B, and so forth (.M is the last). I haven't seen this convention used elsewhere, but it  seems to make sense internally.

dretrex: Okay.  Maybe @beaured0 can comment on it.  Then we'll know for sure!

PR #1244, opened 2015-03-17
Pull request for Issue1118: Creating mono controls

helfrij:
In branch Issue1118, both Side and Main Acquaman have been updated to include mono controls--mainly the new region control, mono calibration, and a couple of new views.


source/beamline/BioXAS/BioXASSSRLMonochromatorEnergyControl.cpp
```diff
@@ -0,0 +1,23 @@
...
+{
+
+}
+
+BioXASSSRLMonochromatorEnergyControl::~BioXASSSRLMonochromatorEnergyControl()
+{
+
+}
+
+AMControl::FailureExplanation BioXASSSRLMonochromatorEnergyControl::move(double setpoint)
```
dretrex: Is this to compensate for the fact that 0.001 is the precision of the mono energy PV and if you send an identical move it doesn't do anything - leading to AM procedures hanging?

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
```diff
@@ -0,0 +1,846 @@
...
+		braggAtCrystalChangePositionStatus_ &&
+		crystalChange_ &&
+		crystalChangeCWLimitStatus_ &&
+		crystalChangeCCWLimitStatus_ &&
+		regionAStatus_ &&
+		regionBStatus_
+	);
+
+	if (exists) {
+		connected = isConnected();
```
dretrex: Is the reason you overloaded the ```canMove()``` because you want to guarantee everything is connected but then knowing that not all connected controls should be checked for ```canMove()```?  Or even, the other controls would always create issue with properly reporting back the correct state of ```canMove()```?

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
```diff
@@ -0,0 +1,846 @@
...
+	if (!canMove())
+		return AMControl::LimitFailure;
+
+	// Now we can assume that the region control is ready for a crystal change.
+	// Update the saved setpoint value.
+
+	setSetpoint((int)setpoint);
+
+	if (!validRegionSetpoint(setpoint_)) {
+		return AMControl::LimitFailure;
```
dretrex: Would this be a good place to put an AMErrorMon of some kind?  Maybe even with some of it's state checks in the code above it?

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
```diff
@@ -0,0 +1,846 @@
...
+		addChildControl(regionAStatus_);
+
+		connect( regionAStatus_, SIGNAL(valueChanged(double)), this, SLOT(onRegionControlValueChanged()) );
+		connect( regionAStatus_, SIGNAL(movingChanged(bool)), this, SIGNAL(movingChanged(bool)) );
+	}
+}
+
+void BioXASSSRLMonochromatorRegionControl::setRegionBStatusControl(AMControl *regionStatus)
+{
+	if (!regionBStatus_ && regionBStatus_ != regionStatus) {
```
dretrex: A general comment about these ```setXYZControl()``` methods: 

I know that these are only called once inside the constructor, but if you were to use these methods in a more general sense (because you needed to change the controls for some reason) you would get some very bizarre behaviour.  For example, you set the new control normally, but you don't remove the old control from the child list and don't disconnect the signals from the old control.  This would lead to weird misfiring of slots and possibly weird behaviour in your status getters like (connected(), canMove(), etc.).

There is also a bit of an ownership issue.  Does this class own the control?  If so, it should handle deleting it.  If it doesn't, it should have a signal that lets others know that it no longer is looking after that control and those other classes can handle deleting/saving/handling it.

Just some things to keep in mind :)

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
```diff
@@ -0,0 +1,846 @@
...
+}
+
+void BioXASSSRLMonochromatorRegionControl::onRegionChangeStarted()
+{
+	setMoveInProgress(true);
+}
+
+void BioXASSSRLMonochromatorRegionControl::onRegionChangeCancelled()
+{
+	moveCleanup(sender());
```
dretrex: Although incredibly handy, using ```sender()``` isn't a good object-oriented programming principle.  To quote the Qt docs:

> Warning: This function violates the object-oriented principle of modularity. However, getting access to the sender might be useful when many signals are connected to a single slot.
Warning: As mentioned above, the return value of this function is not valid when the slot is called via a Qt::DirectConnection from a thread different from this object's thread. Do not use this function in this type of scenario.

This behaviour can be completely reproduced using the QSignalMapper class.  It's more code, but it is more reliable and future-proof.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.h
```diff
@@ -0,0 +1,284 @@
...
+	/// Called when the internal crystal change action has been cancelled. Handles emitting moveFailed(...) with the WasStoppedFailure code and deleting the action.
+	void onRegionChangeCancelled();
+	/// Called when the internal crystal change action has failed. Handles emitting moveFailed(...) with the OtherFailure code and deleting the action.
+	void onRegionChangeFailed();
+	/// Called when the internal crystal change action has succeeded! Handles emitting moveSucceeded() and deleting the action.
+	void onRegionChangeSucceeded();
+
+protected:
+	/// Returns a new action that performs a crystal change to change the region.
+	AMListAction3* createChangeRegionAction(int newRegion);
```
dretrex: Is there a particular reason why the controls are exposed as publicly and the action getters are protected?  Just curious.

source/ui/BioXAS/BioXASSSRLMonochromatorRegionControlView.cpp
```diff
@@ -0,0 +1,177 @@
...
+BioXASSSRLMonochromatorRegionControlView::~BioXASSSRLMonochromatorRegionControlView()
+{
+
+}
+
+void BioXASSSRLMonochromatorRegionControlView::setRegionControl(BioXASSSRLMonochromatorRegionControl *newControl)
+{
+	if (regionControl_ != newControl) {
+
+		if (regionControl_)
```
dretrex: See my earlier comment about the setXYZControl methods.  This is halfway there though.  It disconnects the old control from object, but doesn't handle the ownership issue.

source/ui/BioXAS/BioXASSidePersistentView.cpp
```diff
@@ -61,7 +72,7 @@ BioXASSidePersistentView::~BioXASSidePersistentView()
 void BioXASSidePersistentView::onCalibrateEnergyButtonClicked()
 {
 	bool inputOK = false;
-	double newEnergy = QInputDialog::getDouble(this, "Monochromator Energy Calibration", "Enter current calibrated energy:", BioXASSideBeamline::bioXAS()->mono()->energyControl()->value(), 0, 10000000, 1, &inputOK, Qt::Sheet);
+	double newEnergy = QInputDialog::getDouble(this, "Monochromator Energy Calibration", "Enter current calibrated energy:", BioXASSideBeamline::bioXAS()->mono()->energyControl()->value(), -10000000, 10000000, 1, &inputOK, Qt::Sheet);
```
dretrex: Why is the minimum changed to -1000000?  I would have thought the old value of 0 would be appropriate for the energy.

dretrex:
Overall, this looks great @helfrij!  Great work!! :+1:

source/beamline/BioXAS/BioXASSSRLMonochromatorEnergyControl.cpp
davidChevrier: That's a good question @dretrex 
@helfrij this is one of those things that crops up a lot on actual controls and can cause no end of headaches. What behavior are you trying to address with this check? Maybe the three of us should meet up and discuss it.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
davidChevrier: Good point @dretrex 
I'm as guilty as anyone of using sender() ... guiltier than most, really. QSignalMapper is more work but is a much better solution.

davidChevrier:
I would love to echo @dretrex 's sentiment ... but, sadly, I haven't read through the actual full changes :'(

That said, thanks for the highlights @dretrex , I did check each of your commented sections.

I did pull and build yesterday and pushed a change back up so that acquaman_suite can build, so in my view things are good to go and can be merged whenever (whether you want to merge and then make new issues to address any comments, or whether you want to address them now, or whether you want to leave them alone is your call @helfrij )

Cheers!

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
helfrij: I thought a little about these issues, but wasn't quite sure how to address them. I thought using the methods in this way would be safe enough if the slots were protected and had a (!control_) guard in their bodies, which I thought would mean once set they couldn't be changed. As written, the controls belong to the region control subclass, though nowhere is this enforced. There is certainly room for improvement where flexibility is concerned.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.h
helfrij: I wanted the control to be operated using only the general control interface (manipulating the control using an action or using the move slot). No other reason.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
dretrex: You are completely correct that the code - as currently written - is safe.  They are only called in the constructor and for the time being that is all that matters.  However, there is a case where a subclass could do ```setXYZControl(0)``` which would cause all kinds of issues.  

I wouldn't worry about it too much at the moment.  Maybe just flag it for code cleanup at a later date?

source/ui/BioXAS/BioXASSidePersistentView.cpp
helfrij: The value itself was arbitrarily chosen. At present, it's possible to have a negative energy reading on the beamline, which is why the minimum here is negative.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.h
dretrex: Ah, okay.  And the AMControl * getters are there just so you can retrieve the various mono controls from the beamline?

source/beamline/BioXAS/BioXASSSRLMonochromatorEnergyControl.cpp
helfrij: Hmmm, as far as I know, this class was existing before this issue was created, and I'm not sure I remember why. I did rename it and am thinking I might move the energy calibration code into it eventually. Is there a problem with it as written?

source/ui/BioXAS/BioXASSidePersistentView.cpp
dretrex: I knew that BioXAS was trying to push the limits, but negative energy is pretty bold ;).  Out of curiosity, how is it possible to have a negative energy?  Otherwise, the change itself is okay.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
helfrij: It's possible as written that some of the controls won't be set properly; the region control shouldn't move if this is the case. I'm not sure that all of the child controls can actually move (I'm thinking of the status controls in particular), and I didn't want that to impact the overall canMove state of the region control. But there is definitely a problem if moving parts (slit motors, paddle motor, bragg motor, and crystal change motor controls) can't move.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
helfrij: Great idea!

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.h
helfrij: That's right.

source/ui/BioXAS/BioXASSidePersistentView.cpp
helfrij: I think it may technically be possible at some point during a crystal change, though I'd have to run one though again to check.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
dretrex: Okay, got it.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.h
dretrex: Gotcha.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
helfrij: Good idea. I'll create an issue now so I don't forget.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
helfrij: I agree that this should be changed. If it's alright, I'd like to take care of it in a separate issue.

dretrex: Dealing with it in a separate issue is fine.

helfrij:
Thanks for your input today, everyone. Concerns were raised about the use of AMErrorMon, funny setXYZControl methods, and sender(). ErrorMon stuff has been addressed in this branch, I believe. The latter two will be addressed as separate issues (#1248, #1249). If there's nothing else, I'm ready to merge.

dretrex:
@helfrij , since @davidChevrier jumped the gun a bit, you may want to make another issue about your AMErrorMons.  I believe you wanted to make alerts and not errors.  The proper syntax for that is:

```AMErrorMon::alert(this, errorCode, description);```

and not ```AMErrorMon::error()```.

The error code is usually a ```#define``` (lots of examples to look at currently for format, choice, etc.).  You'll probably want a different code for each type of failure so you can really lock down what was the real error.

helfrij:
Thanks for the headsup @dretrex. I'll make a new issue to deal with this now.

davidChevrier:
I'm a disaster. My bust.

Issue #1215, opened 2015-02-27
SXRMB: add save feature for BrukeDetector data


XuetaoLiu-CLS:
Users of SXRMB BL asked for a feature to save the spectum data  and the bruker data

dretrex:
You did this already, right @XuetaoLiu-CLS?  What branch did you do it in?

XuetaoLiu-CLS:
yes. I didn't get an opportunity to test it yet

XuetaoLiu-CLS:
took about 3ds

Issue #1191, opened 2015-02-10
Generalize XRF scans


dretrex:
Since all the XRF scans behave in the same way.  They can and should be generalized.  They would greatly benefit from a more out of the box implementation.  This would likely come with a new file loader, but that won't be too bad as it will be able to reuse almost everything from the existing one.  Or this could be a good opportunity to test out making our own binary file loader?  Thoughts @davidChevrier?

davidChevrier:
I'm all for trying this one out as binary. quint32 with no compression, here we come! Still, a substantial improvement over ASCII text.

dretrex:
This is getting closer to being a real priority due to #1376.  @XuetaoLiu-CLS, this is the issue I was mentioning earlier.  You don't need to worry about it, but I wanted to flag it since we were talking about it earlier.

dretrex:
Now that we are where we are at, the XRF scan controllers should use the CDF file loader.

dretrex:
Or HDF5 when we switch to it.

dretrex:
Time Estimate: 30 minutes

dretrex:
Done in #1928.

PR #1246, opened 2015-03-18
PullRequest for Issue1215: add function to save bruker scan data

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1215

Add XRF scan controller/configuration for bruker detector, so that the user can save/export bruker detector scan data.

@dretrex @davidChevrier


source/acquaman/SXRMB/SXRMBXRFScanConfiguration.h
```diff
@@ -0,0 +1,61 @@
...
+#ifndef VESPERSXRFSCANCONFIGURATION_H
+#define VESPERSXRFSCANCONFIGURATION_H
+
+#include "acquaman/AMScanConfiguration.h"
+
+class SXRMBXRFScanConfiguration : public AMScanConfiguration
+{
+	Q_OBJECT
+
+	Q_CLASSINFO("AMDbObject_Attributes", "description=VESPERS XRF Scan Configuration")
```
dretrex: This VESPERS description should probably be changed ;)

source/acquaman/SXRMB/SXRMBXRFScanController.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+	int elements = detector_->elements();
+
+	for (int i = 0; i < elements; i++){
+
+		scan_->rawData()->addMeasurement(AMMeasurementInfo(QString("raw%1").arg(i+1), QString("Element %1").arg(i+1), "eV", detector_->axes()));
+		scan_->addRawDataSource(new AMRawDataSource(scan_->rawData(), i), true, false);
+	}
+
+	for (int i = 0; i < elements; i++){
```
dretrex: I thought the Bruker didn't have ICR and OCR?  Won't this cause crashes?

source/acquaman/SXRMB/SXRMBXRFScanController.cpp
```diff
@@ -0,0 +1,185 @@
...
+	disconnect(detector_, SIGNAL(acquisitionStateChanged(AMDetector::AcqusitionState)), this, SLOT(onStatusChanged()));
+	disconnect(detector_, SIGNAL(elapsedTimeChanged(double)), this, SLOT(onProgressUpdate()));
+
+	QVector<double> spectrum = QVector<double>(detector_->size(0));
+
+	for (int i = 0, elements = detector_->elements(); i < elements; i++){
+
+		detector_->rawSpectrumSources().at(i)->values(AMnDIndex(0), AMnDIndex(detector_->size(0)-1), spectrum.data());
+		scan_->rawData()->setValue(AMnDIndex(), i, spectrum.constData());
+		if (detector_->inputCountSourceAt(i))
```
dretrex: Same comment as earlier.

source/application/BioXAS/BioXASImagingAppController.cpp
```diff
@@ -21,6 +21,7 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "BioXASImagingAppController.h"
 
+#include "beamline/CLS/CLSBeamlines.h"
```
dretrex: I'm confused.  What is this?

source/application/SXRMB/SXRMB.h
```diff
@@ -19,7 +19,9 @@
 namespace SXRMB {
 
 	enum SXRMBErrorCodes {
-		ErrorSXRMBBeamlineShuttersTimeout = 290101  // 2-CLS 9-SXRMB 01 module-01 errorID-01
+		ErrorSXRMBBeamlineShuttersTimeout = 290101,  // 2-CLS 9-SXRMB 01 module-01 errorID-01
```
dretrex: I'm not entirely sold on this since we did not come up with a specific solution.  What are your thoughs @davidChevrier ?

source/util/AMErrorMonitor.h
```diff
@@ -161,9 +163,11 @@ class AMErrorMon : public QObject {
 	}
 
 	/// Report an alert level AMErrorReport.  Builds an AMErrorReport using the information provided. This function is thread-safe.
-	static void alert(const QObject *src = 0, int code = 0, const QString &desc = "")
+	static void alert(const QObject *src = 0, int code = 0, const QString &desc = "", bool showWarningMessage=false)
```
dretrex: What is this?  Why is it only in alerts?

source/acquaman/SXRMB/SXRMBXRFScanController.cpp
XuetaoLiu-CLS: `inputCountSourceAt(0)` will return "0" when there is no ICR. This  `if (detector_->inputCountSourceAt(i))` guarded this scenario.

source/application/SXRMB/SXRMB.h
XuetaoLiu-CLS: @dretrex @davidChevrier yes, we don't have a solution yet. This is what I have already before our discussion. I simply follow what I have had already. It will be an easy fix when we have a solution.

source/util/AMErrorMonitor.h
XuetaoLiu-CLS: I added this flag for code simplification.` AMErrorMonitor::alert()` will only create an alert message in the error log window. Sometime, we might want a visual alert box popped up. 

I have two usages for that case. That's why I was thinking of making changes here. I can change it back for sure if you are against this.

source/application/BioXAS/BioXASImagingAppController.cpp
XuetaoLiu-CLS: if you take a look at the first commit only, you might understand. I was trying to put our Beamline definitions in a single file, so that we won't have duplicated definitions in the future anymore.

source/acquaman/SXRMB/SXRMBXRFScanController.cpp
davidChevrier: Will it? Am I missing something here? The call to `detector_->inputCountSourceAt(i)` actually just retrieves from a QList. The `at()` function is only safe for numbers less than size. Unless I'm missing something the size of that list right now should be 0, so I'm not sure what this is going to do.

dretrex: @davidChevrier is correct.  If there are no ICR or OCR data sources you will get an index out of bounds exception from QList.

XuetaoLiu-CLS: @davidChevrier @dretrex I will make one more test with this.

source/application/BioXAS/BioXASImagingAppController.cpp
davidChevrier: I can see what you're trying to do here Sheldon. I think it's a good idea. I'm not a big fan of the nomenclature for the QStrings though, it really should be camel case and fully worded:
`SGMBeamlineName` rather than `SGM_bl_name`
Also, I would say that it really should be a function of the namespace, something that takes one of the enums and returns the correct string.
`QString beamlineName(CLSBeamline::CLSBeamlines beamline)`

source/application/SXRMB/SXRMB.h
davidChevrier: For the time being, let's use the old error code style (#defines in the header of the file). I want to get this sorted out, but I want to do it consistently.

source/util/AMErrorMonitor.h
davidChevrier: I can see the value of this, but if it's going to be available for alert, then it should be available for the other ones as well.

davidChevrier:
@XuetaoLiu-CLS I think part of the problem here is that the issue, as stated, is relatively straight forward. However, including other larger/framework additions muddies the waters a bit. I can see this both ways: it would be easier for review to split these other changes out into other issues however it would also make more work for you when you're actually coding. Maybe a good rule of thumb would be to make new issues for framework changes if the original issue you're working on is specific to one beamline. Probably worth discussing in the group setting ... thoughts @dretrex @helfrij @iainworkman @epengr @XuetaoLiu-CLS @beaured0 @dleacock ?

XuetaoLiu-CLS:
@davidChevrier I agree with a separate issue for code refactoring.

Everytime we think we need a code refactor, we might need to raise an issue and make the code refactor. Then in the original issue, we can merge that PR in.

davidChevrier:
Since this is all coupled together at the moment, I think you can address the comments and we can still merge this all as one issue/pull request this time (I'm loathe to make more work). I think we can discuss things on Monday and figure out how best to do it going forward. @dretrex will be away for the bulk of today, but if you push up changes I'll try my best to review them and get this pull request merged in.

source/util/AMErrorMonitor.h
XuetaoLiu-CLS: @davidChevrier I agree.

I prefer to add it to other functions. But I was given a "hard" lesson when I was the other company for this type of changes. They do NOT want to add features to the codes if they are not used now. :disappointed:

source/application/BioXAS/BioXASImagingAppController.cpp
XuetaoLiu-CLS: I will rename the Names. Unfortunately, we can't use a function to get the beamlineName now as you suggested since out Beamline IDs are NOT unique. For example, 5 could be either REIX beamline or IDEAS beamline.

source/util/AMErrorMonitor.h
davidChevrier: Let's add it to the other functions. Acquaman is still agile enough that we can plan for the future as we make code changes.

source/application/BioXAS/BioXASImagingAppController.cpp
davidChevrier: Yeah, I knew that would be a problem when I suggested it. However, it's a convenience function ... so you don't have to call it. I would say the best thing to do is to return some sort of string that indicates the problem. Where there are problem we can still directly pass a QString as:
`AMRun firstRun("REIXS", CLSBeamline::REIXS);`
I would prefer to stay away from global or namespaced const variables.

source/acquaman/SXRMB/SXRMBXRFScanController.cpp
XuetaoLiu-CLS: @davidChevrier @dretrex Accessing a QList out of its range doesn't throw an exception. Instead, it assumes that you already checked that the index is in the range. 

I need to add more strict protection here.

davidChevrier: Holy wow! That might explain a lot. @dretrex and I have seen different behaviors causing crashes or not on different OS's ... maybe it has to do with how Qt was compiled in the first place. I see Q_ASSERT failures all of the time on my mac when I index outside of the bounds of a QList. Very interesting.

dretrex: I also see Q_ASSERT failures all the time on my ubuntu machine.

XuetaoLiu-CLS: This is my testing codes:
```
	QList<int> testQList;
	qDebug() << "QList size: " << testQList.size();
	qDebug() << "QList at 0: " << testQList.at(0);
	qDebug() << "QList at 1: " << testQList.at(1);
```
and output:
```
QList size:  0 
QList at 0:  0 
QList at 1:  1833979424 
```

dretrex: That is quite troubling.  You should not be able to access that data at all, let alone get numbers from 0 AND 1.  If this is a scientific linux thing, that would explain why sometimes these bad array indexing bugs bite us so bad.  We would find them immediately on ubuntu/mac machines, but would just get chaotic behaviour from beamline testing.

dretrex: Good find @XuetaoLiu-CLS.

XuetaoLiu-CLS:
@davidChevrier @dretrex please take a look at the new changes. Thanks

source/acquaman/SXRMB/SXRMBXRFScanController.cpp
```diff
@@ -0,0 +1,189 @@
...
+	if (detector_->outputCountSources().size() > 0) {
+		for (int i = 0; i < elements; i++){
+
+			scan_->rawData()->addMeasurement(AMMeasurementInfo(QString("ocr%1").arg(i+1), QString("Output count rate %1").arg(i+1), "%", QList<AMAxisInfo>()));
+			scan_->addRawDataSource(new AMRawDataSource(scan_->rawData(), i+2*elements), false, true);
+		}
+	}
+
+	for (int i = 0; i < elements; i++){
+
```
dretrex: The dead time analysis block shouldn't be used either if the ICR/OCR sources don't exist.

XuetaoLiu-CLS: @dretrex Thanks for pointing this out. It is fixed.

XuetaoLiu-CLS:
@dretrex  yes, they can. My intention to leave there is that you mentioned we are going to have a general version for this. I was thinking this can make it clear to point the difference with the VESPERS version.

dretrex:
Unfortunately, based on our discussion of what will likely cause crashes, your qDebug() statement would likely cause crashes on my ubuntu machine and macs.  I think it's fine to remove the qDebug() statements and we can handle the generality issue in #1191.

XuetaoLiu-CLS:
This will do for now. We will continue our discussion about error codes and similar topics at a later date.

PR #1252, opened 2015-03-19
Pull request for Issue1223: Create mono configuration view

helfrij:
In this branch, I've created a simple configuration view for the SSRL monochromators. It displays the region status view and enables the user to view and configure some bragg motor properties.


source/ui/BioXAS/BioXASSSRLMonochromatorConfigurationView.h
```diff
@@ -0,0 +1,116 @@
...
+#include <QLabel>
+
+#include "beamline/BioXAS/BioXASSSRLMonochromator.h"
+#include "ui/BioXAS/BioXASSSRLMonochromatorRegionControlView.h"
+#include "ui/BioXAS/BioXASSSRLMonochromatorRegionControlEditor.h"
+
+#define BRAGG_OFFSET_MIN -1000
+#define BRAGG_OFFSET_MAX 1000
+#define BRAGG_SETTLING_TIME_MIN -1000
+#define BRAGG_SETTLING_TIME_MAX 1000
```
davidChevrier: Hey @helfrij , normally we don't do this type of `#define` action ... that being said we usually do something way worse and just put magic numbers right in the code. I don't imagine there's one-size-fits-all solution here ... but I bet it's the start of a good conversation. Maybe we can have some guidelines as to when to use `#define` versus member functions versus enums versus magic numbers. 

@dretrex @epengr @iainworkman @XuetaoLiu-CLS @dleacock 

(As it stands I'm fine with this. Everything looks good to me otherwise. I can probably merge if you like ... if I don't jump the gun this time.)

epengr: Agreed.  This is a serious issue for REIXS.  The alignment and calibration of the spectrometer is hard coded magic numbers, which means that when I install the new gratings (when I eventually get the new gratings), and the calibration values change, not only will I have to tweak numerous hard-coded magic numbers, all previously collected data will not display correctly (the energy axis of old XES scans will be wrong).  

Creating a "beamlineCalibration" class that's a snap shot of all magic numbers is clearly overkill for this particular situation, but something like it may have some merit.

PR #1259, opened 2015-03-23
Pull request for Issue1253: Adding bragg motor power changes to scan controller initialization and cleanup actions.

helfrij:


dretrex:
Looks good @helfrij.

PR #1261, opened 2015-03-23
Pull request for Issue1251: Creating unique error codes for BioXASSSRLMonochromatorRegionControl.

helfrij:


dretrex:
Hey @helfrij, I just wanted to double check that you actually wanted to call ```error()``` as opposed to ```alert()```.  Since you used the alert flag as your error code, I figured your error monitors would be alerts.  If you want them to be errors, they look good and I'll merge them in.

helfrij:
@dretrex 
I think I'll keep them as errors because they happen when something about the region control is absolutely incorrect (as opposed to a 'nice-to-know'). That okay?

dretrex:
Yup, totally fine.

PR #1271, opened 2015-03-26
Pull request for Issue1263: Modifying mono configuration view.

helfrij:


dretrex:
In general, this looks good.  I noticed one place where you could save a bit of code in your super class implementation of the SSRL mono control class.  Rather than making the getters pure virtual functions, you could have the four controls be members in the super class, and have all the accessors implemented.  Then your subclass would only need to instantiate the controls in their constructor.

helfrij:
Thanks @dretrex. I'll have a look and make the changes after the weekend.

PR #1281, opened 2015-03-30
Pull request for Issue1264: Implementing dark current correction on Main.

helfrij:


source/acquaman/BioXAS/BioXASMainXASScanActionController.cpp
```diff
@@ -99,13 +99,35 @@ AMAction3* BioXASMainXASScanActionController::createInitializationActions()
     stage4->addSubAction(scaler->createStartAction3(true));
     stage4->addSubAction(scaler->createWaitForDwellFinishedAction(regionTime + 5.0));
 
+    // setup beamline for dark current correction.
+
+    AMListAction3 *darkCurrentSetup = new AMListAction3(new AMListActionInfo3("BioXAS Main Dark Current Setup", "BioXAS Main Dark Current Setup"), AMListAction3::Parallel);
+
+    for (int i = 0, size = configuration_->detectorConfigurations().count(); i < size; i++) {
+	    AMDetector *detector = AMBeamline::bl()->exposedDetectorByInfo(configuration_->detectorConfigurations().at(i));
+	    bool sharedSourceFound = false;
```
dretrex: I don't think ```sharedSourceFound``` ever changes.  Maybe it was just left behind from refactoring?

source/acquaman/BioXAS/BioXASMainXASScanActionController.cpp
```diff
@@ -99,13 +99,35 @@ AMAction3* BioXASMainXASScanActionController::createInitializationActions()
...
+
+	    if (detector) {
+		    int detectorIndex = scan_->indexOfDataSource(detector->name());
+
+		    // if we have a valid detector that can perform dark current correction, make that detector perform a dark current measurement.
+		    if (detectorIndex != -1 && detector->rank() == 0 && detector->canDoDarkCurrentCorrection()) {
+			    bool sourceShared = detector->sharesDetectorTriggerSource();
+
+			    if ((sourceShared && !sharedSourceFound) || !sourceShared)
```
dretrex: Also, using it here after it is never changed unnecessarily complicates this if statement.

source/acquaman/BioXAS/BioXASMainXASScanActionController.cpp
```diff
@@ -127,7 +149,39 @@ AMAction3* BioXASMainXASScanActionController::createCleanupActions()
...
+
+	int dwellTimeIndex = scan_->indexOfDataSource(BioXASMainBeamline::bioXAS()->dwellTimeDetector()->name());
+
+	if (dwellTimeIndex != -1) {
+		AMDataSource *dwellTimeSource = scan_->dataSourceAt(dwellTimeIndex);
+
+		for (int i = 0, size = configuration_->detectorConfigurations().count(); i < size; i++) {
+			AMDetector *detector = AMBeamline::bl()->exposedDetectorByInfo(configuration_->detectorConfigurations().at(i));
+
+			if (detector) {
```
dretrex: This is an interesting state check.  This implies that a detector that was added to the configuration was not available in the beamline.  Maybe this should have an else with an error monitor to notify if this ever happens?

source/actions3/actions/AMDetectorSetDarkCurrentTimeAction.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+AMDetectorSetDarkCurrentTimeAction::~AMDetectorSetDarkCurrentTimeAction()
+{
+
+}
+
+void AMDetectorSetDarkCurrentTimeAction::startImplementation()
+{
+	if (!detector_)
```
dretrex: This check will cause the action to hang.  Since this is very undesirable, you likely need to call ```setFailed()```.

source/actions3/actions/AMDetectorSetDarkCurrentTimeActionInfo.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+	setShortDescription(typeDescription());
+	setLongDescription(typeDescription());
+}
+
+AMDetectorSetDarkCurrentTimeActionInfo::AMDetectorSetDarkCurrentTimeActionInfo(const AMDetectorSetDarkCurrentTimeActionInfo &other) :
+	AMActionInfo3(other)
+{
+	detectorInfo_.setValuesFrom( *(other.detectorInfo()) );
+	darkCurrentTime_ = other.darkCurrentTime_;
```
dretrex: There might be a weird copy constructor rule that allows you to do this, but generally this should be ```darkCurrentTime_ = other.darkCurrentTime();``` because ```darkCurrentTime_``` is protected.

source/actions3/actions/AMDetectorSetDarkCurrentTimeAction.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+	detector_ = 0;
+	setDetector(detector);
+}
+
+AMDetectorSetDarkCurrentTimeAction::AMDetectorSetDarkCurrentTimeAction(const AMDetectorSetDarkCurrentTimeAction &other) :
+	AMAction3(other)
+{
+	detector_ = 0;
+	setDetector(other.detector_);
```
dretrex: See previous comment on using ```other.detector()``` instead of ```other.detector_```.

source/actions3/actions/AMDetectorSetDarkCurrentValidStateAction.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+	detector_ = 0;
+	setDetector(detector);
+}
+
+AMDetectorSetDarkCurrentValidStateAction::AMDetectorSetDarkCurrentValidStateAction(const AMDetectorSetDarkCurrentValidStateAction &other) :
+	AMAction3(other)
+{
+	detector_ = 0;
+	setDetector(other.detector_);
```
dretrex: Here too.

source/actions3/actions/AMDetectorSetDarkCurrentValidStateAction.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+AMDetectorSetDarkCurrentValidStateAction::~AMDetectorSetDarkCurrentValidStateAction()
+{
+
+}
+
+void AMDetectorSetDarkCurrentValidStateAction::startImplementation()
+{
+	if (!detector_)
+		setDetector(0);
```
dretrex: State machine will get hung up here.

source/actions3/actions/AMDetectorSetDarkCurrentValidStateActionInfo.cpp
```diff
@@ -0,0 +1,30 @@
...
+	detectorInfo_.setValuesFrom(detectorInfo);
+
+	setShortDescription(typeDescription());
+	setLongDescription(typeDescription());
+}
+
+AMDetectorSetDarkCurrentValidStateActionInfo::AMDetectorSetDarkCurrentValidStateActionInfo(const AMDetectorSetDarkCurrentValidStateActionInfo &other) :
+	AMActionInfo3(other)
+{
+	darkCurrentState_ = other.darkCurrentState_;
```
dretrex: Here too.

source/actions3/actions/AMDetectorSetDarkCurrentValueAction.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+	detector_ = 0;
+	setDetector(detector);
+}
+
+AMDetectorSetDarkCurrentValueAction::AMDetectorSetDarkCurrentValueAction(const AMDetectorSetDarkCurrentValueAction &other) :
+	AMAction3(other)
+{
+	detector_ = 0;
+	setDetector(other.detector_);
```
dretrex: Here too.

source/actions3/actions/AMDetectorSetDarkCurrentValueAction.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+AMDetectorSetDarkCurrentValueAction::~AMDetectorSetDarkCurrentValueAction()
+{
+
+}
+
+void AMDetectorSetDarkCurrentValueAction::startImplementation()
+{
+	if (!detector_)
+		setDetector(0);
```
dretrex: Gets hung up here.

source/actions3/actions/AMDetectorSetDarkCurrentValueActionInfo.cpp
```diff
@@ -0,0 +1,32 @@
...
+	detectorInfo_.setValuesFrom(detectorInfo);
+
+	setShortDescription(typeDescription());
+	setLongDescription(typeDescription());
+}
+
+AMDetectorSetDarkCurrentValueActionInfo::AMDetectorSetDarkCurrentValueActionInfo(const AMDetectorSetDarkCurrentValueActionInfo &other):
+	AMActionInfo3(other)
+{
+	darkCurrentValue_ = other.darkCurrentValue_;
```
dretrex: Here too.

source/actions3/actions/AMDetectorSetDarkCurrentValueActionInfo.cpp
```diff
@@ -0,0 +1,32 @@
+#include "AMDetectorSetDarkCurrentValueActionInfo.h"
+
+AMDetectorSetDarkCurrentValueActionInfo::AMDetectorSetDarkCurrentValueActionInfo(double newValue, const AMDetectorInfo &detectorInfo, QObject *parent) :
+    AMActionInfo3(QString(), QString(), QString(), parent)
```
dretrex: Could these have names and descriptions?  I think the earlier actionInfo's are like this too.  Generally names and descriptions are handy when debugging buggy scans in the scanActionDatabase.

source/actions3/actions/CLSSIS3820ScalerDarkCurrentMeasurementActionInfo.h
```diff
@@ -0,0 +1,37 @@
...
+#include "actions3/AMActionInfo3.h"
+#include "beamline/CLS/CLSSIS3820Scaler.h"
+
+class CLSSIS3820ScalerDarkCurrentMeasurementActionInfo : public AMListActionInfo3
+{
+    Q_OBJECT
+
+public:
+	/// Constructor.
+	Q_INVOKABLE CLSSIS3820ScalerDarkCurrentMeasurementActionInfo(CLSSIS3820Scaler *scaler, double secondsDwell, QObject *parent = 0);
```
dretrex: This is more of a question for @davidChevrier.  Although this is definitely the most convenient and fast way to get a hold of the scaler, won't this cause issues with database loading?  This would need access to the beamline for loading which is generally something we try to avoid with all info and dataman classes.  I think we may need to add a generic way to handle getting the scaler from beamline classes.

source/actions3/actions/CLSSIS3820ScalerDarkCurrentMeasurementAction.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+void CLSSIS3820ScalerDarkCurrentMeasurementAction::initialize()
+{
+
+}
+
+void CLSSIS3820ScalerDarkCurrentMeasurementAction::startImplementation()
+{
+	qDebug() << "CLSSIS3820ScalerDarkCurrentMeasurementAction::startImplementation is executing.";
```
dretrex: You probably don't need to reimplement the startImplementation anymore since it was only used for debugging.

source/actions3/actions/CLSSIS3820ScalerDarkCurrentMeasurementActionInfo.cpp
```diff
@@ -0,0 +1,30 @@
+#include "CLSSIS3820ScalerDarkCurrentMeasurementActionInfo.h"
+
+CLSSIS3820ScalerDarkCurrentMeasurementActionInfo::CLSSIS3820ScalerDarkCurrentMeasurementActionInfo(CLSSIS3820Scaler *scaler, double secondsDwell, QObject *parent) :
+	AMListActionInfo3(QString(), QString(), QString(), parent)
```
dretrex: @helfrij, this action/actionInfo looks good other than my comment to @davidChevrier.  We may need to come up with a better way to get a hold of the scaler pointer, but everything looks fine.

source/application/BioXAS/BioXASMainAppController.cpp
```diff
@@ -116,8 +120,25 @@ void BioXASMainAppController::onScalerConnected()
 	CLSSIS3820Scaler *scaler = BioXASMainBeamline::bioXAS()->scaler();
 
     if (scaler && scaler->isConnected() && !scalerView_) {
-	scalerView_ = new CLSSIS3820ScalerView(scaler, false);
-	mw_->addPane(scalerView_, "Detectors", "Scaler", ":/system-search.png", true);
+
+	    scalerView_ = new CLSSIS3820ScalerView(scaler, true);
```
dretrex: I'm noticing a general trend where we have widgets and then we box them together with squeeze's and add a top frame.  Maybe this could be a nice handy convenience method (something like ```createMainWindowPane(const QString &topFrameName, const QString &topFrameIcon, QWidget *widget);```?  Thoughts @davidChevrier?

@dleacock would you be interested in this task?

source/beamline/AMDetector.cpp
```diff
@@ -428,28 +459,34 @@ bool AMDetector::clear(){
...
-void AMDetector::setAsDarkCurrentMeasurementValue(){
+void AMDetector::setLastMeasurementAsDarkCurrent()
+{
+	if (canDoDarkCurrentCorrection()) {
+		double newTime = acquisitionTime();
+		double newValue = double(singleReading()) / newTime;
+
+		updateDarkCurrentTime(newTime);
+		updateDarkCurrentValue(newValue);
+		updateDarkCurrentValidState(true);
```
dretrex: Just for naming's sake, I'd change these from update methods to set methods.  This is particularly true for ```updateDarkCurrentValidState()``` as it reads "Should I check if I'm valid? Yes." instead of "You are now valid."

source/beamline/AMDetector.h
```diff
@@ -298,10 +298,17 @@ int outputSize = indexStart.totalPointsTo(indexEnd);
 	/// Returns a newly created action (possibly list of actions) to perfrom the detector cleanup
 	virtual AMAction3* createCleanupActions();
 
-	/// Returns a (list of) actions that will do all necessary steps to do dark current correction. Your subclass needs to implement this function. Dwell time argument is in seconds.
-	virtual AMAction3* createDarkCurrentCorrectionActions(double dwellTime);
-	/// Returns an action that sets the current detector value as the dark current correction value. Your subclass will probably call this in createDarkCurrentCorrectionActions.
-	virtual AMAction3* createSetAsDarkCurrentCorrectionAction();
+	/// Returns new action that initiates a dark current measurement, uses the provided dwell time. An AMDetector subclass that can do dark current correction would have to implement this.
+	virtual AMAction3* createDarkCurrentMeasurementAction(double secondsDwell) { Q_UNUSED(secondsDwell) return 0; }
```
dretrex: If this method isn't being used anymore, should we just remove it?

source/ui/CLS/CLSSIS3820ScalerChannelViewWithDarkCurrent.cpp
```diff
@@ -51,91 +50,76 @@ CLSSIS3820ScalerChannelViewWithDarkCurrent::CLSSIS3820ScalerChannelViewWithDarkC
...
-	// the latest value of the detector's dark current. this is normalized by its acquisition time already (see AMDetector::setAsDarkCurrentMeasurementValue()).
-	double darkCurrent = channel_->detector()->darkCurrentMeasurementValue();
-
-	double correctedMeasurement_ = *newValue / dwellTime - darkCurrent;
-
-	qDebug() << "CLSSIS3820ScalerChannelView has new DC corrected value : " << correctedMeasurement_;
-
-	setCorrectedMeasurementLabel(correctedMeasurement_);
+	switch (newState) {
+	case AMDetector::Succeeded:
```
dretrex: Do we not need to check the other states?  I'm thinking Failed and Cancelled in particular.

dretrex:
There were a couple more constructors that you look at, I just wanted to stop making silly comments like "here too".  You will just need to make sure to go over all the actions and not just the ones I flagged.

source/acquaman/BioXAS/BioXASMainXASScanActionController.cpp
helfrij: I think it was originally included so that the scaler performs one dark current measurement for all channel detectors, not one for each channel detector. Where I think I went wrong here was not including a line that changes sharedSourceFound from false to true when the channel detector has a shared trigger source.

source/actions3/actions/AMDetectorSetDarkCurrentValueActionInfo.cpp
helfrij: Because the name and description are the same as the typeDescription(), I opted to set them in the subclass constructor using setShortDescription() and setLongDescription(). Should this be done another way?

source/beamline/AMDetector.cpp
helfrij: I have public setters already for setDarkCurrentValue(), setDarkCurrentTime(), and setDarkCurrentValidState()--they are the ones that perform the 'can do dark current?' test and then call the protected 'update' methods that actually perform the update. Is the problem that they should be squeezed into one method?

source/beamline/AMDetector.h
helfrij: I think it is technically used elsewhere: implemented in CLSBasicScalerChannelDetector and called in BioXASMainXASScanActionController.

source/actions3/actions/AMDetectorSetDarkCurrentValueActionInfo.cpp
dretrex: No that should work.  You could potentially do:
```AMActionInfo3(typeDescription(), typeDescription(), "some default icon", parent)```

source/ui/CLS/CLSSIS3820ScalerChannelViewWithDarkCurrent.cpp
helfrij: It's a good idea, especially if we want the scaler channel view to report a failed dark current measurement. I hadn't thought of it--what's the way to communicate this best?

source/beamline/AMDetector.cpp
dretrex: Ah, not really.  I didn't realize that one was public whereas the other was protected.  It should be fine this way.

source/beamline/AMDetector.h
dretrex: If it just returns 0 then I think we could probably remove the code that depends on it too then.

source/ui/CLS/CLSSIS3820ScalerChannelViewWithDarkCurrent.cpp
dretrex: Maybe have Failed and Cancelled call ```onDetectorAcquisistionFailed()```.

PR #1291, opened 2015-04-02
Pull Request for Issue1287: Create generic 1D and 2D scans for commissioning

dretrex:
Although this still needs to be tested, I'm going to put this up as a pull request now.  Aside from testing and adding things in as it is used more often I don't think there is necessarily anything missing.  If anyone thinks I am forgetting something, mention it here and I'll implement it.  

Also, everyone should feel free to leave comments.  Even if it is just asking for clarification on why something was implemented the way it was.


source/acquaman/AMGenericStepScanConfiguration.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+AMGenericStepScanConfiguration::AMGenericStepScanConfiguration(const AMGenericStepScanConfiguration &original)
+	: AMStepScanConfiguration(original)
+{
+	setName(original.name());
+	setUserScanName(original.name());
+
+	computeTotalTime();
+
+	for (int i = 0, size = scanAxes_.count(); i < size; i++){
```
XuetaoLiu-CLS: I would like it to be `for (int i = 0; i < scanAxes_.count(); i++)`. It will be the same in code efficiency.

source/acquaman/AMGenericStepScanConfiguration.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+		double size = fabs(double(scanAxes_.at(0)->regionAt(0)->regionEnd())-double(scanAxes_.at(0)->regionAt(0)->regionStart()));
+		int points = int((size)/double(scanAxes_.at(0)->regionAt(0)->regionStep()));
+
+		if ((size - (points + 0.01)*double(scanAxes_.at(0)->regionAt(0)->regionStep())) < 0)
+			points += 1;
+
+		else
+			points += 2;
+
```
XuetaoLiu-CLS: There is no calculation for `totalTime_`. Should we do it here? 
`totalTime_ = points *  (....->regionTime())`

source/acquaman/AMGenericStepScanConfiguration.cpp
davidChevrier: I'm pretty sure that @dretrex has this one right. Using his coding paradigm, the call to scanAxes_.count() is done only once (to assign the value of the variable size).
Using your suggestion the call to scanAxes_.count() should be executed for each iteration of the loop.

source/acquaman/AMGenericStepScanConfiguration.cpp
```diff
@@ -0,0 +1,187 @@
...
+		connect(scanAxisAt(0)->regionAt(0), SIGNAL(regionStartChanged(AMNumber)), this, SLOT(computeTotalTime()));
+		connect(scanAxisAt(0)->regionAt(0), SIGNAL(regionStepChanged(AMNumber)), this, SLOT(computeTotalTime()));
+		connect(scanAxisAt(0)->regionAt(0), SIGNAL(regionEndChanged(AMNumber)), this, SLOT(computeTotalTime()));
+		connect(scanAxisAt(0)->regionAt(0), SIGNAL(regionTimeChanged(AMNumber)), this, SLOT(computeTotalTime()));
+	}
+
+	else if (axisId == 0){
+
+		controls_.replace(0, newInfo);
```
XuetaoLiu-CLS: I'm not sure. Do we need to disconnect the signal/slot with the old control, and then reconnect with the `newInfo`?

source/acquaman/AMGenericStepScanConfiguration.cpp
dretrex: Hey @XuetaoLiu-CLS, good catch.

source/acquaman/AMGenericStepScanConfiguration.cpp
dretrex: Yes, this is done on purpose.  This is one of those trade offs where coding efficiency is traded for run time efficiency.  In general, I will _always_ go for run time efficiency because that is one of the limitations that we can actually run up against.

source/acquaman/AMGenericStepScanConfiguration.cpp
dretrex: No, we don't need to replace the controls because the controlInfo's are not being connected to anything.  All the signals are connected through the scan axes.

source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+#include "ui/AMTopFrame.h"
+#include "beamline/AMBeamline.h"
+
+AMGenericStepScanConfigurationView::AMGenericStepScanConfigurationView(AMGenericStepScanConfiguration *configuration, QWidget *parent)
+	: AMScanConfigurationView(parent)
+{
+	configuration_ = configuration;
+
+	AMTopFrame *frame = new AMTopFrame("BioXAS Commissioning Tool");
```
XuetaoLiu-CLS: Do we want this `BioXAS` String in the "AM..." class?

dretrex: In general, no.  However, I have given a task to @dleacock to make a convenience view constructor for the app controller that automatically builds top frames and the other necessary pieces.  We'll fix what you've tagged here when he finishes that.

source/acquaman/AMGenericStepScanConfiguration.cpp
XuetaoLiu-CLS: @davidChevrier @dretrex 
Yeah. You are right. `.count()` is executed for each loop.

Issue #1293, opened 2015-04-02
BioXAS: add BioXASSideTableCoordinator app


XuetaoLiu-CLS:
Add the BioXASSideTableCoordinator app to associate the PVs of the side table:
 - height / pitch / lateral / yaw
 - Vertical upstream/downstream pvs
 - Horizontal upstream/downstream pvs

XuetaoLiu-CLS:
duplicated with https://github.com/acquaman/acquaman/issues/1280

Issue #1280, opened 2015-03-30
Table coordination


dretrex:
Will need to implement a way to control the table since this will need to be used when scanning the energy on the SSRL mono's.  This will involve an IOC that maps physical motors to pitch, roll, and height.

XuetaoLiu-CLS:
For side table, there will: pitch, yaw, height, and lateral.

dretrex:
Done in #1294.  Time estimate?

XuetaoLiu-CLS:
took about 8 hrs

PR #1294, opened 2015-04-02
PullRequest for Issue1293: Add the BioXASSideTableCoordinator application

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1293
https://github.com/acquaman/acquaman/issues/1280

create the control logic between the new softIOC pvs: BL1607-
(horizontal) downstream  / upstream

@davidChevrier @dretrex @helfrij

### DIFF
--- a/source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
+++ b/source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
@@ -105,30 +105,30 @@ BioXASSideTableCoordinator::BioXASSideTableCoordinator(QObject *parent) :
 
 	connect(allControls_, SIGNAL(connected(bool)), this, SLOT(onAllControlsConnected(bool)));
 
-	connect(verticalUpstreamControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalControlValueChanged(double)));
-	connect(verticalUpstreamFeedbackControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalFeedbackControlValueChanged(double)));
-	connect(verticalUpstreamStatusControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalStatusControlValueChanged(double)));
+	connect(verticalUpstreamControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalControlValueChanged()));
+	connect(verticalUpstreamFeedbackControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalFeedbackControlValueChanged()));
+	connect(verticalUpstreamStatusControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalStatusControlValueChanged()));
 
-	connect(verticalDownstreamControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalControlValueChanged(double)));
-	connect(verticalDownstreamFeedbackControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalFeedbackControlValueChanged(double)));
-	connect(verticalDownstreamStatusControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalStatusControlValueChanged(double)));
+	connect(verticalDownstreamControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalControlValueChanged()));
+	connect(verticalDownstreamFeedbackControl_, SIGNAL(valueChanged(double)), this, SLOT(onVerticalFeedbackControlValueChanged()));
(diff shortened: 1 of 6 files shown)


source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
```diff
@@ -0,0 +1,414 @@
...
+	}
+	if (!softIOCPitchStatusControl_->withinTolerance(status)) {
+		softIOCPitchStatusControl_->move(status);
+	}
+}
+
+
+void BioXASSideTableCoordinator::onHorizontalControlValueChanged(double value)
+{
+	Q_UNUSED(value)
```
dretrex: All of these functions don't need the ```double value``` argument for the slot connection to work.  Slots can have less arguments than the signals that connect them.  I'd remove the value since and the Q_UNUSED for code simplicity.

source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+	// Status value: 0 - "MOVE DONE";  1 - "MOVE ACTIVE"; 2 - "AT LIMIT" ("MINOR"); 3 - "FORCED STOP" ("MINOR"); 4 - "ERROR" ("MAJOR")
+	// Status is "MOVE ACTIVE" if at least one motor is moving or the largest status value
+	double status = 0;
+	if (verticalUpstreamStatusControl_->withinTolerance(1) || verticalDownstreamStatusControl_->withinTolerance(1)) {
+		status = 1;
+	} else {
+		double verticalUpstreamControlStatus = verticalUpstreamStatusControl_->value();
+		double verticalDownstreamControlStatus = verticalDownstreamStatusControl_->value();
+		status = verticalUpstreamControlStatus > verticalDownstreamControlStatus ? verticalUpstreamControlStatus : verticalDownstreamControlStatus;
```
dretrex: What is the reasoning behind ```verticalUpstreamControlStatus > verticalDownstreamControlStatus```?

source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+void BioXASSideTableCoordinator::onYawControlValueChanged(double angle)
+{
+	Q_UNUSED(angle)
+	if(!connectedOnce_)
+		return;
+
+	double delta = tableHorizontalMotorPosition_ * tan(degreeToRadian(angle));
+
+	if (!horizontalUpstreamControl_->withinTolerance(-delta)) {
```
dretrex: This looks like a relative move.  This ```withinTolerance()``` call won't be checking what you think it does @XuetaoLiu-CLS.  Unless there is something else I'm missing.

source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
```diff
@@ -0,0 +1,414 @@
...
+		softIOCYawStopControl_->move(0); // restore the value to 0
+	}
+}
+
+void BioXASSideTableCoordinator::manipulateVerticalPVChange(double upstreamHeight, double downstreamHeight, AMControl* heightPV, AMControl *pitchPV)
+{
+	// The vertical height is the height of the center of the table, which is the average of upstream value and the downstream value
+	double height = (upstreamHeight + downstreamHeight) / 2;
+	// The pitch is the degree of the arcTangent value of (dH - uH) / distance
+	double pitch =  radianToDegree(atan((downstreamHeight - upstreamHeight) / (tableVerticalMotorPosition_ * 2)));
```
dretrex: Are these all relative movements?  Because I think you are going to have issues with these within tolerance checks.

source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
XuetaoLiu-CLS: @dretrex if the statuses of the individual pvs are not "move active", the status of the new PV should be the one with the larger value. For example:
if `upstream` is `move done`,  `at limit` or `force stopped`  but `downstream` is `error`, the status for the combined pv should be `error`.

source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
XuetaoLiu-CLS: as talked, this is not relative. I renamed the variable to avoid confusion.

source/beamline/BioXAS/BioXASSideTableCoordinator.cpp
XuetaoLiu-CLS: it is absolute move. the pitch is the current angle of the table.

PR #1296, opened 2015-04-06
Pull request for Issue1283 -- Updating CLSKeithley428 interface and exposing new features

helfrij:


source/ui/BioXAS/BioXASMainPersistentView.cpp
```diff
@@ -39,19 +40,47 @@ BioXASMainPersistentView::BioXASMainPersistentView(QWidget *parent) :
...
 }
 
 BioXASMainPersistentView::~BioXASMainPersistentView()
 {
 
 }
+
+void BioXASMainPersistentView::showScalerChannels(bool show)
+{
+	channelsView_->setShown(show);
```
dretrex: I think ```setShown()``` is an old Qt3 method.  I'd suggest changing this to ```setVisible()```.

source/ui/BioXAS/BioXASSIS3820ScalerChannelView.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+void BioXASSIS3820ScalerChannelView::setBiasEnabledEditorShown(bool shown)
+{
+	// if the scaler channel given in the constructor is invalid or doesn't have a Keithley amplifier,
+	// then the bias enabled editor will never have been instantiated. Must check whether it's possible
+	// to show the editor, as well as whether the 'shown' arg is different than the current state.
+
+	if (biasEnabledEditor_ && biasEnabledShown_ != shown) {
+		biasEnabledEditor_->setShown(biasEnabledShown_ = shown);
```
dretrex: All uses of ```setShown()``` should be replaced with calls to ```setVisible()```.

source/ui/BioXAS/BioXASSIS3820ScalerChannelView.h
```diff
@@ -0,0 +1,58 @@
...
+	/// Returns whether the bias enabled editor is shown.
+	bool biasEnabledShown() const { return biasEnabledShown_; }
+	/// Returns whether the bias editor is shown.
+	bool biasShown() const { return biasShown_; }
+	/// Returns whether the dark current is shown.
+	bool darkCurrentShown() const { return darkCurrentShown_; }
+
+public slots:
+	/// Sets whether the bias enabled editor is shown.
+	void setBiasEnabledEditorShown(bool shown);
```
dretrex: Okay, okay.  Broken record, I know.  To be consistent with existing code and the Qt API, using visible, show, and hide would be preferable.

source/ui/BioXAS/BioXASSIS3820ScalerChannelsView.h
```diff
@@ -0,0 +1,61 @@
...
+
+public:
+	/// Constructor.
+	explicit BioXASSIS3820ScalerChannelsView(CLSSIS3820Scaler *scaler, bool biasEnabledShown = false, bool biasShown = false, bool darkCurrentShown = false, QWidget *parent = 0);
+	/// Destructor.
+	virtual ~BioXASSIS3820ScalerChannelsView();
+	/// Returns the scaler being viewed.
+	CLSSIS3820Scaler* scaler() const { return scaler_; }
+	/// Returns whether the bias enabled editor is shown, for all able channels.
+	bool biasEnabledEditorShown() const { return biasEnabledEditorShown_; }
```
dretrex: Same "shown" comment here.

source/ui/BioXAS/BioXASSIS3820ScalerView.cpp
```diff
@@ -0,0 +1,133 @@
...
+	biasShown->setChecked(false);
+	darkCurrentShown->setChecked(false);
+
+	optionsBox_->hide();
+	channelsBox_->hide();
+	darkCurrentBox_->hide();
+
+	// Make connections.
+
+	connect( biasEnabledShown, SIGNAL(clicked(bool)), channelsView, SLOT(setBiasEnabledEditorShown(bool)) );
```
dretrex: Just as a note, ```clicked(bool)``` is only the same as ```toggled(bool)``` when buttons are checkable.  You may want to be explicit about your calls, just in case.  Not a big deal though.

source/ui/BioXAS/BioXASSIS3820ScalerView.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void BioXASSIS3820ScalerView::showOptions(bool shown)
+{
+	optionsBox_->setShown(shown);
+}
+
+void BioXASSIS3820ScalerView::showChannelViews(bool shown)
+{
+	if (scaler_ && scaler_->isConnected() && shown)
```
dretrex: You could simplify this code with ```setVisible(scaler_ && scaler_->isConnected() && shown)``` and it would achieve the same thing.  Just mentioning it for interests sake.

source/ui/CLS/CLSSIS3820ScalerControlsView.cpp
```diff
@@ -0,0 +1,246 @@
...
+		scaler_->blockSignals(false);
+	}
+}
+
+void CLSSIS3820ScalerControlsView::setDwellTime()
+{
+	if (scaler_) {
+		scaler_->blockSignals(true);
+
+		double secondsDwell = (double)time_->value() / MILLISECONDS_PER_SECOND;
```
dretrex: You could just make this 1000 since I don't think there is any disabiguation required for millisecond to second conversions.

source/ui/CLS/CLSSIS3820ScalerControlsView.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+		if (secondsDwell != scaler_->dwellTime())
```
dretrex: I'm just as guilty as the next person for this kind of thing, but I just want to use this as an opportunity.  Doing comparisons of ```double != double``` is not safe when doing floating point arithmetic.  The only reason why we are able to get around this in general is because all these values are set explicitly rather than calculated.

source/ui/CLS/CLSSIS3820ScalerControlsView.cpp
```diff
@@ -0,0 +1,246 @@
...
+			scaler_->setDwellTime(secondsDwell);
+
+		scaler_->blockSignals(false);
+	}
+}
+
+void CLSSIS3820ScalerControlsView::setScansPerBuffer()
+{
+	if (scaler_) {
+		scaler_->blockSignals(true);
```
dretrex: Why are these block signals here?  Usually the block signals are on the other side (when the model changes and we want to update the GUI elements) rather than when we set the model from the user interface.

dretrex:
Aside from the mostly aesthetic changes, these look good @helfrij.

helfrij:
@dretrex I think we've talked about all of the concerns you've brought up with this pull request. If there's nothing else, I think it's ready to merge.

helfrij:
@dretrex 
Completed a full build of Acquaman suite, no errors raised. Just tested the general scaler view on BioXASMain, and the look/functionality does not appear to have changed.

PR #1299, opened 2015-04-07
Pull Request for Issue1275: Controlling the reference wheel

dretrex:
I have made an implementation and simple configuration view for modifying the information in the standards wheel.  Unfortunately, there are the uses of the class are highly minimal at the moment so I've just pushed it up into a pull request for the time being.  Feel free to look over it now, but I'm going to likely add more to this later.


helfrij:
Looks great, @dretrex.

PR #1302, opened 2015-04-08
Pull Request for Issue1282: Implement Soller slits

dretrex:
I have implemented the soller slits calibration as a scan.  It's not finished and has not been tested yet, but the remaining aspects require soller slits that I can move, and a working Ge detector.  The last piece will be to have something determine the maximum value and return the position (or move it there automatically?)


dretrex:
This is going to be handled in a different way that will follow #1502 once it is completed.

Issue #1295, opened 2015-04-06
Simplify bragg calibration in BioXASSSRLMonochromator


helfrij:
@beaured0 discovered that there is actually a pv that will perform the calibration for us, if the motor steps calibration slope is non-zero: ...:deg:setPosn. This pv performs the same calculation as the one that's explicitly used now, but is simpler code-wise. The steps slope should never be zero, so that's not a worry.

dretrex:
Done in #1303.  Time estimate?

davidChevrier:
Time Estimate: 0 hours

PR #1303, opened 2015-04-09
PullRequest for issue1295: the :unit:setPosn pv for the CLSMAXvMotor can be used for calibration

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1295

- Expose the ":unit:setPostn" for CLSMAXvMotor so that we can use it for calibration purpose
- modify the BioXASSRLMonochromator to use the new MAXv pv for calibration

@dretrex @davidChevrier @helfrij @beaured0


dretrex:
Is that PV in all MAXv motors?  If it's not then this will cause a lot of issues for existing uses of CLSMAXvMotor.

XuetaoLiu-CLS:
@dretrex @beaured0 confirmed that that PV is in all MAXv motors.

PR #1306, opened 2015-04-09
Pull Request for Issue1192: BioXAS XRF scan controller reads data too early

dretrex:
A preliminary implementation of a system to accommodate the fact that acquisition finishes too early.  @davidChevrier, if you could go over it that would be great.  Due to how everything went, I'm worried that it's missing something really important to function at all.  I'll do tests on the actual detector tomorrow (hopefully).


davidChevrier:
@dretrex I'm guessing that there needs to be an initialized value for this in the constructor. I don't think I saw that. It's a clever way of handling the situation ... so I'm guessing that testing will either show that it works or it doesn't. I think that it's good to go for testing otherwise, let's see how it does.

source/application/BioXAS/BioXASSideAppController.cpp
```diff
@@ -221,10 +222,16 @@ void BioXASSideAppController::setupUserInterface()
...
 
 	// Create the 32 element Ge detector view.
-	BioXAS32ElementGeDetectorView *geDetectorView = new BioXAS32ElementGeDetectorView(BioXASSideBeamline::bioXAS()->ge32ElementDetector());
-	geDetectorView->buildDetectorView();
-	geDetectorView->addEmissionLineNameFilter(QRegExp("1"));
-	geDetectorView->addPileUpPeakNameFilter(QRegExp("(K.1|L.1|Ma1)"));
+//	BioXAS32ElementGeDetectorView *geDetectorView = new BioXAS32ElementGeDetectorView(BioXASSideBeamline::bioXAS()->ge32ElementDetector());
+//	geDetectorView->buildDetectorView();
+//	geDetectorView->addEmissionLineNameFilter(QRegExp("1"));
+//	geDetectorView->addPileUpPeakNameFilter(QRegExp("(K.1|L.1|Ma1)"));
```
dretrex: I'm leaving the 32 element code commented out for the time being until it's up and running.

PR #1307, opened 2015-04-09
Pull request for Issue1276: Incorporating Carbon Filter Farm

helfrij:
Created general controls for the actuator arms and the filter farm, with a Side beamline implementation. The status pvs for the actuator arms are not available, so no real testing is able to be performed yet.


source/beamline/BioXAS/BioXASCarbonFilterFarm.cpp
```diff
@@ -0,0 +1,395 @@
...
+}
+
+BioXASCarbonFilterFarmControl::~BioXASCarbonFilterFarmControl()
+{
+
+}
+
+bool BioXASCarbonFilterFarmControl::canStop() const
+{
+	bool result = false;
```
dretrex: Unless you think more logic is required later, you could replace this function with a ```return isConnected();``` statement.

source/beamline/BioXAS/BioXASCarbonFilterFarm.cpp
```diff
@@ -0,0 +1,395 @@
...
+}
+
+void BioXASCarbonFilterFarmControl::moveCleanup(QObject *action)
+{
+	if (action) {
+		disconnect( action, 0, cancelledMapper_, 0 );
+		disconnect( action, 0, failedMapper_, 0 );
+		disconnect( action, 0, succeededMapper_, 0 );
+
+		cancelledMapper_->removeMappings(action);
```
dretrex: Since there is only one action being used at any given instance, the signal mapper may be a bit overkill since you probably could replace it with a individual connect signals and just keep the action as a member variable.  It still works though and it does easily give you access to the pointer of the action once it's done - which is the important thing for the memory management aspect that you've already handled.

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorControl.cpp
```diff
@@ -0,0 +1,333 @@
...
+		emit movingChanged(moveInProgress_);
+
+		if (moveInProgress_)
+			emit moveStarted();
+	}
+}
+
+void BioXASCarbonFilterFarmActuatorControl::onMoveStarted()
+{
+	setMoveInProgress(true);
```
dretrex: Does anything call ```setMoveInProgress(false);``` in either of these controls?  Or is that handled by the superclass?

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorControl.cpp
```diff
@@ -0,0 +1,333 @@
...
+		}
+	}
+
+	return action;
+}
+
+void BioXASCarbonFilterFarmActuatorControl::moveCleanup(QObject *action)
+{
+	if (action) {
+		disconnect( action, 0, cancelledMapper_, 0 );
```
dretrex: You could also just call ```action->disconnect();```.  That will disconnect the object from everything in a single function call.

source/beamline/BioXAS/BioXASSideCarbonFilterFarmControl.cpp
```diff
@@ -0,0 +1,113 @@
...
+	default:
+		break;
+	}
+
+	return result;
+}
+
+AMAction3* BioXASSideCarbonFilterFarmControl::createMoveAction(double setpoint)
+{
+	AMListAction3 *action = new AMListAction3(new AMListActionInfo3("Moving filter farm actuators", "Moving filter farm actuators"), AMListAction3::Parallel);
```
dretrex: From what I can tell, this could be in the super class.  Nothing seems specific enough to require being in the subclass.

source/ui/BioXAS/BioXASSidePersistentView.cpp
```diff
@@ -64,11 +67,13 @@ BioXASSidePersistentView::BioXASSidePersistentView(QWidget *parent) :
...
 	layout->addWidget(regionControlEditor_);
 	layout->addWidget(braggControlEditor_);
+	layout->addWidget(carbonFilterFarmEditor_);
+	layout->addWidget(actuatorEditor);
 	layout->addWidget(channelViews_);
 	layout->addStretch();
 
 	setLayout(layout);
-	setFixedWidth(350);
+	setFixedWidth(355);
```
dretrex: Looks like the persistent view got a little thicker around midsection. ;)

dretrex:
Overall, this looks pretty good.

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorControl.cpp
helfrij: It should have been called in the moveCleanup() methods of both the actuator control and the general filter farm control, but was missing. Should be included now.

PR #1308, opened 2015-04-10
Pull request for Issue1285: Adding detector stage motor

helfrij:
Created new control for the detector stage motor in the Side beamline class and added it to the list of exposed controls. No real testing done, though throwing an AMExtendedControlEditor with the motor into the persistent view shows that the pv connects.

### DIFF
--- a/source/beamline/BioXAS/BioXASBeamlineDef.h
+++ b/source/beamline/BioXAS/BioXASBeamlineDef.h
@@ -20,8 +20,7 @@ namespace BioXASBeamlineDef {
 		PseudoM1Motor,
 		PseudoMaskMotor,
 		PseudoMonoMotor,
-		PseudoM2Motor,
-		DetectorStageMotor
+		PseudoM2Motor
 	};
 
 	/// check whether a given motor type is pseudo motor or not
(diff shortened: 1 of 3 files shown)


source/beamline/BioXAS/BioXASSideBeamline.cpp
```diff
@@ -573,6 +578,11 @@ void BioXASSideBeamline::setupSampleStage()
 
 }
 
+void BioXASSideBeamline::setupDetectorStage()
+{
+	detectorStageLateral_ = new AMPVwStatusControl("Lateral Motor Detector Stage", "SMTR1607-6-I22-16:mm:fbk", "SMTR1607-6-I22-16:mm:sp", "SMTR1607-6-I22-16:status", "SMTR1607-6-I22-16:stop", this, 0.001);
```
dretrex: Is this not a MAXv motor?

source/beamline/BioXAS/BioXASBeamlineDef.h
```diff
@@ -20,7 +20,8 @@ namespace BioXASBeamlineDef {
 		PseudoM1Motor,
 		PseudoMaskMotor,
 		PseudoMonoMotor,
-		PseudoM2Motor
+		PseudoM2Motor,
+		DetectorStageMotor
```
dretrex: This might not be specific enough (maybe it's not relevant to have it defined here either) because the Main endstation will have 2 detector stage motors and the imaging will have at least 1 (maybe two as well).

PR #1300, opened 2015-04-07
Pull Request for Issue1286: Create a BioXASScanConfigurationDbObject

dretrex:
Created a configuration database object.  Currently it holds onto the edge, edge energy, and the regions of interest.  More will be added as more components come online.

PR #1311, opened 2015-04-10
Pull Request for Issue1309: Create a user configuration for BioXAS

dretrex:
Added a user configuration that saves configuration details on a scan by scan basis for a user.  For the time being there is only the regions of interest for the 32 element detector.  It has commented out lines for adding the regions to the scan configuration because that is stored in the BioXASScanConfigurationDbObject which is handled in #1300 and hasn't been merged in yet.  Once it is, those lines can be uncommented and then merged in.

### DIFF
--- a/source/application/BioXAS/BioXASSideAppController.cpp
+++ b/source/application/BioXAS/BioXASSideAppController.cpp
@@ -270,7 +270,7 @@ void BioXASSideAppController::onUserConfigurationLoadedFromDb()
 
 		AMRegionOfInterest *newRegion = region->createCopy();
 		detector->addRegionOfInterest(newRegion);
-//		configuration_->addRegionOfInterest(region);
+		configuration_->addRegionOfInterest(region);
 	}
 
 	// This is connected here because we want to listen to the detectors for updates, but don't want to double add regions on startup.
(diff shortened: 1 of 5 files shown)


source/application/BioXAS/BioXASSideAppController.cpp
```diff
@@ -94,7 +97,15 @@ bool BioXASSideAppController::startup()
...
 		makeConnections();
-        applyCurrentSettings();
+		applyCurrentSettings();
+
+		if (!userConfiguration_->loadFromDb(AMDatabase::database("user"), 1)){
+
+			userConfiguration_->storeToDb(AMDatabase::database("user"));
+			// This is connected here because our standard way for these signal connections is to load from db first, which clearly won't happen on the first time.
+			connect(BioXASSideBeamline::bioXAS()->ge32ElementDetector(), SIGNAL(addedRegionOfInterest(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestAdded(AMRegionOfInterest*)));
+			connect(BioXASSideBeamline::bioXAS()->ge32ElementDetector(), SIGNAL(removedRegionOfInterest(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestRemoved(AMRegionOfInterest*)));
```
XuetaoLiu-CLS: I know this happened only on a fresh run. But I think it will be consistent if we can call `onUserConfigurationLoadedFromDb()` here. Then we won't need to worry about the same signal/slot connections in two different spots.

PR #1314, opened 2015-04-13
Pull  Request for Issue1313: Make a CSV parser

dretrex:
Added a CSV parser that handles commas inside of the comma deliniated cells.  This has no explicit purpose for AM in particular, but @davidChevrier and myself have had to write one when helping the user's office.

@davidChevrier, can you give this a quick look over to see if I may have missed something that you may have added to your implementation?


davidChevrier:
Looks good to me. It might suffer from newlines inside of escaping quotes, but I think that you can cross that bridge when you get to it.

Issue #1279, opened 2015-03-30
Integrate XIA filters


dretrex:
Requires mapping various "In/Out" positions to physical thicknesses.

helfrij:
- [ ] Create new control class for the device, similar to the carbon filter farm (#1276)
  - Map four individual filter controls to filter thicknesses.
- [ ] Create appropriate view.

helfrij:
Development on the AM interface will be stopped until we learn more about what filters will be installed and how the control will be used. A mental note for later: work done so far has taken about two days.

helfrij:
Spoke with the BioXAS scientists today and they requested a control view for the XIA filters that shows the In/Out status of each filter control, buttons to add/remove a given filter, and some kind of feedback for the total active filter thickness. We still don't know what the individual filter thicknesses are yet, but basic controls can be added.

dretrex:
Done in #1315.  Time estimate?

helfrij:
Probably about two days total.

PR #1315, opened 2015-04-13
Pull request Issue1279: Integrating XIA Filters

helfrij:


helfrij:
This branch now incorporates the BioXAS scientists' input from #1279: I've created basic QObject and QWidget subclasses that provide access to the XIA filter controls. Ready for review.

source/ui/BioXAS/BioXASCarbonFilterFarmView.cpp
```diff
@@ -0,0 +1,48 @@
...
+BioXASCarbonFilterFarmView::~BioXASCarbonFilterFarmView()
+{
+
+}
+
+void BioXASCarbonFilterFarmView::setFilterFarm(BioXASCarbonFilterFarmControl *newFilterFarm)
+{
+	if (filterFarm_ != newFilterFarm) {
+
+		if (filterFarm_) {
```
dretrex: Are you using this to delete the old editor before making the new one and then just remaking the layout with a new editor inside it?  If so, I'd recommend making the filterFarmEditor and making it a member as well as the layout.  Then you could just do,

```
farmLayout_->removeWidget(filterFarm_);
filterFarm_->deleteLater();
filterFarm_ = newFilterFarm;
farmLayout_->addWidget(filterFarm_);
```

source/ui/BioXAS/BioXASCarbonFilterFarmView.h
```diff
@@ -0,0 +1,33 @@
...
+
+class BioXASCarbonFilterFarmView : public QWidget
+{
+    Q_OBJECT
+
+public:
+	/// Constructor.
+	explicit BioXASCarbonFilterFarmView(BioXASCarbonFilterFarmControl *filterFarm, QWidget *parent = 0);
+	/// Destructor.
+	virtual ~BioXASCarbonFilterFarmView();
```
dretrex: For completeness, you may want to be able to return the control as well since you have a setter and signal for it already.

source/ui/BioXAS/BioXASXIAFiltersView.cpp
```diff
@@ -0,0 +1,54 @@
...
+BioXASXIAFiltersView::~BioXASXIAFiltersView()
+{
+
+}
+
+void BioXASXIAFiltersView::setFilters(BioXASXIAFilters *newFilters)
+{
+	if (filters_ != newFilters) {
+
+		if (filters_) {
```
dretrex: I think you can do the same thing here as the above comment.

source/ui/BioXAS/BioXASXIAFiltersView.h
```diff
@@ -0,0 +1,34 @@
...
+class BioXASXIAFiltersView : public QWidget
+{
+    Q_OBJECT
+
+public:
+	/// Constructor.
+	explicit BioXASXIAFiltersView(BioXASXIAFilters *filters, QWidget *parent = 0);
+	/// Destructor.
+	virtual ~BioXASXIAFiltersView();
+
```
dretrex: You can add a control getter here too.

helfrij: Added the control getter.

source/ui/BioXAS/BioXASCarbonFilterFarmView.h
helfrij: Added the control getter.

helfrij:
Added layout and editor member variables for both the BioXASXIAFiltersView and BioXASCarbonFilterFarmView. Was this more what you had in mind, @dretrex?

dretrex:
Yup, that's exactly what I meant.  I hope it also makes sense why I suggested it.  I'm assuming it still works as you intended as well?  If so, I can merge it in.

helfrij:
Yup, it makes sense and works as it should.

Issue #1248, opened 2015-03-18
Beef up setXYZControl flexibility in BioXASSSRLMonochromatorRegionControl.


helfrij:
Subclasses of the general BioXASSSRLMonochromatorRegionControl use beamline specific controls by setting a bunch of child controls in the subclass constructor (setUpperSlitControl(), setLowerSlitControl(), setPaddleStatusControl(), etc.). There are some improvements that can be made with how this is done:
- the child controls should be reparented to the ancestor class in each setXYZControl() body.
- old controls should be removed from the list of child controls when replaced with new controls.
- if old controls belong to the parent, they should be deleted when replaced with new controls.
- each setXYZControl() should be able to successfully handle null arguments.

EDIT:
- old controls should be disconnected, also. Thanks @dretrex.

dretrex:
And old controls should be disconnected.

dretrex:
Done in pull request #1317.  Time estimate?

helfrij:
Took about 30 mins.

PR #1317, opened 2015-04-14
Pull request for Issue1248: Adding better memory management to the setXYZControl() methods in the mono region control

helfrij:
Added two new protected methods to the general mono region control class, controlSetup() and controlCleanup(). These two methods basically handle adding/removing child controls from the list of children, manage connect statements, and reparenting new controls. They have been added to all of the setXYZControl() slots.

I think this branch addresses all suggestions brought up in the original issue #1248--ready for review.

### DIFF
--- a/source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
+++ b/source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
@@ -911,6 +911,23 @@ QString BioXASSSRLMonochromatorRegionControl::stepNotes(int stepIndex)
 	return result;
 }
 
+void BioXASSSRLMonochromatorRegionControl::controlSetup(AMControl *control)
+{
+	if (control) {
+		control->setParent(this);
+		addChildControl(control);
+	}
+}
+
+void BioXASSSRLMonochromatorRegionControl::controlCleanup(AMControl *control)
+{
+	if (control) {
+		control->disconnect();
+		children_.removeOne(control);
+		control->deleteLater();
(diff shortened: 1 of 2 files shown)


source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
```diff
@@ -170,148 +170,238 @@ AMControl::FailureExplanation BioXASSSRLMonochromatorRegionControl::move(double
...
+
+		if (crystalChangeCCWLimitStatus_) {
+			controlCleanup(crystalChangeCCWLimitStatus_);
+			crystalChangeCCWLimitStatus_ = 0;
+		}
+
 		crystalChangeCCWLimitStatus_ = limitStatus;
-		addChildControl(crystalChangeCWLimitStatus_);
 
-		connect( crystalChangeCCWLimitStatus_, SIGNAL(movingChanged(bool)), this, SIGNAL(movingChanged(bool)) );
```
dretrex: Is the reason for removing this connect statement because the composite control already handles it by adding it to the child controls?

helfrij: Yup! This connect statement (and it's parallels) should never have been there, as the composite control interface handles this behavior. I may have added them at one point to test something, but that test is no longer useful/relevant.

dretrex: Okay, great.  Looks good.

PR #1320, opened 2015-04-15
Pull request for Issue1319: adding m2 mirror fluorescence screen controls

helfrij:
<b>Summary:</b>
Created new general m2 mirror class with control for the fluorescence screen. Created new subclass for the Side beamline m2 mirror and added instance to BioXASSideBeamline. Added screen control to the list of exposed controls in the beamline class.

Created simple view for the m2 mirror, added it to the "General" tab in the Side app controller.

Added m2 mirror and m2 mirror view to BioXASMain.

Moved motors from beamline classes to the m2 mirror class, for Side and Main.


source/beamline/BioXAS/BioXASSideM2Mirror.cpp
```diff
@@ -0,0 +1,18 @@
+#include "BioXASSideM2Mirror.h"
+
+BioXASSideM2Mirror::BioXASSideM2Mirror(QObject *parent) :
+    BioXASM2Mirror(parent)
+{
+	// Initialize member variables.
+
+	screen_ = new AMPVControl("BioXAS Side M2 Mirror Screen Control", "VSC1607-5-I22-02:InBeam", "VSC1607-5-I22-02:InBeam", QString(), this);
```
dretrex: You could use AMSinglePVControl for succinctness.  It's purely aesthetic in that it is a convenience constructor for AMPVControl when there is only one PV for moving, feedback, and has no stop.

dretrex:
In terms of code, I'd say everything looks good.  

My only real comment is actually the choice of name for the screen encapsulation.  You call it a mirror but it is actually only the fluorescence screen.  I think you'll find that confusing when you look back on this code in the future since it actually doesn't control the mirror itself at all.

helfrij:
When I first started building BioXASM2Mirror and BioXASM2MirrorView classes, I had imagined that maybe it might be a good place to stick other m2 motors and controls. Right now, they live in the beamline class. Is this a good idea?

dretrex:
Encapsulating them inside the M2Mirror class is definitely okay and will make things easier to find and keep track of.  If you plan on migrating the actual mirror controls to this class then I have no issues with the naming at all.

helfrij:
Alright, I'll get on it. Thanks @dretrex.

helfrij:
The m2 mirror screen controls have been added to Side and Main AM, and the mirror controls have been migrated from the beamline class to the m2 mirror class.

dretrex:
Since the bulk of the latest changes you made were just moving the controls from the beamline into the mirror class, there isn't much to test right?

helfrij:
Yup, that's right. The beamline staff were already using the command line with the same pv names to move the screens; testing here would just need to include that using the control editor interface achieves the same effect.

dretrex:
Okay.  That's what I figured.  Since I'm confident that the changes should be fine, I'll merge this in and if there are any tweaks we'll deal with them as they arise.

source/application/BioXAS/BioXASMainAppController.cpp
```diff
@@ -225,9 +227,29 @@ void BioXASMainAppController::setupUserInterface()
...
+	QHBoxLayout *hSqueeze = new QHBoxLayout();
+	hSqueeze->addStretch();
+	hSqueeze->addWidget(m2MirrorView);
+	hSqueeze->addStretch();
+
+	QVBoxLayout *vSqueeze = new QVBoxLayout();
+	vSqueeze->addWidget(new AMTopFrame("M2 Mirror"));
+	vSqueeze->addStretch();
+	vSqueeze->addLayout(hSqueeze);
+	vSqueeze->addStretch();
```
XuetaoLiu-CLS: Dave created a common function for this, isn't it @dretrex ?

PR #1324, opened 2015-04-20
Pull request Issue1277: Incorporating DBHR mirror controls.

helfrij:


source/ui/BioXAS/BioXASDBHRMirrorView.cpp
```diff
@@ -0,0 +1,74 @@
...
+			m2Editor_ = 0;
+		}
+
+		mirror_ = newMirror;
+
+		if (mirror_) {
+
+			// create mirror editors, add to view layout.
+
+			pitchEditor_ = new AMExtendedControlEditor(mirror_->pitchControl());
```
dretrex: Hey @helfrij, since you've had to implement methods like this a couple times over the past few tasks, how would you feel about adding the AMControlEditor by having a ```setControl()``` method?  Since you've been having to remove and delete the widgets and then making them again seems like a bit of a waste.  Maybe make a new issue for adding to the functionality and then doing some code cleanup on these?

Issue #1338, opened 2015-04-24
SXRMB: Ambiant w/o gas chamber motor view dialog reponse to "Enter" key incorrectly


XuetaoLiu-CLS:
In Ambiant w/o gam chamber motor view dialog, when typing in a value to try to move the motor and clicking "enter" key, the screen will be closed.

XuetaoLiu-CLS:
Took about 1 ~ 1.5 hours to try different solutions

dretrex:
Done in #1341.

PR #1341, opened 2015-04-27
PullRequest for Issue1338: disable the enter key for Ambiant Motor view dialog

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1338

@davidChevrier @dretrex

### DIFF
--- a/source/application/SXRMB/SXRMBAppController.cpp
+++ b/source/application/SXRMB/SXRMBAppController.cpp
@@ -490,11 +490,15 @@ void SXRMBAppController::onShowAmbiantSampleStageMotorsTriggered()
 {
 	QString motorGroupName = SXRMBBeamline::sxrmb()->ambiantWithoutGasChamberSampleStageMotorGroupObject()->name();
 
-	AMMotorGroupView *motorGroupView = new AMMotorGroupView(SXRMBBeamline::sxrmb()->motorGroup(), AMMotorGroupView::Exclusive);
-	motorGroupView->setMotorGroupView(motorGroupName);
-	motorGroupView->showAvailableMotorGroupChoices(false);
+	if (!ambiantSampleStageMotorGroupView) {
+		ambiantSampleStageMotorGroupView = new AMMotorGroupView(SXRMBBeamline::sxrmb()->motorGroup(), AMMotorGroupView::Exclusive);
+		ambiantSampleStageMotorGroupView->setMotorGroupView(motorGroupName);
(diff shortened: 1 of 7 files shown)


dretrex:
@XuetaoLiu-CLS, your changes from #1342 is also inside this pull request.

source/ui/util/AMDialog.cpp
```diff
@@ -37,6 +41,24 @@ void AMDialog::layoutDialogButtons()
...
+void AMDialog::setEnableDialogEnterKeyBehavior(bool enable)
+{
+	enableDefaultEnterBehavior_ = enable;
+}
+
+void AMDialog::keyPressEvent(QKeyEvent *e)
+{
+	switch (e->key ()) {
+	case Qt::Key_Return:
+	case Qt::Key_Enter:
```
dretrex: I don't think this is the way you want to use this.  This negates any use of the enter key not just closing the dialog.  What you likely need to do is change the default key of the dialog to not trigger the accept() flag.

dretrex:
Like I mentioned in the meeting, it may also be better to treat the popup as a widget rather than a dialog.  Then there is no implied default behaviour.

XuetaoLiu-CLS:
@dretrex Daren, you can review this now. The changes for AMDialog is reverted and I'm using the MotorView directly. It looks good to me.

XuetaoLiu-CLS:
tested

source/application/SXRMB/SXRMBAppController.cpp
```diff
@@ -494,9 +494,7 @@ void SXRMBAppController::onShowAmbiantSampleStageMotorsTriggered()
 	motorGroupView->setMotorGroupView(motorGroupName);
 	motorGroupView->showAvailableMotorGroupChoices(false);
 
-	AMDialog *showMotorDialog = new AMDialog(motorGroupName);
-	showMotorDialog->layoutDialogContent(motorGroupView);
-	showMotorDialog->exec();
+	motorGroupView->show();
```
dretrex: @XuetaoLiu-CLS, you would be better off using ```QWidget::raise()``` rather than ```show()```.  If you only use show the window can get lost and if you trigger the view again it will still be hidden behind other windows.  Using raise will always bring it to the forefront.

XuetaoLiu-CLS: @dretrex Thanks for the suggest. It is solved

Issue #1355, opened 2015-04-30
SXRMB: when crystal change completed, auto move the the crystal theta and chi PV


XuetaoLiu-CLS:
The Beamline staff asked a feature for crystal selection:
When the crystal selection is completed, it will be nicer if we can move the crystal theta pv and chi pv to preset values.

XuetaoLiu-CLS:
Took about 15 mins

dretrex:
Done in #1356.

PR #1356, opened 2015-04-30
PullRequest for Issue1355: Change the value of crystal Theta PV and the crystal Chi PV when crystal is changed

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1355

When crystal is changed, we move the crystal Theta and the crystal Chi to a given position for the convenience of the beamline staff

@dretrex @davidChevrier

### DIFF
--- a/source/acquaman/BioXAS/BioXASMainXASScanActionController.cpp
+++ b/source/acquaman/BioXAS/BioXASMainXASScanActionController.cpp
@@ -30,28 +30,29 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 #include "beamline/AMBasicControlDetectorEmulator.h"
 #include "util/AMErrorMonitor.h"
 #include "analysis/AM1DExpressionAB.h"
+#include "analysis/AM1DDerivativeAB.h"
 #include "dataman/export/AMExporterOptionGeneralAscii.h"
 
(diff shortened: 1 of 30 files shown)


XuetaoLiu-CLS:
Test done

source/beamline/AMPVControl.cpp
```diff
@@ -213,7 +213,7 @@ AMControl::FailureExplanation AMPVControl::move(double setpoint) {
 void AMPVControl::onNewFeedbackValue(double) {
 
 	// If we're not in the middle of a move, don't really care about changing values.
-	if(!moveInProgress_)
+	if(!moveInProgress_ && !completionTimer_.isActive())
```
dretrex: Why is this necessary?

XuetaoLiu-CLS: @dretrex for the crystal selection, it needs a long to move. The value of feedback will change twice during the process. When the final value is changed, the `moveInProgress_` is False already. But we still have the `completionTimer_` active (60s timout for this case). 

With this check, we will be able to have the complete signal ASAP. This is our original design as well (in the codes below, we stopped the timer. But it will only works when the move is quick enough).

dretrex: There is another issue if your crystal selection move is taking too long.  We need to think about how these changes can affect other implementations and a general change to AMPVControl for a single exception does not seem like a good idea to me.

PR #1364, opened 2015-05-05
PullRequest for Issue1357: add an iteration spinbox for the AMActionRunnerAddActionBar3

XuetaoLiu-CLS:
As discussed, we will add an iteration spinbox to the AMActionRunnerAddActionBar3 so that we can add a group of same actions at the same time.

@dretrex @davidChevrier @helfrij


dretrex:
Hey @XuetaoLiu-CLS, this code actually exists already.  At the moment there are two examples in VESPERSScanConfigurationViewHolder3 and IDEASScanConfigurationViewHolder3.  Since they were originally one off solutions this was acceptable, but maybe this is the time to try and make this more general?  The nice thing is that by inserting the scans into a loop action, the workflow and all affiliated views/models update and take note of the changes in iteration by default.

Sorry that you did this work when it was already done.

source/acquaman/IDEAS/IDEASXASScanConfiguration.h
```diff
@@ -31,7 +31,6 @@ class IDEASXASScanConfiguration : public AMStepScanConfiguration
 	Q_PROPERTY(QString edge READ edge WRITE setEdge)
 	Q_PROPERTY(double edgeEnergy READ energy WRITE setEnergy)
 	Q_PROPERTY(bool useFixedTime READ useFixedTime WRITE setUseFixedTime)
-	Q_PROPERTY(int numberOfScans READ numberOfScans WRITE setNumberOfScans)
```
XuetaoLiu-CLS: @dretrex I'm not sure about this changes. It should affect the database. But we really don't need it anymore.

source/acquaman/VESPERS/VESPERSEXAFSScanConfiguration.h
```diff
@@ -52,10 +52,8 @@ class VESPERSEXAFSScanConfiguration : public AMStepScanConfiguration, public VES
 	Q_PROPERTY(double yPosition READ y WRITE setY)
 	Q_PROPERTY(QString header READ headerText WRITE setHeaderText)
 	Q_PROPERTY(bool useFixedTime READ useFixedTime WRITE setUseFixedTime)
-	Q_PROPERTY(int numberOfScans READ numberOfScans WRITE setNumberOfScans)
```
XuetaoLiu-CLS: same question here

source/acquaman/IDEAS/IDEASXASScanConfiguration.h
dretrex: It does affect the database, but not in a way that's necessarily bad.  What this will do is "orphan" that column in the database.  The proper "robust and complete" solution would actually involve a database upgrade where the column is deleted from the necessary tables, but that is not necessary in this case.

dretrex:
@XuetaoLiu-CLS, the issue I see with this this is that you are now putting _every_ scan inside of a loop.  In general, it is only the XAS scans that will be put inside loops.  If you want to go with this solution, I would make another scan configuration view holder specifically for XAS scans.

XuetaoLiu-CLS:
@dretrex can you review this? I added a flag, now only  XAS scans will have the iteration spinbox and loop action

dretrex:
@XuetaoLiu-CLS, this looks alright.  However, I would like it if you could change the names of the bool, createLoopAction() and createAction().  They aren't very descriptive and will probably confuse us more than anything if we look back on this code in the future.  Maybe something along the lines of _repeatScans_, _createScan()_, and _createMultipleScans()_?

Also, @davidChevrier, could you go over this too?

PR #1365, opened 2015-05-06
Pull Request for Issue1255: Detector emulator default visibility

dretrex:
Altered the default state of the detector emulators to be not visible and hidden from users.

I also changed instatiations (mostly in VESPERS) to reflect the change. Although BioXAS uses it as well, they are still in commissioning stages and don't want to hide away some of this data yet.


iainworkman:
I noticed that in VESPERS2DScanActionController's constructor the visible status is set to the default values of visible(false) and hidden(true) [here](https://github.com/acquaman/acquaman/blob/master/source/acquaman/VESPERS/VESPERS2DScanActionController.cpp#L135-L151) and then in the cleanup operations set back to what used to be the defaults [here](https://github.com/acquaman/acquaman/blob/master/source/acquaman/VESPERS/VESPERS2DScanActionController.cpp#L135-L136).

The same thing is also done on the Bruker detector in the SXRMB2DScanActionController [here](https://github.com/acquaman/acquaman/blob/master/source/acquaman/SXRMB/SXRMB2DScanActionController.cpp#L63-L64) and [here](https://github.com/acquaman/acquaman/blob/master/source/acquaman/SXRMB/SXRMB2DScanActionController.cpp#L143-L144) 

Will these leave mean these detectors are visible(false) and hidden(true) on construction, then visible(true) and hidden(false) after each of these scans is run?

dretrex:
@iainworkman, the difference between that one and what I've done here is that the single element vortex detector (note: not detector emulator) is visible by default because in 1D scans it is completely reasonable to look at its output.

iainworkman:
Gotcha. Other than that everything else looks fine.

Issue #1367, opened 2015-05-06
SXRMB: add stop PV for Ambiant Table Height


XuetaoLiu-CLS:
- [x] Add the BL1606-B1-1:AddOns:Ambiant:TableHeight:stop PV
- [x] Modify the SXRMBAddOnsCoordinator
- [x] Modify the SXRMB Ambiant Motor view to add a stop button
- [ ] Deploy softIOC db
- [ ] Deploy AddOns coordinator
- [ ] rerun the softIOC app and the coordinator app

dretrex:
Done in #1368.  Time estimate?

XuetaoLiu-CLS:
took about 1.5 hrs
need test

PR #1368, opened 2015-05-07
PullRequest for Issue1367: add stop PV for Ambiant Table Height

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1367

- Add a stop PV to Ambiant Table Height Pseudo motor;
- Add a stop button to the table Height control edit

### DIFF
--- a/source/beamline/SXRMB/SXRMBBeamline.cpp
+++ b/source/beamline/SXRMB/SXRMBBeamline.cpp
@@ -642,19 +642,22 @@ void SXRMBBeamline::setupSampleStage()
 	solidStateSampleStageControlSet_->addControl(solidStateSampleStageZ_);
 	solidStateSampleStageControlSet_->addControl(solidStateSampleStageR_);
 
-	// Ambiant Endstation sample state
+	// Ambiant Endstation sample stage
+	ambiantTableHeight_ = new AMPVwStatusControl("Ambiant Table Height", "TBL0000-E07-01:Y:mm:fbk", "TBL0000-E07-01:Y:mm", "TBL0000-E07-01:Y:mm:status", QString(), this, 0.005, 2.0, new CLSMAXvControlStatusChecker());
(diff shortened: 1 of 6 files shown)


source/beamline/SXRMB/SXRMBBeamline.cpp
```diff
@@ -642,19 +642,22 @@ void SXRMBBeamline::setupSampleStage()
 	solidStateSampleStageControlSet_->addControl(solidStateSampleStageZ_);
 	solidStateSampleStageControlSet_->addControl(solidStateSampleStageR_);
 
-	// Ambiant Endstation sample state
+	// Ambiant Endstation sample stage
+	ambiantTableHeight_ = new AMPVwStatusControl("Ambiant Table Height", "TBL0000-E07-01:Y:mm:fbk", "TBL0000-E07-01:Y:mm", "TBL0000-E07-01:Y:mm:status", "BL1606-B1-1:AddOns:Ambiant:TableHeight:stop", this, 0.005, 2.0, new CLSMAXvControlStatusChecker());
```
dretrex: @XuetaoLiu-CLS, can you change the name of this PV?  The Y direction is always used as the beam direction in beamlines.  Could you rename it to either Z or H?

dretrex: Maybe even be explicit and use Height?

XuetaoLiu-CLS: @dretrex `TBL0000-E07-01:Y` is the existing pseudo PV for height control. I only defined a `Stop` pv for table height.

XuetaoLiu-CLS: It's a lazy solution. I don't have to do the logic of the table height logic.

dretrex: Really?  That's disappointing. If that's how it has been defined then it's okay to leave it as is and remain consistent.

dretrex: Which is a lazy solution?  Sorry I missed your last message while I was writing out my last one.

XuetaoLiu-CLS: I mean that I didn't create another set of pvs for the table heights. I just borrowed the PVs we have now and only created the stop PV.

dretrex: That's fine.  It might still be something we want to tell Tonia so that stopping the height can be done outside of the main application.

PR #1369, opened 2015-05-07
Pull Request for Issue1348: Cleanup actions are not called when scans are cancelled.

dretrex:
Made the scan action controller a little more robust rather than just adding in the cancelled aspect of the state machine cleanup.  @davidChevrier could you go over this for me?


davidChevrier:
Looks good to me @dretrex 

I guess the real test, as it always is for the scan controllers, is live action. Let me know how it goes.

dretrex:
Testing on BioXAS appears to have the desired new effect with cancel without destroying normal operations.

Issue #1214, opened 2015-02-27
Refactoring the Dataview and associated classes


iainworkman:
[Pull request #1372]
There are still a few issues with the dataview, namely double clicking and thumbnail indexing. When looking at this issues it tends to come to light that they stem from a set of general smells, especially from the loading of data and the updating of data when signals are emitted from the database object. In light of this, rather than attempt to re-factor it piecemeal in order to solve a few issues separately, I'll just look at the whole shebang at once.

As such I'm going to close:
#909 - Drag & Drop Unavailable
#1213 - Double Click Behaviour
#1179 - Bad Dynamic Indexing of Thumbnails
...as the re-factor will aim at sorting out these problems.

iainworkman:
Current issues that I can see, in general terms:
- The AMLightweightScanInfoCollection does ''A LOT''. It builds the collection from the database initially, it listens to all the changes coming in and updates its data based on these changes, and ultimately it stores all the scan infos. Really it should just do the last. A factory which does the initial bulk data load might be a good idea, but doesn't solve the issue of changes to the data while the system is running.
- Oh, Thumbnails. Why can't you just be good? I need to figure out either a robust way to ensure that the data in the collection is maintained for the thumbnails correctly, or radically alter the way they're stored.
- AMScanThumbnailView is still too big. I would suggest a single thumbnail class, an AMGridLayout which handles the layout stuff and then the AMScanThumbnailView itself.

davidChevrier:
Thanks for looking into this and organizing it @iainworkman 
I definitely think you're on the right track here, let me know if there's anything I can do to help.

iainworkman:
Changed mapping of AMLightweightScanInfos and Experiments to be *:* instead of 1:1

91b2f850c2b02d0c053d4a4ca1acb48139b4f748

iainworkman:
Started on moving the loading of data from the database from AMLightweightScanInfoCollection to AMLightweightScanInfoFactory. At the moment just the initial bulk loading is working, the data reacting to updates in the database which live are still to be done.

d9da4ead0ddfb946221e0d094c8849f46e913023

iainworkman:
Altered: 
- AMLightweightScanInfoFactory ~ Now properly implements loading of a single ScanInfo, as well as updated a passed ScanInfo reference to have the same information which is currently stored against it in the database. 
- AMLightweightScanInfoCollection ~ Now updates/adds scan info to the collection when an update/insert notification is received from the database on the AMScan_table

[...]

PR #1372, opened 2015-05-07
Pull Request for Issue1214: Refactoring the DataView classes

iainworkman:
#1214
## Work Done:
- Drag and Drop now works on all the views
- Runs are properly added to the system if they are created while the application is running
- Double click behaviour is now consistent (the timer has been reduced a little and two consecutive clicks on different scans are never interpreted as a double click)
- The LightweightScanInfo* classes were reworked and a few added to separate the different tasks that were being performed
- The scan thumbnail view was rewritten to separate a bunch of tasks that it was doing, and to simplify each one

## Testing done:
- I've verified that the interactions behave consistently
- I've verified that adding or removing an analysis tool to a scan correctly updates the thumbnails
- I've verified that adding a run while the system is running correctly results in the run being displayed in the filter widget

## Testing which I can't do on SGM at the moment:
- Can't test the a new scan gets added properly to the view

### NOTE
Once this is all done, tested and the dust has settled the replaced view class (AMScanThumbnailView) can be removed, as it has entirely been replaced by AMScanThumbnailGridView.


source/dataman/AMLightweightScanInfoCollection.cpp
```diff
@@ -43,512 +44,84 @@ int AMLightweightScanInfoCollection::indexOf(AMLightweightScanInfo *scan, int fr
...
-			{
-				lightweightScanInfoIndex = iScanInfo;
-				matchedScanInfo = scanInfos_.at(iScanInfo);
-				lightweightScanInfoFound = true;
+	if(tableName == AMDbObjectSupport::s()->tableNameForClass("AMScan")) {
+		int dirtyScanInfoIndex = indexOfScanWithId(id);
+		if(dirtyScanInfoIndex >= 0) {
+			AMLightweightScanInfo* dirtyScanInfo = scanInfos_.at(dirtyScanInfoIndex);
+			if(dirtyScanInfo->thumbnailCount() > 0) {
+				emit scanThumbnailAboutToBeRemoved(dirtyScanInfoIndex, 0, dirtyScanInfo->thumbnailCount()-1);
```
dretrex: Why is the scanThumbnailAboutToBeRemoved emitted when it appears like it's only being updated?

source/dataman/AMLightweightScanInfoModel.h
```diff
@@ -40,9 +39,19 @@ class AMLightweightScanInfoModel : public QAbstractItemModel
...
+	/// Returns the flags for the items. In cases where the index is valid and has no
+	/// valid parent (ie is for a scan, not a data source) then the flags are default
+	/// with DragEnabled, else they are just the default flags obtains from QAbstractItemModel::flags()
+	Qt::ItemFlags flags(const QModelIndex& index) const;
+	/// Returns the data for drag operations which will be passed outside of the views. In this case
+	/// it is a list of QUrls which indicate where the Scans can be loaded from.
+	QMimeData* mimeData(const QModelIndexList &indexes) const;
+	/// Whether or not the scan with the provided Index belongs to the experiment with the
+	/// provided id.
+	bool belongsToExperiment(const QModelIndex index, int experimentId);
```
dretrex: This would be a ```const``` function.  I think there are a couple more too.  Maybe go over all the headers and add const to the ones that need them.

source/ui/dataman/AMScanThumbnailGridInputManager.cpp
```diff
@@ -0,0 +1,221 @@
...
+				doubleClickTimer_.start(DOUBLE_CLICK_INTERVAL_MS);
+			}
+		}
+	} else if(!mouseButtons & Qt::RightButton){
+		// Right button clicked
+
+		if(dragInProgress_) {
+			// Drag is in progress need to cancel it
+		} else {
+			int travelDistanceX = int(std::abs(double(positionX - startPositionX_)));
```
dretrex: You don't need to change it necessarily.  But in general, use qAbs or fabs.  Just so we are consistently using the same math tools.

source/ui/dataman/AMScanThumbnailGridView.cpp
```diff
@@ -0,0 +1,422 @@
...
+void AMScanThumbnailGridView::onItemDoubleClicked(int itemIndex)
+{
+	QModelIndex doubleClickedItemIndex = model()->index(itemIndex, 0, QModelIndex());
+	emit doubleClicked(doubleClickedItemIndex);
+}
+
+void AMScanThumbnailGridView::onDragBegun()
+{
+	QDrag* drag = new QDrag(this);
+	QMimeData* mimeData = model()->mimeData(selectionModel()->selectedIndexes());
```
dretrex: Does drag handle the memory of mimeData?  If not, this is a memory leak.

source/ui/dataman/AMScanThumbnailGridView.h
```diff
@@ -0,0 +1,209 @@
...
+	 * Creates a new instance of an AMScanThumbnailGridView which has no attached
+	 * model
+	 * @param parent::QObject ~ The parent QObject
+	 */
+	explicit AMScanThumbnailGridView(QWidget *parent = 0);
+
+	/**
+	  * Cleans up resources owned by the AMScanThumbnailGridView
+	  */
+	~AMScanThumbnailGridView();
```
dretrex: Virtual destructor missing here.  Could you please double check to make sure all the destructors are virtual?

source/ui/dataman/AMScanThumbnailGridViewItemDelegate.cpp
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	painter->restore();
+}
+
+void AMScanThumbnailGridViewItemDelegate::paintThumbnail(QPainter *painter,
+															  const QStyleOptionViewItem &option,
+															  const QPixmap &pixmap) const
+{
+	painter->drawPixmap(option.rect, pixmap);
```
dretrex: This is more of a curiosity question.  Does this method draw the pixmap within the confines of the provided QRect or does it draw what it can be chop extra or have dead space if the pixmap is smaller than the rect?

source/ui/dataman/AMScanThumbnailGridViewItemDelegate.h
```diff
@@ -0,0 +1,57 @@
...
+{
+	Q_OBJECT
+public:
+	/**
+	 * Creates an instance of an AMScanThumbnailGridViewItemDelegate
+	 * @param parent
+	 */
+	explicit AMScanThumbnailGridViewItemDelegate(QObject *parent = 0);
+
+	/// Paints the data associated with the provided index according to options stored in the given
```
dretrex: No destructor.

source/ui/util/AMGridFlowGeometryManager.h
```diff
@@ -0,0 +1,241 @@
...
+	QRect managedGeometry() const;
+
+	/**
+	 * Helper method which calculates the content geometry for an item within
+	 * a cell based on the current cellContentAlignment, the cell size and the
+	 * item size.
+	 */
+	QRect contentGeometryFromCell(const QRect& cellGeometry) const;
+AMScanTreeView
+private:
```
dretrex: Is there a specific reason why these are  private rather than protected?

dretrex:
In general, this looks good.  @iainworkman, can you remove the old code (for completeness since we don't need it for backwards compatibility) and deal with the merge conflicts?  Once you do that I'll do another quick look over again and we can merge it in.

source/dataman/AMLightweightScanInfoCollection.cpp
iainworkman: This captures the case when a scan is saved having had a change made which will affect the thumbnails. Because they're all removed from the database and then re-added we go ahead here and refresh the whole bunch. It's a little less efficient to prompt a redraw of the thumbnail here each time the scan is saved (which is what the signal eventually triggers) - but it is significantly cleaner and less complex than the previous work around which attempted to listen to changes to the thumbnail table and then in some way figure out which scan that related to (as we only receive a id of -1 in that case)

source/ui/dataman/AMScanThumbnailGridView.cpp
iainworkman: The drag operation takes parentage of the mimeData when drag->setMimeData(mimeData) is called. So long as the drag is constructed with a QWidget then the cleanup happens (according to the documentation).

source/ui/dataman/AMScanThumbnailGridViewItemDelegate.cpp
iainworkman: When I altered the size values in the geometry manager it scaled the pixmaps to fit the rectangles. This wasn't ideal, as scaling the thumbnails of 1D plots more than a very small amount produced gaps in the lines. I ended up specifying a size in the geometry manager that didn't scale them enough that this effect occurred.

source/ui/dataman/AMScanThumbnailGridViewItemDelegate.h
iainworkman: With there being no resources owned by the delegate I just left it up to the default destructor. I can add an explicit destructor in here if required.

source/ui/dataman/AMScanThumbnailGridInputManager.cpp
iainworkman: qAbs is much better, as its a templated function which means I don't have to the that ghastly cast to double then cast to int.

iainworkman:
Thanks. I've pushed some changes and will run a test build on it to make sure everything’s alright when I'm next in. Once that's done I'll fix up the conflict and merge it in.

source/ui/dataman/AMScanThumbnailGridViewItemDelegate.h
iainworkman: Mentioned this to David this morning and he added that due to this being possibly involved in an inheritance chain in the future it will need to explicitly specify a virtual destructor. In future I'll ensure all such classes have explicit virtual destructors even in cases where they're not owning resources.

source/ui/dataman/AMScanThumbnailGridView.cpp
dretrex: Ah, okay.  Great.

dretrex:
I think I outlined a couple places where you could add the const qualifier to functions.  Could you add those in as well before we merge it in?

iainworkman:
I added those to all that I could see (will go through and have another double check). There were a few in the factory class that can't have the const qualifier as they produced a compile error (though it wasn't entirely clear why - none of these seem to call non const methods or otherwise alter data in the this pointer).

dretrex:
It might be due to the fact that they are static methods.

Sent from my iPhone

On May 11, 2015, at 11:37 AM, iainworkman <notifications@github.com<mailto:notifications@github.com>> wrote:


I added those to all that I could see (will go through and have another double check). There were a few in the factory class that can't have the const qualifier as they produced a compile error (though it wasn't entirely clear why - none of these seem to call non const methods or otherwise alter data in the this pointer).

iainworkman:
Derp. Missed that as I was looking at the .cpp file for the answer, not the .h

iainworkman:
Okay, found one more function which can have the const qualifier. Added it and pushed. This should now be ready for review @dretrex.

dretrex:
Looks good @iainworkman.

PR #1383, opened 2015-05-12
Pull Request for Issue813: Refactoring REIXSAppController

iainworkman:
Structured the class to be more like VESPERS, IDEAS, BIOXAS and SXRMB. I've tested it on my OPI down on SGM (with an SGM style Acquaman.ini - so slightly different setup) and it runs okay. Need to pull it down for testing on REIXS to make sure nothing is out of the ordinary.

@epengr @XuetaoLiu-CLS @dretrex


source/application/REIXS/REIXSAppController.cpp
```diff
@@ -78,85 +78,134 @@ REIXSAppController::REIXSAppController(QObject *parent) :
...
 
-	// Initialize the central beamline object
-	REIXSBeamline::bl();
-	// Initialize the storage ring.
-	CLSStorageRing::sr1();
+void REIXSAppController::shutdown() {
 
-	connect(REIXSBeamline::bl()->spectrometer(), SIGNAL(connected(bool)), REIXSBeamline::bl()->spectrometer(), SLOT(updateGrating()));
+	// Make sure we release/clean-up the beamline interface
+	AMBeamline::releaseBl();
```
XuetaoLiu-CLS: You might want to call `AMStorageRing::releaseStorageRing()` here as well.

iainworkman: I'll add that line. I notice that a lot of the other beamlines don't release the storage ring in their app controller's shutdown function. Is this something that should be added to the others too?
@dretrex

iainworkman: Also, if this is the case across the board it's probably something that could be added to AMAppController::shutdown() instead of individually to each subclass.

XuetaoLiu-CLS: @iainworkman actually I preferred to add to `AMAppController::shutdown()` as well. But I forgot why I didn't do that when I noticed this :hushed:

dretrex: Hmm, interesting.  I think this must have slipped through the cracks because beamlines instantiate CLSStorageRing and CLS classes don't belong in AM classes (like AMAppController).  However, since the releaseStorageRing method is general anyway, you should be able to use ```AMStorageRing::releaseStorageRing()``` from ```AMAppController::shutdown()```.

iainworkman: Okay, will do.

-----Original Message-----

dretrex:
What is the testing that is required for this issue?

iainworkman:
Well all I did here was run the REIXS version of acquaman, which doesn't really connect me to a lot of stuff. I was thinking that basically all I'd need to do is run it on there just to make sure there wasn't anything related to signals coming from detectors that got missed off.

dretrex:
Did you manage to catch up with @epengr to do this test?

iainworkman:
Not yet. I've got a couple of issues that need testing on REIXS (and one on ideas). I'll try and track him down now.

iainworkman:
Testing on this revealed that there were some tweaks the I needed to do regarding the chooseDataFolderDialog, that can be solved pretty easily. There were also some issues with the fact that REIXS is in a state where it is semi transferred between the two, and it was somewhat difficult to determine if issues were being caused by the network synchronization system - or whether it was because of changes to the app controller. I will continue with tweaks.

dretrex:
Hey @iainworkman, what is left for this issue?  I'd like to get it wrapped up before we forget about it and go full steam ahead on SGM tasks.

iainworkman:
Yeah, this just needs a run of testing on REIXS. This should just be a case making sure the start up procedure is as expected. I'll speak with @epengr about getting a bit of time on REIXS today.

iainworkman:
Okay, this has been tested on REIXS and seems to be working as expected.

Issue #1177, opened 2015-02-01
IDEAS XRF acquire disable 


epengr:
IDEAS XRF panel acquire button was set to disable when a scan complete and enable when is is saved. It isn't doing this now.

iainworkman:
Okay, the changes I made in c174407 should, theoretically, solve this - though I won't know till I've tried it out on IDEAS - will create a pull request at that point. 

I do have a few conceptual questions about this class, however, and how it's used to conduct XRF scans on IDEAS. If I'm understanding the logic correctly, this allows a user to start an acquisition over a certain dwell period - and then once this has acquired successfully allows the user to ... repeat the same acquisition this time as an actual scan? I think the name of the Save button was confusing me a bit if this is the case, because you're not actually saving what has just been acquired, no? Of course, I might just have the total wrong end of the stick here.

dretrex:
You can make a pull request now and just put the tag "awaiting testing".

epengr:
The save button doesn't trigger another acquisition, it just creates a "scan" that instantly completes and stores the data currently in the buffer.

The act of acquiring (filling the buffer), disables the acquire button so that the user can't inadvertently start a new acquisition (over-writing the buffer) until the data currently in the buffer is saved.  So, the act of saving re-enables the acquire button.  The stop button also re-enables the acquire button so that aborted scans don't prevent you from starting a new one, and it gives the user a way to choose not to save the scan if they don't want to.

iainworkman:
Ahhhh. That makes sense. Wasn't aware of the buffer and the scan completing instantly. 
Thanks Dave

-----Original Message-----

XuetaoLiu-CLS:
Time Estimate: 4 hours

PR #1390, opened 2015-05-14
Pull Request for Issue1177: IDEAS XRF acquire disable

iainworkman:
#1177 

- Altered the onAcquireSucceeded() function in IDEASXRFDetailedDetectorViewWithSave so that it explicitly disables the acquire button. 
- Altered the onSaveScanButtonClicked() function in IDEASXRFDetailedDetectorViewWithSave so that it connects the slot cleanupScanAction() to the created scan's onSucceeded, onCancelled and onFailed signals.
- Created slot cleanupScanAction() in IDEASXRFDetailedDetectorViewWithSave to disconnect the scan from those slots, schedule it for deletion, null our reference to it and explicitly enable to acquire button again.


source/ui/IDEAS/IDEASXRFDetailedDetectorViewWithSave.cpp
```diff
@@ -253,3 +255,15 @@ void IDEASXRFDetailedDetectorViewWithSave::onDeadTimeCheckButtonClicked()
...
+void IDEASXRFDetailedDetectorViewWithSave::cleanupScanAction()
+{
+	acquireButton_->setEnabled(true);
+	saveScanButton_->setText("Scan Saved");
+	disconnect(scanAction_, SIGNAL(cancelled()), this, SLOT(cleanupScanAction()));
+	disconnect(scanAction_, SIGNAL(succeeded()), this, SLOT(cleanupScanAction()));
+	disconnect(scanAction_, SIGNAL(failed()), this, SLOT(cleanupScanAction()));
+
+	scanAction_->scheduleForDeletion();
+	scanAction_ = 0;
```
iainworkman: I wasn't sure if there was a mechanism in the ScanAction system for the deletion of scans once they are complete, or whether it was the responsibility of the creator. I added a cleanup mechanism here which should return the detector view to the state it was in before it started, but if I'm doubling up duty here for something that the ScanAction system already takes care of let me know and I'll get rid of these.

dretrex: Because all normal scans (essentially, non-XRF scans) are run through the workflow, their memory is properly handled since the AMActionRunner calls ```scheduleForDeletion()``` on all completed actions.  Then, with my changes with cancel, scans will call their cleanup actions regardless of the final state they arrive in.

dretrex:
@XuetaoLiu-CLS can you review this?

XuetaoLiu-CLS:
should we implement this in the same way as SXRMB and VESPERS?
- create XRFScanConfiguration and XRFScanActionController
- start XRF Scan when `startAcquisition` is called (eg. `SXRMBXRFDetailedDetectorView::startAcquistion()`)
- when `save` button is clicked, we will pop up the filtered scan dialog and export the scan data.

@dretrex what do you think?

dretrex:
No, the behaviours are different.  Personally, I prefer the SXRMB/VESPERS method over this one.  We can discuss this at the meeting though.

dretrex:
Wait, I think I misread what you wrote.  Did you suggest that you want to make this like VESPERS/SXRMB or change VESPERS/SXRMB to be like this solution?

XuetaoLiu-CLS:
@dretrex I want to make this like VESPERS/SXRMB ones.

dretrex:
Ah.  Well, we can still talk about it in the meeting.

dretrex:
@XuetaoLiu-CLS were you going to do the changes we talked about in the meeting in this issue or do it in its own issue?

XuetaoLiu-CLS:
@dretrex I can do it here. 
sorry @iainworkman I will steal this issue from your plate :smiley:

iainworkman:
Sure, that works for me.

dretrex:
Hey @XuetaoLiu-CLS, what kind of testing do you still require?  How close is this to being merged into master?

dretrex:
@XuetaoLiu-CLS, can you provide a small update on where this is at?

dleacock:
This has been tested on IDEAS as of today. Dave M approves it and it can be closed.

Issue #917, opened 2014-10-21
Setup auto export for REIXS XAS scans


epengr:
Might as well have REIXS scans auto export as well.  Unless and until there's some kind of automatic/default curvature correction for XES scan, auto export is useless there.

XuetaoLiu-CLS:
@dretrex Darren, I noticed that all beamline set XRF scan auto export false. The XAS scans should be auto-exported. If this is the case, then this can be closed.

Pls correct me if I'm wrong. Thanks.

dretrex:
The XAS scans are not auto-exported because there hasn't been any scan configurations registered to an exporter and exporter option.  Take a look at ```VESPERSAppController::setupExporterOptions()``` for the code necessary to get auto-export up and running.

XuetaoLiu-CLS:
took about 2hrs

dretrex:
Done in #1391.

Issue #813, opened 2014-08-27
REIXSAppController should have code rearranged to match current App controller conventions


dretrex:
While perusing through the REIXSAppController to add some code, I found that it looks and is implemented VERY differently from how every other app controller is made.  We should probably look into and spend some time getting it to match all the others for future ease.

iainworkman:
Looking over this briefly there seems to be two paradigms at play here. 
REIXS and SGM both seem to use a similar paradigm, in which most of the startup work is done in reimplemented virtual functions inherited from AMAppController (startupRegisterDatabases(), startupCreateUserInterface(), startupAfterEverything() etc.). BIOXAS main, side, VESPERS, SXRMB and IDEAS all nominally just use a reimplemented startup() function to do their work (passing it off to functions they define themselves such as registerClasses(), setupUserInterface() etc.)

Should this issue then be extended to SGM too - or is the REIXS and SGM method preferable and should be extended out to the others?

dretrex:
@davidChevrier can correct me if I'm wrong, but the SGM application in its entirety will likely be redone.  I think we'll probably be better off starting from scratch rather than worrying about adding/removing code.

iainworkman:
Sounds good to me.

Thanks Darren.

dretrex:
Done in #1383.  Time estimate?

iainworkman:
Time Estimate: 3 hours

PR #1391, opened 2015-05-14
PullRequest for Issue917: setup auto export for REIXS XAS scan

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/917


source/application/REIXS/REIXS.h
```diff
@@ -0,0 +1,84 @@
...
+		exporterOption->setSeparateHigherDimensionalSources(true);
+		exporterOption->setSeparateSectionFileName("$name_$dataSetName_$fsIndex.dat");
+		exporterOption->setHigherDimensionsInRows(exportSpectraInRows);
+		exporterOption->storeToDb(AMDatabase::database("user"));
+
+		return exporterOption;
+	}
+
+	/// Helper method that takes a time in seconds and returns a string of d:h:m:s.
+	inline QString convertTimeToString(double time)
```
XuetaoLiu-CLS: @dretrex @davidChevrier 
We have exactly the same codes on all the beamlines. I'm wondering whether we should put it somewhere (such as `AMDateTimeUtils`) so that all the beamlines can share the codes?

Same question for `buildStandardExporterOption`.

dretrex: Actually, I think this method is probably unnecessary.  I would be quite shocked if we couldn't get the desired behaviour out of QDateTime::toString().

dretrex: As for buildStandardExporterOption, we want to be a bit more careful.  But conceptually, you are correct if all the beamlines are actually using the same default exporter options.  This likely needs a bit of investigation to determine what a good standard exporter option is.

source/application/REIXS/REIXSAppController.cpp
```diff
@@ -228,6 +227,7 @@ void REIXSAppController::shutdown() {
 
 	// Make sure we release/clean-up the beamline interface
 	AMBeamline::releaseBl();
+	AMStorageRing::releaseStorageRing();
```
dretrex: As per your discussion with @iainworkman earlier, I think you could just put that inside of AMAppController::shutdown().

source/util/AMDateTimeUtils.cpp
```diff
@@ -148,3 +148,39 @@ QString AMDateTimeUtils::gramaticallyCorrectList(const QStringList &listItems){
 	}
 	return retVal;
 }
+
+/// Helper method that takes a time in seconds and returns a string of d:h:m:s.
+QString AMDateTimeUtils::convertTimeToString(double time)
```
dretrex: This also works.  I always forget about this class.

source/application/REIXS/REIXSAppController.cpp
iainworkman: When I moved that call up the hierarchy in #813 I removed it from here at the same time

dretrex:
@iainworkman can you review this? @XuetaoLiu-CLS is this only awaiting testing?

iainworkman:
Sure!

XuetaoLiu-CLS:
@dretrex yes. I will setup a build and test it today.

source/acquaman/SGM/SGMXASScanConfiguration.cpp
```diff
@@ -94,9 +94,9 @@ AMScanConfigurationView* SGMXASScanConfiguration::createView(){
 }
 
 QString SGMXASScanConfiguration::detailedDescription() const{
-	double exitSlit;
-	double grating;
-	double harmonic;
+	double exitSlit = 0;
+	double grating = 0;
+	double harmonic = 0;
```
iainworkman: Looking at this now I'm wondering if there is actually a chance that there might not be a match found for each of these, and if this is the case whether we should catch that in the formatting of the string at the bottom. 
After all if there is no element of the group found with the exit slit gap name an output that indicates the exit slit gap was set to 0 would be included, which isn't technically correct. 

Of course if the group is guaranteed to contain these elements it is a non issue. 
@davidChevrier Is this guarantee the case?

source/application/REIXS/REIXSAppController.cpp
iainworkman: The above comment was made thinking I was looking at a different App Controller. The releaseStorageRing() call here should be removed, as it now lives in AMAppController::shutdown().

XuetaoLiu-CLS:
@dretrex test done

source/acquaman/SGM/SGMXASScanConfiguration.cpp
XuetaoLiu-CLS: Thanks @iainworkman for pointing this out. I have the same question when I was working on this. But if you take a look at the codes L346 `SGMBeamline.cpp`, you will find that we guarantee that we have `grating_`, `harmonic_`, and `exitSlitGap_` initialized.

Issue #1392, opened 2015-05-15
Expanding Scan System Documentation


iainworkman:
It is not obvious on first looking over the Scan classes what job each of them conceptually performs in the running of a scan on the beamline. In the interest of making this a bit clearer the documentation for the scan system should be clarified in their comments.

No code need actually be changed - just the documentation comments for the classes:
- [x] AMScan
- [x] AMScanConfiguration
- [x] AMScanController
- [x] AMScanActionController
- [x] AMScanAction
- [x] AMScanActionInfo
- [x] AMScanActionControllerScanAssembler

#1396

dretrex:
Done in #1396.  Time estimate?

iainworkman:
8.5 hours

PR #1396, opened 2015-05-19
Pull Request for Issue1392: Expanding Scan System Documentation

iainworkman:
#1392 

Added/expanded upon documentation for the scan system classes.
I added a few notes in the comments on stuff that didn't seem quite right. (I'll add line notes in the pr here for those as soon as possible).
Let me know if I'm way off base with any of this.

@dretrex I was wondering also if I should just add everyone to this, as its a discussion that I think pretty much everyone should be part of.

### DIFF
--- a/source/acquaman/AMScanActionController.h
+++ b/source/acquaman/AMScanActionController.h
@@ -131,7 +131,7 @@ protected slots:
 	virtual void onScanningActionsGenerated(AMAction3 *actionTree);
 
 	/*!
-	  * Handles the cleanup after the inisialization actions have succeeded.
+	  * Handles the cleanup after the initialization actions have succeeded.
 	  */
 	virtual void onInitializationActionsListSucceeded();
 
@@ -223,7 +223,7 @@ protected:
 	virtual bool initializeImplementation();
 
 	/*!
-	  * Starts the scan aciton controller actions.
+	  * Starts the scan action controller actions.
 	  */
 	virtual bool startImplementation();
 
--- a/source/acquaman/AMScanActionController.h
+++ b/source/acquaman/AMScanActionController.h
@@ -33,14 +33,28 @@ class QThread;
 #define AMSCANACTIONCONTROLLER_CANNOT_RESUME 262003
 #define AMSCANACTIONCONTROLLER_CANNOT_CANCEL 262004
 
-/// This class provides a simple extension to AMScanController to expose it to some of the AMAction API and AMAgnosticData API.
+/*!
+  * An abstract class which extends the AMScanController with those elements necessary
+  * to interface with the AMAction and AMAgnosticData APIs. Scans which make use
+  * of a scan action controller define the individual steps required to perform
+  * a scan as tree of actions. The actions are defined in three sets:
+  *  - Initialization Actions ~ A list of actions required to prepare the scan for
+  *    running.
+  *  - Scanning Actions ~ A list of actions required to actually perform the scan.
+  *  - Cleanup Actions ~ A list of actions which are performed on completion of
+  *	   the scan.
+  */
 class AMScanActionController : public AMScanController
 {
 	Q_OBJECT
 
 public:
 
-	/// Enum that holds what the final state of the scan action controller is.  This is important for setting the final state after the cleanup actions have been executed.
+	/*!
+	  * Enum which describes the final states which a scan action controller can
+	  * be in. This is required for setting the final state after the cleanup actions
+	  * have been executed.
+	  */
 	enum ScanningActionsFinalState
 	{
 		NotFinished = 0,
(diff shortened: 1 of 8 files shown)


dretrex:
May as well.  Give it a look @acquaman/acquaman-working-group.

dleacock:
This is really helpful, especially how its formatted.  Looks great.

source/actions3/actions/AMScanAction.h
```diff
@@ -39,42 +39,95 @@ class AMScanController;
...
+
 	Q_INVOKABLE AMScanAction(AMScanActionInfo *info, QObject *parent = 0);
-	/// Copy constructor.
+
+	/*!
+	  * Creates a shallow copy of the provided other scan action, but with no
+	  * valid scan controller set.
+	  * //// Is this intended ^ or am I mistaken about the shallow part. The info
+	  * is stored as a reference so we'll just copy the reference, no? ////
+	  */
```
iainworkman: As mentioned the copy c'tor just copies the reference to other's scan action info in the line [here](https://github.com/acquaman/acquaman/blob/master/source/actions3/actions/AMScanAction.cpp#L52). This does make sense, as performing a value copy of a scan action isn't going to want to create a new action info every time, just wanted to make sure.

source/actions3/actions/AMScanAction.h
```diff
@@ -82,46 +135,129 @@ protected slots:
...
 	void onControllerProgressChanged(double elapsed, double total);
-	/// Helper slot that updates the status text when the controller changes state.
+
+	/*!
+	  * Handles the controller signaling that it has made a state transition.
+	  * Updates the status text to reflect the new state.
+	  *
+	  * /// Is this still the case. There's nothing in the body of this method ///
+	  */
```
iainworkman: I just copied the comment for this one, as the body for the method is empty. See [here](https://github.com/acquaman/acquaman/blob/master/source/actions3/actions/AMScanAction.cpp#L432-L435).

source/actions3/actions/AMScanAction.h
```diff
@@ -82,46 +135,129 @@ protected slots:
...
 	void onControllerStateChanged();
 
-	/// Handles making sure that we're ready for deletion
+	/*!
+	  * Prepares the scan action for deletion. The scan action will be added to
+	  * the deletion queue when the logging for this scan has been completed.
+	  * // I'm going off the body here. The name seems to suggest some other logic///
+	  */
```
iainworkman: I think this one is just a naming thing. checkReadyForDeletion() seems to suggest different behaviour than 'waiting till can be deleted, then delete' which the code seems to be doing. See [here](https://github.com/acquaman/acquaman/blob/master/source/actions3/actions/AMScanAction.cpp#L217-L224).

source/actions3/actions/AMScanAction.h
```diff
@@ -82,46 +135,129 @@ protected slots:
...
 //	void onReadyForDeletionChanged(bool isReady);
 
 protected:
-	/// This function is called from the Starting state when the implementation should initiate the action. Once the action is started, you should call notifyStarted().
+
+	/// A lot of these comments (which come down from AMAction3 and are the same
+	/// across most AMAction sub-classes) all mention that you should call notifyX()
+	/// for most of these implemetation, but none of these notify functions exist.
+	/// Is this a remenant from an earlier time?
+
```
iainworkman: I altered these comments to remove the instruction to call notifyX(), but an example can be seen in AMAxisStartedAction::startImplementation() [here](https://github.com/acquaman/acquaman/blob/master/source/actions3/actions/AMAxisStartedAction.h#L53).

source/actions3/actions/AMScanAction.h
dretrex: I believe that this should actually be ```other_.scanInfo()->createCopy()```.  I think we've always just gotten away with it because they haven't been stored to the database yet and therefore haven't had the need to be disassociated.  Not necessarily in this issue, but could you have a look @iainworkman at the other actions to see if this is a consistent bug?

iainworkman: Will do.

iainworkman: Realized that the above example isn't correct. The copy assignment operator isn't defined for the AMScanAction, so the final line won't compile (The Q_OBJECT macro makes it private).

Altering the last line to 
```C++
AMScanAction scanAction2(scanAction1);
```
would produce the mentioned effect, but in this case accidentally creating such a copy is a much less likely situation without coder awareness of a copy taking place. 

The bug still exists, just that explanation of it was not correct.

dretrex: Although I've talked to @iainworkman about this already.  It would be good for the @acquaman/acquaman-working-group knows this as well.  Although we follow the same syntax format as real copy constructors, our copy constructors are __fake__.  Except for a few of our non-QObject classes, the copy constructor is a function that looks like a copy constructor so that our ```createCopy()``` methods can be as simple as possible.

source/acquaman/AMScanActionController.h
```diff
@@ -49,97 +63,232 @@ class AMScanActionController : public AMScanController
...
+
+	/*!
+	  * Handles the setting of the root action in the action tree. Optimizes and
+	  * validates the action tree.
+	  */
 	virtual void onScanningActionsGenerated(AMAction3 *actionTree);
 
-	/// Handles the cleanup after the initialization actions succeed.  Can be reimplemented for more specific behaviour.
+	/*!
+	  * Handles the cleanup after the inisialization actions have succeeded.
```
dretrex: Typo in initialization.

source/acquaman/AMScanActionController.h
```diff
@@ -49,97 +63,232 @@ class AMScanActionController : public AMScanController
...
+
+	/*!
+	  * Builds the initialization actions and the cleanup actions and then starts
+	  * the initialization actions list.
+	  */
 	virtual bool initializeImplementation();
-	/// Simple implementation that starts the scan action controller actions.
+
+	/*!
+	  * Starts the scan aciton controller actions.
```
dretrex: Typo in action.

source/acquaman/AMScanController.h
```diff
@@ -65,147 +84,403 @@ class AMScanController : public QObject
...
+	  * NOTE: This could be called from any state.
+	  *
+	  * In the case class this is  pure virtual function.
+	  */
 	virtual void cancelImplementation() = 0;
-	/// Implement to stop a scan.  This can only be called from the running or paused state.  This will be responsible for calling setFinished() or ensuring that it is called.  Pure-virtual because if you can cancel, then you can also stop.  This is defaulted to an empty string because you may not need to distinguish between reasons.
+
+	/*!
+	  * Performs the actions required to stop a scan.
+	  * \param command ( = "" ) ~ An optional reason for stopping the scan.
```
dretrex: The command string is to allow a general stopping mechanism.  There could be a variety of reasons why/how you want to stop and having a string encapsulate that is the most general way.

source/actions3/actions/AMScanAction.h
dretrex: @iainworkman can you update the comment with the information you know about it now?

source/actions3/actions/AMScanAction.h
```diff
@@ -39,42 +39,95 @@ class AMScanController;
...
+	/*!
+	  * Virtual destructor of the scan action. If there is a valid scan controller
+	  * then the scan controller is disconnected and scheduled for deletion.
+	  */
 	virtual ~AMScanAction();
 
-	/// Creates an identical copy of the action with the exception of setting the state to Constructed.
+	/*!
+	  * Creates a new copy of the scan action with the state set to Constructed.
+	  * //// This was the original comment for this. Is this true. I'm not sure
```
dretrex: Is this comment within a comment still necessary?

source/actions3/actions/AMScanAction.h
dretrex: Well, it's kind of both.  First, if it's ready, then it deletes it.  If it's not ready, it sets it up so that when it _is_ ready, it deletes itself.

source/actions3/actions/AMScanAction.h
dretrex: Yeah good catch.  They are relic comments.  You can remove your note now.

source/actions3/actions/AMScanActionInfo.h
```diff
@@ -40,49 +48,105 @@ class AMScanActionInfo : public AMActionInfo3
...
+	  * \param id ~ The new id of the scan which the scan action info will be
+	  * associated with.
+	  */
 	void setScanID(int id) { scanID_ = id; }
 
-	/// Overrides the warnings string to check warnings from the scan configuration
+	/*!
+	  * The load warning from the scan action info. Actually returns the load
+	  * warning for the contained scan configuration.
+	  * // Is this a null pointer reference waiting to happen? (ie if the current
```
dretrex: It does seem that way.  We probably get away with it because the specific case that would cause the null pointer is not one we can get to based on the implementations we've created.

source/dataman/AMScan.h
```diff
@@ -55,28 +55,26 @@ class AMScanController;
...
+  * AMScan's contain:
+  *		- The meta-data for the scan (start and end date/time, the facility at which
+  *		  it was run, notes etc.).
+  *		- The file format of and paths to the raw data collected by the scan.
+  *		- A container for the raw and analyzed data sources
+  *		- The configuration of the scan
+  *		- A list of the control values storing the initial conditions on the beamline
+  *		  when the scan took place.
+  *
+  *	NOTE: Memory Management of an AMScan is handled through a retain/release mechnaism.
```
dretrex: There is no retain/release system anymore.

source/dataman/AMScan.h
```diff
@@ -122,103 +120,278 @@ class AMScan : public AMDbObject {
...
+	  * \param[out] scanName (= 0) ~ If a reference is provided its value will be
+	  * set with the full name of the scan (if it is found).
+	  */
 	static AMScan* createFromDatabaseUrl(const QUrl& url, bool allowIfScanning = false, bool* isScanning = 0, QString* scanName = 0);
 
 
-	/// Destructor: deletes all channels.
+	/*!
+	  * Cleans up all the resources for this scan.
+	  * NOTE: Due to the retain/release mechanism of AMScan this (nor deleteLater())
```
dretrex: No retain or release methods.  It would be best to remove these comments.

dretrex:
Overall this looks really good @iainworkman.  Thanks for putting in all that effort to make it clearer for everyone else!  Once you have made the couple changes I suggested it's good to be merged in.

source/actions3/actions/AMScanActionInfo.h
iainworkman: Would you mind if I altered it to:
```C++
if(configuration()) {
	return configuration()->dbLoadWarnings();
} else {
        return QString("Unknown error occurred loading the scan configuration");
}
```
Or something to that effect?

dretrex: Yeah, I think that would be a good change.  Let me know once you've added it in and we can merge this in.

iainworkman: Okay @dretrex - that change is made. Should be good to merge now.

Issue #678, opened 2014-06-18
REIXS: Numerous energy move/tolerance issues


AcquamanIssues:
13007 in [AMControlMoveAction3]: There was an error moving the control 'beamlineEV' into position. Target: 858 eV. Actual position: 858.033 eV. Tolerance 1.79769e+308. Reason: An undocumented failure happened.. Please report this problem to the beamline staff.

XuetaoLiu-CLS:
The tolerance is defined as `1000` not `1.79769e+308`. Need some test for this one.

AMPVwStatusControl* directEnergy = new AMPVwStatusControl("beamlineEV", "REIXS:MONO1610-I20-01:user:energy:fbk", "REIXS:user:energy", "REIXS:status", "REIXS:energy:stop", 0, 1000);

XuetaoLiu-CLS:
This actually happens clicking "Stop" when moving energy

dretrex:
That's likely because when you stop it will change the status away from "Move Active" and then it will check to see if it is within tolerance.

XuetaoLiu-CLS:
Took about 1 hr

dretrex:
Done in #1405.

PR #1405, opened 2015-05-21
PullRequest for Issue678: Refine the stop logic so that we display the correct message when controlAction failed

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/678

Original: 
    when stop moving button clicked, the message shows `Reason: An undocumented failure happened`

New fix:
   the error message shows `Reason: The move was manually interrupted or stopped.`

### DIFF
--- a/source/actions3/actions/AMControlMoveAction3.cpp
+++ b/source/actions3/actions/AMControlMoveAction3.cpp
@@ -151,7 +151,6 @@ void AMControlMoveAction3::onMoveReTargetted()
 
 void AMControlMoveAction3::onMoveFailed(int reason)
 {
-	bool reportMoveError = true;
 	bool retryAvailable = false;
 
 	int definedFailureReason;
@@ -169,7 +168,6 @@ void AMControlMoveAction3::onMoveFailed(int reason)
 		break;
 	case AMControl::WasStoppedFailure:
 		definedFailureReason = AMCONTROLMOVEACTION3_WAS_STOPPED_FAILURE;
-		reportMoveError = false;
 		break;
 	case AMControl::AlreadyMovingFailure:
 		definedFailureReason = AMCONTROLMOVEACTION3_ALREADY_MOVING_FAILURE;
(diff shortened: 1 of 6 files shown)


source/beamline/REIXS/REIXSBeamline.cpp
```diff
@@ -874,8 +874,8 @@ AMControl::FailureExplanation REIXSBrokenMonoControl::move(double setpoint)
 bool REIXSBrokenMonoControl::stop()
 {
 	if(moveAction_) {
-		moveAction_->cancel();
 		stopInProgress_ = true;
+		moveAction_->cancel();
```
XuetaoLiu-CLS: This is the key change. We need to set the `stopInProgress_` flag before call the `moveAction_->cancel()`

XuetaoLiu-CLS:
Tested. 
The current error message is 
`Alert: "in [AMControlMoveAction3]: There was an error moving the control 'beamlineEV' into position (Retry No, attempt 1 of 2). Target: 646.017 eV. Actual position: 650.582 eV. Tolerance 0.25. Reason: The move was manually interrupted or stopped... Please report this problem to the beamline staff. (code 13004)"`

source/ui/REIXS/REIXSActionBasedControlEditor.cpp
```diff
@@ -42,19 +42,9 @@ void REIXSActionBasedControlEditor::onNewSetpointChosen(double value)
...
-	}
-	else {
-
-		AMAction3* action = new AMControlMoveAction3(new AMControlMoveActionInfo3(setpoint), control_);
-		connect(action, SIGNAL(succeeded()), action, SLOT(deleteLater()));
-		connect(action, SIGNAL(failed()), action, SLOT(deleteLater()));
-		connect(action, SIGNAL(cancelled()), action, SLOT(deleteLater()));
-		action->start();
-	}
+	AMAction3* action = new AMControlMoveAction3(new AMControlMoveActionInfo3(setpoint), control_);
```
dretrex: You should use AMActionSupport for building control move actions. It is cleaner.

iainworkman:
Everything looks good. The only small thing is that, from a UX perspective, it seems a little odd to be having the users told to report the problem to the beamline staff in the case that they've  manually interrupted or stopped the action. 

There are three places that the failed signal is caught, but only one of them (the general case in AMControlMoveAciton::onMoveFailed(int)) appends the part about reporting the issue to the beamline staff. The question then becomes: is it true for all cases which an AMControlMoveAction encounters a stop command that the beamline staff don't need to be notified? If it is then I would imagine the call to AMErrorMon could be altered in the case that it was an OnStoppedFailure.

XuetaoLiu-CLS:
@iainworkman I added a check for this special case. But I think the better solution is to go through all the error codes we have and the error messages and redesign the alert mechanism.

iainworkman:
Yes, I think you're probably right. Assuming that would be a separate issue, this current branch looks good.

source/actions3/actions/AMControlMoveAction3.cpp
```diff
@@ -151,6 +151,7 @@ void AMControlMoveAction3::onMoveReTargetted()
 
 void AMControlMoveAction3::onMoveFailed(int reason)
 {
+	bool reportMoveError = true;
```
dretrex: This bool is unnecessary because you can just compare against the enum failure value itself.  I'll go ahead and change it since @XuetaoLiu-CLS is away.

PR #1406, opened 2015-05-21
Pull Request for Issue1394: Couple small VESPERS issues

dretrex:
Managed to complete the small list of issues for VESPERS.  Making the fast shutter close at the end of scans was more difficult than I had planned since it required making its own PV control.


source/beamline/VESPERS/VESPERSEndstation.h
```diff
@@ -111,7 +114,7 @@ public slots:
 	/// Sets the current control that should be focused on.  It will find the correct control based on the name.  Control is set to 0 if invalid name is given.
 	void setCurrent(QString name);
 	/// Toggles the current state of the laser power.
-	void toggleLaserPower() { toggleControl(laserPower_); emit laserPoweredChanged(); }
+	void toggleLaserPower() { laserPower_->move(1); emit laserPoweredChanged(); }
```
iainworkman: Just to make sure, setting the toggle PVs to 1 initiates a change in state from on to off or vice verse (it isn't a 1 to on, 0 to off deal)?

dretrex: No, these PVs are poorly designed.  Writing a 1 _toggles_ the state.  If you were at 0, writing a 1 would result in a state of 1.  If you were at state 1, writing a 1 would result in a state of 0.  It's lame because the state PV uses a 0 and 1 as real states.

source/acquaman/VESPERS/VESPERSScanController.cpp
```diff
@@ -129,7 +129,7 @@ AMAction3 *VESPERSScanController::buildCleanupAction()
 	CLSSIS3820Scaler *scaler = VESPERSBeamline::vespers()->scaler();
 
 	AMListAction3 *cleanup = new AMListAction3(new AMListActionInfo3("VESPERS Cleanup", "VESPERS Cleanup"), AMListAction3::Sequential);
-	cleanup->addSubAction(scaler->createDwellTimeAction3(1.0));
+//	cleanup->addSubAction(scaler->createDwellTimeAction3(1.0));
```
iainworkman: Can this commented line be removed?

source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -215,28 +215,24 @@ void AMPseudoMotorControl::removeChildControl(AMControl *control)
 void AMPseudoMotorControl::onMoveStarted(QObject *action)
 {
 	Q_UNUSED(action)
-	qDebug() << name() << "move action started.";
```
iainworkman: There are a couple more debugs in this file (setMoveInProgress() and setIsMoving()). Are these still required? If so can they be an AMErrorReport::Information or AMErrorReport::Debug, then the #include qDebug can be got rid of too.

source/beamline/VESPERS/VESPERSEndstation.h
iainworkman: Couple of questions:
- Are there other beamlines that use these kind of toggle pvs, and if so is there a case to be made for VESPERSToggleControl to be CLSToggleControl?
- If the laser power is toggled in the fashion, can it be a VESPERSToggleControl (or is there some detail of the toggle control/laser behaviour that I'm missing)?

iainworkman:
Aside from a could of questions this looks good.

source/beamline/VESPERS/VESPERSEndstation.h
dretrex: No, thank god VESPERS is the only one to have a PV state machine in this way.  If it was I'd agree, but since I've only seen it on VESPERS I'm okay leaving it as a VESPERS specific control.

source/beamline/AMPseudoMotorControl.cpp
dretrex: @helfrij has handled this in one of her branches.

Issue #1085, opened 2014-12-11
REIXS XAS scan naming issue


epengr:
Auto naming seems to be putting an "Unnamed" in the middle of the sample name for some reason.

dretrex:
Is this a sample naming problem or a scan naming problem?

epengr:
Is appearing in scan names, could originate from sample names, but I suspect not. I may have been unclear (imprecise) above.

iainworkman:
Stole this off you @epengr, as I had a good idea exactly where it was.

dretrex:
Done in #1411.  Time estimate?

iainworkman:
30 minutes

PR #1411, opened 2015-05-25
Pull Request for Issue1085: REIXS XAS scan naming issue

iainworkman:
- Removed the word 'Unnamed' from REIXSXASScanConfiguration::autoScanName() ~ This was presumably this way before intending that it would only be used in cases where a scan name was not set. This isn't currently the case in REIXSXASScanActionController.
- Added the word 'Unnamed' to the REIXSXASScanActionController constructor, in which it initializes the scan and sets its name, for the specific case when we know for certain that the scan name is empty.
- Altered the string returned by REIXSXESScanConfiguration::autoScanName() from 'XES' to 'XES Scan' so both scans are of the same format.

#1085


iainworkman:
I'd usually suggest that this could probably go ahead without necessarily booking time on REIXS, but I'm going to be on there later today, anyway - so might as well put it through the mill before merging.

iainworkman:
Testing revealed this also behaved as expected.

PR #1419, opened 2015-05-27
Pull Request for Issue1416: SXRMB: Add the horizontal slit control to the exposed controls list

dretrex:
Add the horizontal gap control to the exposed controls so that it can be used in the workflow.


iainworkman:
Code looks good.

PR #1428, opened 2015-06-03
Pull Request for Issue1427: BioXAS Commissioning Issues

dretrex:
Fixed the couple small issues that were noticed by BioXAS staff.  The changes will affect all applications, which is nice since they are quality of life bug fixes.


helfrij:
Looks great, @dretrex.

PR #1429, opened 2015-06-04
Pull Request for Issue1420: SMAK exporter exports all data even if last row is incomplete

dretrex:
Logic to not export the last row when exporting SMAK if it is incomplete.


helfrij:
Looks good @dretrex.

PR #1431, opened 2015-06-04
Pull Request for Issue1421: Adding setpoint as option for AMBasicControlDetectorEmulator

helfrij:
Added enum to the AMBasicControlDetectorEmulator class that identifies the control value or setpoint as options for the emulator to monitor, with getter and setter for the new member variable. Set to value in the constructor. Set the option to setpoint for the energySetpointDetector in BioXASSideBeamline.


dretrex:
Looks good @helfrij.

PR #1436, opened 2015-06-08
Pull Request for Issue1120: BioXAS file format exporter

dretrex:
This contains a new exporter type that implements the XDI Format specification for XAS scans.  It will likely be the standard exporter used on BioXAS.


helfrij:
Looks good, @dretrex.

PR #1441, opened 2015-06-10
Pull Request for Issue1159: IDEAS XAS Setup Tweaks

dleacock:
Altered:
* delta E spin box steps 0.5eV instead of 1.0eV
* Moved XRF detector inside analyzed data source check box

Added:
* Extended default near edge region
* Add  detector information (peaking info) to XAS View

https://github.com/acquaman/acquaman/commit/54f4af351bc9f5d2cc21c972ac5f1c5cdb649be7

@epengr - If you'd like to check anything.  Here is a shot of the added detector information in the XAS view after it's been selected in the detector. http://imgur.com/Ab5JjKC


dleacock:
Removed testing label. Just checked it on IDEAS and it works.

source/ui/IDEAS/IDEASXASScanConfigurationView.cpp
```diff
@@ -398,3 +418,8 @@ void IDEASXASScanConfigurationView::onFluorescenceChoiceChanged(int id)
 {
 	configuration_->setFluorescenceDetector(id);
 }
+
+void IDEASXASScanConfigurationView::onDetectorSettingsChange(const QString &args)
```
iainworkman: I don't think this slot is connected to any signal. As this is the only means through which the detectorSettings_ label is updated, I believe it will always be blank.

dleacock: I'm not sure why it's not showing the rest of the code, which is.

void IDEASXASScanConfigurationView::onDetectorSettingsChange(const QString &args)
{
	detectorSettings_->setText(args);
}

source/application/IDEAS/IDEASAppController.cpp
```diff
@@ -206,6 +208,9 @@ void IDEASAppController::onEnergyConnected(bool connected){
 
 		connect(xasScanConfiguration, SIGNAL(totalTimeChanged(double)), xasScanConfigurationHolder3_, SLOT(updateOverallScanTime(double)));
 		xasScanConfigurationHolder3_->updateOverallScanTime(xasScanConfiguration->totalTime());
+
+		connect(ideasKETEKDetailedDetectorViewWithSave_, SIGNAL(peakingChanged(const QString&)), xasScanConfigurationView_, SLOT(onDetectorSettingsChange(const QString&)));
+		connect(ideas13ElementGeDetailedDetectorViewWithSave_, SIGNAL(peakingChanged(const QString&)), xasScanConfigurationView_, SLOT(onDetectorSettingsChange(const QString&)));
```
dleacock: @iainworkman The connection is made here, in the app controller. Is that what you meant? Would there be a better place for it?

iainworkman: Me and @dleacock spent a bit of time on this yesterday. At first glance it looks like wiring a view to another view is a little fragile, but there doesn't seem to be another option with the current  way the detectors are setup in the application. I'm going to keep digging with this, because I feel like there should be a more robust way of listening for this change that by wiring up to the widgets themselves.

davidChevrier:
@iainworkman 
Out of curiosity, what are you guys trying to achieve by doing this? What's the end goal?

iainworkman:
I believe this relates to the 'Add feedback on detector rate mode to XAS setup panel' part of the issue, where @dleacock is attempting to display a label for the peaking time into IDEASXASScanConfigurationView. This will need to update when the peaking time is changed elsewhere, and so we were looking at the best way.

iainworkman:
I feel like, as the detector view seems to be hopping over to the beamline class to make the change, the most robust way, given the current structure, would probably be to listen for the ketek peaking time controls moveSucceeded() signal.

davidChevrier:
So, I think you've identified that view to view signalling is a bad idea. I think the actual core problem is that IDEASKetekDetector isn't giving you the signal you actually need. It also looks like some of the controls for the ketek don't actually live in the ketek. 

I think that the ketek should be extended to make those peaking time and preamp gain controls part of the class. Then you can add signalling coming out of the ketek when the state changes. I think that an enum would be the best way to go and the logic could be moved out of the IDEASXRFDetailedDetectorViewWithSave class.

dretrex:
@davidChevrier is correct.  Views should only represent the values their model provide and should only have an effect on their particular model as well.  If the ketek needs to update the peaking time, then it should have the corresponding getters/setters/signals to facilitate that.

dleacock:
@dretrex I believe I'm finished what we discussed last week. Just needs a few remaining PV values for the 13 Element detector that weren't there.

source/application/IDEAS/IDEASAppController.cpp
```diff
@@ -156,7 +160,7 @@ void IDEASAppController::setupUserInterface()
 
 	mw_->insertHeading("XRF Detectors", 1);
 
-	ideasKETEKDetailedDetectorViewWithSave_ = new IDEASXRFDetailedDetectorViewWithSave(IDEASBeamline::ideas()->ketek());
+	KetekDetailedDetectorView *ideasKETEKDetailedDetectorViewWithSave_ = new KetekDetailedDetectorView(IDEASBeamline::ideas()->ketek());
```
dretrex: Hey @dleacock, you forgot the IDEAS prefix to the class name.  The views should be _IDEASKetekDetailedDetectorView_.  Could you go ahead and change these?  You'll need to make sure all the includes change, the class protection macro, and the .pri/.pro files.

source/application/IDEAS/IDEASAppController.cpp
```diff
@@ -164,7 +168,7 @@ void IDEASAppController::setupUserInterface()
 	ideasKETEKDetailedDetectorViewWithSave_->addCombinationPileUpPeakNameFilter(QRegExp("(Ka1|La1|Ma1)"));
 	mw_->addPane(ideasKETEKDetailedDetectorViewWithSave_, "XRF Detectors", "KETEK", ":/system-search.png");
 
-	ideas13ElementGeDetailedDetectorViewWithSave_ = new IDEASXRFDetailedDetectorViewWithSave(IDEASBeamline::ideas()->ge13Element());
+	g13ElementDetailedDetectorView *ideas13ElementGeDetailedDetectorViewWithSave_ = new g13ElementDetailedDetectorView(IDEASBeamline::ideas()->ge13Element());
```
dretrex: Same thing with this.  A couple other small pointers:
- Classes should all begin with capitals.
- Since this is the 13 element, it should be Ge (for germanium) not just g.
- Beamline specific classes should have their specific identifier (ie: IDEAS, SGM, etc.)

source/beamline/IDEAS/IDEAS13ElementGeDetector.h
```diff
@@ -47,7 +47,15 @@ class IDEAS13ElementGeDetector : public AMXRFDetector
 	/// The vortex detectors support elapsed time.
 	virtual bool supportsElapsedTime() const { return true; }
 
+	double peakingTime() const { return peakingTimeControl_->value(); }
```
dretrex: Remember that functions should always be commented.  Even if it is just as simple as "Returns the peaking time for the detector."  It's always good practice to do it as you go because you often-times don't make it back to comment it when you're done.

source/beamline/IDEAS/IDEAS13ElementGeDetector.h
```diff
@@ -56,6 +64,21 @@ public slots:
...
 	/// Vortex detectors do not support clearing
 	virtual bool clear() { return false; }
+
+	void setPeakingTime(double time);
+
+	void setPreampGain(double value);
+
+protected:
+	AMReadOnlyPVControl *ge13ElementRealTimeControl_;
+	AMDetector *ge13ElementRealTime_;
```
dretrex: Did you add the real time detector to the class @dleacock?  Typically, we've been putting the one-off detectors inside the beamline class rather than detector classes themselves.

source/beamline/IDEAS/IDEAS13ElementGeDetector.h
```diff
@@ -56,6 +64,21 @@ public slots:
...
+
+	AMPVControl *peakingTimeControl_;
+
+	AMControl *triggerLevel_;
```
dretrex: These should also be commented.

source/beamline/IDEAS/IDEASBeamline.cpp
```diff
@@ -196,8 +175,6 @@ void IDEASBeamline::setupExposedDetectors()
 	addExposedDetector(ReferenceIonChamberScaler_);
 	addExposedDetector(ketek_);
 	addExposedDetector(ge13Element_);
-	addExposedDetector(ketekRealTime_);
```
dretrex: These removals (and the corresponding removals above of the exposed controls) is why we put the one off detectors inside the beamline class.  The exposed detectors and controls are incredibly important to running of generic scans and taking advantage of many of the "for free" features that AM provides.

source/beamline/IDEAS/IDEASKETEKDetector.h
```diff
@@ -67,6 +70,18 @@ class IDEASKETEKDetector : public AMXRFDetector
     /// The KETEK detectors support elapsed time.
     virtual bool supportsElapsedTime() const { return true; }
 
+    double peakingTime() const { return peakingTimeControl_->value(); }
```
dretrex: Comments.

source/beamline/IDEAS/g13ElementDetailedDetectorView.h
```diff
@@ -0,0 +1,34 @@
...
+
+	/// Destructor
+	virtual ~g13ElementDetailedDetectorView();
+
+	/// Re-implementing to add peaking time spin box
+	virtual void buildDetectorView();
+
+public slots:
+	/// Handles setting the peaking time.
+	void onPeakingTimeChanged();
```
dretrex: There is a naming convention that we can talk about some time for slots like these.  They are handy for identifying what the slot is doing and what caused it to trigger.

source/ui/IDEAS/IDEASXASScanConfigurationView.cpp
```diff
@@ -398,3 +420,17 @@ void IDEASXASScanConfigurationView::onFluorescenceChoiceChanged(int id)
 {
 	configuration_->setFluorescenceDetector(id);
 }
+
+void IDEASXASScanConfigurationView::onPeakingTimeChanged(double value)
```
dretrex: I don't think I understand, but why is this here and in the detector view?

dleacock: In Issue1159 that last request was for the any changes in peaking time made in the detector view were also shown in the scan configuration view. I have the scan configuration view listen for peaking changes made in the detector.

dretrex: Ah, I see.  Unfortunately, we need to be a little careful.  We want to minimize the amount of beamline calls inside of scan configurations and configuration views.  We can have a chat about that sometime as well if you'd like.

source/beamline/IDEAS/IDEAS13ElementGeDetector.h
dleacock: The real time detector stuff originally was in the beamline class but I moved it out to keep things consistent but I can easily move it back.

source/application/IDEAS/IDEASAppController.cpp
```diff
@@ -156,7 +160,7 @@ void IDEASAppController::setupUserInterface()
 
 	mw_->insertHeading("XRF Detectors", 1);
 
-	ideasKETEKDetailedDetectorViewWithSave_ = new IDEASXRFDetailedDetectorViewWithSave(IDEASBeamline::ideas()->ketek());
+	IDEASKETEKDetailedDetectorView *ideasKETEKDetailedDetectorViewWithSave_ = new IDEASKETEKDetailedDetectorView(IDEASBeamline::ideas()->ketek());
```
XuetaoLiu-CLS: you should remove `IDEASKETEKDetailedDetectorView *`. Otherwise, it will make `ideasKETEKDetailedDetectorViewWithSave_` a local variable instead a class property.

source/application/IDEAS/IDEASAppController.cpp
```diff
@@ -164,7 +168,7 @@ void IDEASAppController::setupUserInterface()
 	ideasKETEKDetailedDetectorViewWithSave_->addCombinationPileUpPeakNameFilter(QRegExp("(Ka1|La1|Ma1)"));
 	mw_->addPane(ideasKETEKDetailedDetectorViewWithSave_, "XRF Detectors", "KETEK", ":/system-search.png");
 
-	ideas13ElementGeDetailedDetectorViewWithSave_ = new IDEASXRFDetailedDetectorViewWithSave(IDEASBeamline::ideas()->ge13Element());
+	IDEAS13ElementGeDetailedDetectorView *ideas13ElementGeDetailedDetectorViewWithSave_ = new IDEAS13ElementGeDetailedDetectorView(IDEASBeamline::ideas()->ge13Element());
```
XuetaoLiu-CLS: same here

source/application/IDEAS/IDEASAppController.cpp
XuetaoLiu-CLS: I also did the same change in Issue1177, to rename `IDEASXRFDetailedDetectorViewWithSave` as `IDEASXRFDetailedDetectorView` :+1: :smile:

source/beamline/IDEAS/IDEAS13ElementGeDetector.cpp
```diff
@@ -17,6 +18,18 @@ IDEAS13ElementGeDetector::IDEAS13ElementGeDetector(const QString &name, const QS
...
+	// Detector functionality
+	ge13ElementRealTimeControl_ = new AMReadOnlyPVControl("13-el Ge Real Time", "dxp1608-B21-13:ElapsedReal", this);
+	ge13ElementRealTime_ = new AMBasicControlDetectorEmulator("13E_dwellTime", "13-element Ge dwell time", ge13ElementRealTimeControl_, 0, 0, 0, AMDetectorDefinitions::ImmediateRead, this);
+
+
+	// Missing PV values for peaking, baseline threshold, preamp gain control. Waiting to hear back.
+	peakingTimeControl_;
+	triggerLevel_;
+	baselineThreshhold_;
+	preampGainControl_;
```
XuetaoLiu-CLS: Don't forget to initialize those PVs

dleacock: Just waiting to hear back what the PV values are then I can update it.

dleacock:
Thank you for your comments @XuetaoLiu-CLS

source/beamline/IDEAS/IDEAS13ElementGeDetailedDetectorView.cpp
```diff
@@ -0,0 +1,18 @@
+#include "IDEAS13ElementGeDetailedDetectorView.h"
+
+IDEAS13ElementGeDetailedDetectorView::IDEAS13ElementGeDetailedDetectorView(IDEAS13ElementGeDetector *detector, QWidget *parent)
+	: IDEASXRFDetailedDetectorViewWithSave(detector, parent)
```
XuetaoLiu-CLS: The super class should be `IDEASXRFDetailedDetectorView`

source/beamline/IDEAS/IDEAS13ElementGeDetailedDetectorView.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+IDEAS13ElementGeDetailedDetectorView::~IDEAS13ElementGeDetailedDetectorView()
+{
+
+}
+
+void IDEAS13ElementGeDetailedDetectorView::buildDetectorView()
+{
+	/// Call inherited parent classes buildDetectorView
+	IDEASXRFDetailedDetectorViewWithSave::buildDetectorView();
```
XuetaoLiu-CLS: This function / implementation can be deleted since you don't have specialized implementation in the subclass

source/beamline/IDEAS/IDEAS13ElementGeDetailedDetectorView.h
```diff
@@ -0,0 +1,30 @@
...
+
+	/// Re-implementing to add peaking time spin box
+	virtual void buildDetectorView();
+
+public slots:
+
+
+protected:
+	/// Pointer to actual ketek detector for setting peaking time
+	IDEAS13ElementGeDetector *ge13ElementDetector_;
```
XuetaoLiu-CLS: didn't see usage for this variable

epengr:
Just a note - the Ketek and 13Element detectors do not use the same electronics - the ketek is usinf and XIA Mercury and the element is using XIA XMAP.

dretrex:
Is this guy all ready for a final review now @dleacock?

dleacock:
Ready for review!

source/beamline/IDEAS/IDEASBeamline.h
```diff
@@ -218,8 +208,20 @@ protected slots:
 	CLSSIS3820Scaler *scaler_;
 
 	IDEASKETEKDetector *ketek_;
+	AMControl *ketekPeakingTime_;
```
epengr: I thought these controls were all moved inside the IDEASKETEKDetector class?

dretrex: That might be from a different pull request.  Maybe we'll need to make sure that code is merged in before a final review.

dleacock: @dretrex Remember when we had that discussion about the "one-off" features? I had removed them from IDEASBeamline but you advised to put them back in. Those are for the exposedControls in IDEASBeamline::setupExposedControls.

dretrex: Right.

source/beamline/IDEAS/IDEAS13ElementGeDetector.cpp
```diff
@@ -17,6 +18,27 @@ IDEAS13ElementGeDetector::IDEAS13ElementGeDetector(const QString &name, const QS
...
 
+	// Detector functionality
+	ge13ElementRealTimeControl_ = new AMReadOnlyPVControl("13-el Ge Real Time", "dxp1608-B21-13:ElapsedReal", this);
+	ge13ElementRealTime_ = new AMBasicControlDetectorEmulator("13E_dwellTime", "13-element Ge dwell time", ge13ElementRealTimeControl_, 0, 0, 0, AMDetectorDefinitions::ImmediateRead, this);
+
+    // Energy Peaking Time
+    peakingTimeControl_ = new AMSinglePVControl("Peaking Time", "dxp1608-B21-13:EnergyPkTime", this, 0.01);
+
+    // preamp Gain controls for elements 1-12
+    preampGainControl1_ = new AMPVControl("13-el Ge Preamp Gain 1","dxp1608-B12-13:dxp1:PreampGain_RBV","dxp1608-B12-13:dxp1:PreampGain", QString(), this, AMCONTROL_TOLERANCE_DONT_CARE);
```
dretrex: This could be replaced by a list (i.e.: `preampGainControls_`) and then you culd have made a simple loop such as:

```
for (int i = 0; i < 12; i++){
     preampGainControls_ << new AMPVControl(QString("13-el Ge Preamp Gain %1").arg(i+1), QString(dxp1608-B12-13:dxp%1:PreampGain_RBV").arg(i+1), "dxp1608-B12-13:dxp%1:PreampGain", QString(), this);
}
```

source/beamline/IDEAS/IDEAS13ElementGeDetector.h
```diff
@@ -51,8 +51,39 @@ class IDEAS13ElementGeDetector : public AMXRFDetector
 	/// The vortex detectors support elapsed time.
 	virtual bool supportsElapsedTime() const { return true; }
 
+	/// Returns the real time for the detector.
+	AMDetector *dwellTime() const {return ge13ElementRealTime_; }
+
+    /// Returns AMPVControl for peaking time
+    double peakingTime() const { return peakingTimeControl_->value(); }
+    /// Returns AMPVControl for preamp gain
+    double preampGain1() const  { return preampGainControl1_->value(); }
```
dretrex: I would change this to `double preampGainAt(int index) const` and use the list method I described.

source/ui/IDEAS/IDEASXASScanConfigurationView.cpp
```diff
@@ -155,15 +157,19 @@ IDEASXASScanConfigurationView::IDEASXASScanConfigurationView(IDEASXASScanConfigu
 	mainVL->addWidget(regionsView_);
 	mainVL->addLayout(numberOfScansLayout);
 
-
 	QLabel *settingsLabel = new QLabel("Scan Settings:");
 	settingsLabel->setFont(QFont("Lucida Grande", 12, QFont::Bold));
 
+	QHBoxLayout *XRFDetector = new QHBoxLayout();
```
dretrex: Member variables typically shouldn't start with capitals since classes always do start with capitals.  `XRFDetector` -> `xrfDetector`

source/beamline/IDEAS/KetekDetailedDetectorView.cpp
```diff
@@ -0,0 +1,72 @@
+#include "KetekDetailedDetectorView.h"
```
dretrex: You are missing the IDEAS prefix in the name of the class and the file name.

source/beamline/IDEAS/g13ElementDetailedDetectorView.cpp
```diff
@@ -0,0 +1,72 @@
+#include "g13ElementDetailedDetectorView.h"
```
dretrex: You are missing the IDEAS prefix here too.  Also, classes don't start with lower case letters.  You should also include the "e" for the germanium detector so that it is explicit (i.e.: IDEASGe13ElementDetailedDetectorView).

dretrex:
Okay.  I think that this is good to go now.  Coming up with a time estimate is going to be very tough, but if everyone can try and put an estimate that would be beneficial.

For myself, I'm going to say I probably spent a day all things considered.

Time estimate: 1 day

Issue #1245, opened 2015-03-17
Change data updates in all core classes


dretrex:
While looking into #1240 I noticed that there are some very important things that hold us back in terms of optimization of data.  

- First is the inefficient way we update data ranges.  They default to using null AMnDIndex's which are valid for 0D data and can complicate logic when trying to cache or minimize data handling.
- Second is MPlot will ALWAYS go through every piece of data regardless of what changed.

There are also a couple things I would like to do while I'm at it.
- Change the need to create large CDF data stores right off the bat for images.
- Have proper range checking and caching for all MPlot and AMDataSource classes.

The to do list for this issue is the following:
- [x] Change MPlot to use bounds when emitting data changed.
- [ ] Update all the MPlot drawing and data classes to properly use ranges and caching.
  - __1D Plots__
  - [ ] ... to be fleshed out...
  - __2D Images__
  - [x] MPlotAbstractImageData
  - [x] MPlotSimpleImageData
  - [x] ~~MPlotSimpleImageDatawDefault~~ __Removed__
  - [ ] MPlotAbstractImage
  - [ ] MPlotImageBasic
  - [x] ~~MPlotImageBasicwDefault~~ __Removed__
  - [x] AMDataSourceImageData
  - [x] ~~AMDataSourceImageDatawDefault~~ __Removed__
  - [ ] MPlotColorLegend
- [x] Change AMDataSource to not use defaulted arguments for emitDataChanged().
- [x] Get all subclasses of AMDataSource to properly handle the fact that they can always trust the range of the data they are getting.
- [x] Make convenience methods (if they don't exist already) for giving the minimum and maximum values in AMnDIndex. This may have already been done in #1240 and #1454.
- [ ] Update the Scan editor for maps to generate a correctly sized original image even if its data is only a fraction of what is displayed.

dretrex:
This will also fix #1090 because the issue with that is I haven't been prefilling the data.  I will probably be able to rework the MPlotImagewDefault a bit while doing this upgrade.

dretrex:
So far I've spent about a 8 hours on this.

dretrex:
There is a lot of good thoughts about this, but I'm going to close this and start over.  New issues will be created to make this a more manageable task.

PR #1454, opened 2015-06-18
Pull Request for Issue1240: Optimize analysis blocks to cache and retrieve data only when necessary

dretrex:
This is a somewhat comprehensive list of analysis blocks that I have implemented caching in.  They are mostly tested but still require some tests with running scans to make sure that they handle dynamic data changes rather than just the case of reloading old scans.

@iainworkman, I'd like you to review it for me.  Try and be as scrupulous as possible because it's all indexing craziness.


iainworkman:
Sure. I'm just going to spend a bit of time familiarizing myself with the AB interface and AMnDIndex

source/analysis/AM1DDerivativeAB.cpp
```diff
@@ -301,15 +296,15 @@ bool AM1DDerivativeAB::values(const AMnDIndex &indexStart, const AMnDIndex &inde
 				badIndices.append(totalSize-1);
 
 			// Compute all the values
-			outputValues[0] = (data.at(1)-data.at(0))/(axis.at(1)-axis.at(0));
-			outputValues[totalSize-1] = (data.at(totalSize-1)-data.at(totalSize-2))/(axis.at(totalSize-1)-axis.at(totalSize-2));
+			cachedOutput_[0] = (data.at(1)-data.at(0))/(axis.at(1)-axis.at(0));
```
iainworkman: This part seems to be confusing me a little bit (not specifically what you've done, but the already present checks for divide by zero).
If `axis[0] == axis[1]` (for example) the check finds this and adds 0 to the badIndices list. However we still then proceed to 
```C++
cachedOutput_[0] = (data.at(1)-data.at(0))/(axis.at(1)-axis.at(0));
```
which would surely still cause a divide by zero error, no? The inherent imprecision in doubles is part of what's confusing me here.

iainworkman:
Apart from the above confusion this looks fine to me. I'm going to try and find you this morning regarding the indexing you're using (just to make sure I have it all clear in my mind).

source/analysis/AM1DDerivativeAB.cpp
dretrex: You are correct, however, this probably doesn't affect anything because that value is not used and then is overwritten by the bad indices check.  I agree that it should be checked and fixed though.  Good catch.

dretrex:
2 hours

dretrex:
It turns out the caching of indices is fundamentally flawed for handling data coming in one at a time.  Or at the vary least, I need to look at the logic again.  It seems to not handle the fact that multiple updates happen.  The first update will likely have the correct "dirty" indices, but the second won't and therefore will likely be incorrect.   I think an appropriate approach would be to start with full recalculation caching and then move towards the full caching that I wanted.

dretrex:
I'm closing this pull request because I'm going to start this issue over but armed with the lessons learned from this and #1245.

Issue #1444, opened 2015-06-15
Add display for the x,y coordinates of a clicked position in a 1D scan.


helfrij:
In a 2D scan, clicking on the scan view will cause the x, y, and intensity at that position to be displayed. We'd like to be able to add this functionality to the 1D scan view.

iainworkman:
Merged in #1455.

dretrex:
Since this appears to be all done, I'm closing this out.  Time estimate @helfrij?

helfrij:
Time estimate: 2 weeks

PR #1455, opened 2015-06-22
Pull Request for Issue1444: Create view to display (x,y) clicked position information in AMScanView

helfrij:
- [x] Create view that allows user to select the tools to be applied to the plot
- [x] Add tool view to AMScanView.
- [x] Create 'model' class that the tool view listens to, with instances inside each AMScanViewInternal subclass
- [x] Coordinate tools applied to the plot and the tool view selections through the model class.
- [x] Create view to display specific information related to the data position tool, add to tool view.
- [x] Only show the data position tool view when the data position tool is selected.
- [x] Communicate the values to be displayed to the data position tool view, from the plot that's clicked.
- [x] Add data position tool marker for clicked positions
- [x] Testing


helfrij:
@iainworkman @XuetaoLiu-CLS

The work done in this pull request is mostly complete--the desired features have been added to AMScanView. It's at a point now where review would be helpful to iron out any kinks. Would you mind taking a look, when you have a chance?

iainworkman:
Sure, will have a look this morning.

source/ui/dataman/AMScanView.h
```diff
@@ -323,26 +346,17 @@ class AMScanViewMultiView : public AMScanViewInternal {
...
-
-	/// Sets the visibility of the plot item cursor.
-	void setPlotCursorVisibility(bool visible);
-	/// Sets the coordinates for the plot item cursor.
-	void setPlotCursorCoordinates(const QPointF &coordinates);
-	/// Overloaded.  Sets the x-axis coordinate only.
-	void setPlotCursorCoordinates(double xCoordinate);
-	/// Sets the color of the plot cursor.
-	void setPlotCursorColor(const QColor &color);
-
```
iainworkman: During compile I get errors related to classes which still call these now missing functions. Are these now handled internally, or is there some other mechanism that external classes can change these settings?

An example is REIXSRIXSScanConfigurationView::addScan() and ::onEnergySpinBoxChanged().

helfrij: Ooops, thanks @iainworkman. I'll take a look at this.

source/ui/dataman/AMScanView.cpp
```diff
@@ -412,7 +424,35 @@ void AMScanView::onScanAdded(AMScan *scan)
 
 	modeBar_->showSpectra_->setEnabled(!sources.isEmpty());
 	modeBar_->showSpectra_->setVisible(!sources.isEmpty());
+
+	// Apply changes to views.
+
 	spectrumView_->setDataSources(sources);
+
+//	QString xUnits = "";
```
XuetaoLiu-CLS: Do we still need those commented off codes?

source/ui/dataman/AMScanView.cpp
```diff
@@ -575,13 +617,39 @@ MPlotGW * AMScanViewInternal::createDefaultPlot()
...
+		if (positionTool) {
+			positionTool->setDataPositionIndicator(plot->axisScaleBottom(), plot->axisScaleLeft());
+		}
+	}
+}
+
+void AMScanViewInternal::removeToolFromPlot(MPlot *plot, MPlotAbstractTool *tool)
+{
+	if (plot && tool && plot->tools().contains(tool)) {
+		plot->removeTool(tool);
```
XuetaoLiu-CLS: Is it the caller's responsibility to call `tool->deleteLater()` or this function's?

source/ui/dataman/AMScanView.cpp
```diff
@@ -689,48 +763,26 @@ AMScanViewExclusiveView::AMScanViewExclusiveView(AMScanView* masterView) : AMSca
...
 
 	reviewPlotAxesConfiguration(plot_);
 	refreshTitle();
 }
 
 AMScanViewExclusiveView::~AMScanViewExclusiveView() {
-	// PlotSeries's will be deleted as children items of the plot.
-
-	delete plotCursor_;
 	plot_->deleteLater();
```
XuetaoLiu-CLS: Don't we need to delete `dragZoomerTool_, wheelZoomerTool_  and dataPositionTool_` here?
`

XuetaoLiu-CLS: Sorry. Never mind. Saw the codes below :+1:

source/ui/dataman/AMScanView.cpp
```diff
@@ -1106,15 +1164,7 @@ void AMScanViewMultiView::addScan(int si) {
...
- for(int si=0; si<plotSeries_.count(); si++)
-  for(int ci=0; ci<model()->scanAt(si)->dataSourceCount(); ci++)
-   if(plotSeries_[si][ci]) {
- plot_->plot()->removeItem(plotSeries_[si][ci]);
- delete plotSeries_[si][ci];
-   }*/
-
-	delete plotCursor_;
+//	delete plotCursor_;
 	plot_->deleteLater();
```
XuetaoLiu-CLS: I guess we need to delete `dragZoomerTool_  and wheelZoomerTool_ ` as what you did for `AMScanViewExclusiveView`

source/ui/dataman/AMScanView.cpp
helfrij: It is the subclasses of AMScanViewInternal (ExclusiveView, MultiView, etc) that create instances of the tools they'd like available--it's probably best if they are each responsible for managing the memory set aside for tools. I'll double-check that this is the case.

source/ui/dataman/AMScanView.cpp
```diff
@@ -507,33 +547,9 @@ void AMScanView::mousePressEvent(QMouseEvent *e)
...
-void AMScanView::setPlotCursorColor(const QColor &color)
-{
-	((AMScanViewExclusiveView *)(views_.at(0)))->setPlotCursorColor(color);
-	((AMScanViewMultiView *)(views_.at(1)))->setPlotCursorColor(color);
-}
-
 #include <QSizePolicy>
 
- AMScanViewInternal::~AMScanViewInternal(){}
+ AMScanViewInternal::~AMScanViewInternal(){ delete tools_;}
```
XuetaoLiu-CLS: or we can loop the tools in the `tools_` to delete them here.

source/ui/dataman/AMScanView.cpp
```diff
@@ -1354,6 +1420,18 @@ AMScanViewMultiScansView::AMScanViewMultiScansView(AMScanView* masterView) : AMS
 	firstPlotEmpty_ = true;
 	plots_ << plot;
 
+	// Create instances of the tools we want available and add them to the tool options.
+
+	dragZoomerTool_ = new MPlotDragZoomerTool();
+	wheelZoomerTool_ = new MPlotWheelZoomerTool();
```
XuetaoLiu-CLS: same here. We need to delete it in the destructor

helfrij:
Taking a look at MPlot::addTool(), it looks like the plot is set explicitly as the parent of the tool here. If the instance of AMScanViewInternal is deleted when the plot is the tool's parent, I think the plot correctly manages the tool's memory. However, the tool is orphaned in MPlot::removeTool(): there may be memory issues if the scan view is deleted when the tool is not added to the plot.

source/ui/dataman/AMScanView.cpp
```diff
@@ -689,48 +763,26 @@ AMScanViewExclusiveView::AMScanViewExclusiveView(AMScanView* masterView) : AMSca
...
-	plotCursor_->setValue(QPointF(xCoordinate, plotCursor_->value().y()));
-}
-
-void AMScanViewExclusiveView::setPlotCursorColor(const QColor &color)
-{
-	plotCursor_->marker()->setPen(QPen(color));
-	plotCursor_->marker()->setBrush(QBrush(color));
+	delete dragZoomerTool_;
+	delete wheelZoomerTool_;
+	delete dataPositionTool_;
```
iainworkman: I think these will need to be a delete later. Each inherits MPlotAbstractTool which is on the QObject inheritance tree.

helfrij: Ah that's right. Thanks @iainworkman.

source/ui/dataman/AMScanView.cpp
helfrij: Nice catch! Thanks @XuetaoLiu-CLS.

helfrij:
Yes, I've made a couple of changes to MPlot that are used in this branch. Changes include a couple of additional functions in the MPlot class and a new tool. Is it safe to update our versions of MPlot, @davidChevrier? I know you had planned to do some work with the package yesterday.

davidChevrier:
@helfrij @iainworkman @XuetaoLiu-CLS @dretrex 
Yes. It should be safe now to work with MPlot again.

source/ui/dataman/AMScanView.h
helfrij: I've reimplemented these methods for AMScanView, which should allow other applications to maintain their current functionality.

source/ui/dataman/AMScanView.cpp
helfrij: I've modified the destructors for all AMScanViewInternal subclasses to call deleteLater() on all tools they've constructed.

source/ui/dataman/AMScanView.cpp
```diff
@@ -412,7 +469,30 @@ void AMScanView::onScanAdded(AMScan *scan)
...
 	spectrumView_->setDataSources(sources);
+
+	QString xUnits = "";
+
+	if (scan && scan->rawData() && scan->rawData()->scanAxesCount() >= 1) {
+		xUnits = scan->rawData()->scanAxisAt(0).units;
+	}
+
+	foreach (AMScanViewInternal *scanView, views_) {
+		if (scanView && scanView->tools()) {
```
XuetaoLiu-CLS: A suggestion:
It might be nicer if we encapsulate this code block in `AMScanViewInternal` with a function name like `setUnits(QString units)`. Then the code in this function will be like:
```
if (scanView)
    scanView->setUnits(xUnits);
```

The user of `AMScanViewInternal` doesn't need to know how the the units if passed to the MPlot tools.

helfrij: I agree that this code block is awkward and can benefit from some simplification. 

It's like this right now because 'units' is neither a property of the scan, the general AMScanView, nor of the AMScanViewInternal subclasses visualizing the scan data (AMScanViewInternal, AMScanViewExclusiveView, AMScanViewMultiView, etc.). The x axis units are a property of the first scan axis in the scan's raw data and the y axis units a property of each individual data source displayed (trickier to get at and manage, which is why y axis units aren't implemented now). 

The main goal of the work in this pull request is to display the data coordinates and units of a click position. Because I'd chosen to go the route of using the tools as models for the tool views, I've added a 'units' variable to MPlotDataPositionTool. So it's really just this particular tool that cares about the units (not any of the views, or other tools). Because there is no requirement that all AMScanViewInternal views use the same kinds of tools, a 'setUnits' type method in these classes would imply a relationship that presently isn't there and do nothing if that view didn't contain an MPlotDataPositionTool, which isn't really desirable either.

An alternative could be to make 'units' a property of the views, then a 'setUnits' type method would make a lot of sense. It might be worth the code refactor, either now or later on.

iainworkman:
@helfrij This looks great to me. There's just a couple of user experience things I noted. 

- Would it be possible to add a tool tip to the tool buttons, just so it's obvious which each one is?
- The 'cursor' line in the data cursor is selectable when you're using the selection tool. Is this the desired behaviour?
- I'm not sure what the magnifying glass tool's use is. It seems to be not responding to clicks.

helfrij:
Great ideas, thanks @iainworkman. The magnifying glass is the tool that uses the mouse wheel to zoom in and out of the plot. I've added tooltips for each tool, so hopefully that should make its function a little more obvious. The cursor is now not selectable (this change requires an MPlot update).

iainworkman:
@helfrij  There's a couple of things that cropped up when testing just now. When the cursor tool is highlighted on an open scan, and I close it I'm getting a seg fault. If I don't select any tools I'm also getting a `QCoreApplication::postEvent: Unexpected null receiver` message (though the application continues as normal after these).

helfrij:
Thanks, @iainworkman, I just reproduced this error. Not sure what's happening yet.

helfrij:
I just tried opening an old scan in master and closing it using the 'x' button on the left-side menu and Acquaman crashed (but is fine when closed using the 'close scan' button). This suggests this crash is a pre-existing problem. GDB shows only ?? when I do a backtrace, which suggests this is a problem in MPlot. Still looking into the possible cause.

helfrij:
I think I've found and fixed the problem now. When they are applied to the plot, the MPlotPoints that make the indicators for MPlotDataPositionTools and cursors for MPlotDataPositionCursorTool are deleted with the rest of the plot items once the plot is deleted--causing a crash when the same items are deleted again from the tool destructors. I think I've fixed the problem by removing the call to delete the indicator altogether from the MPlotDataPositionTool destructor and only deleting the cursor if it isn't displayed on the plot in the MPlotDataPositionCursorTool.

iainworkman:
Grand! I'm pulling down the latest MPlot version and #1444 at the moment. Will try them out and if all looks good I'll merge them in this morning.

helfrij:
Thanks, @iainworkman!

source/ui/dataman/AMScanViewPlotSelectedToolsView.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+		int toolCount = tools_->tools().count();
+
+		for (int toolIndex = 0; toolIndex < toolCount; toolIndex++) {
+			MPlotAbstractTool *tool = tools_->tools().at(toolIndex);
+
+			if (tool) {
+				QWidget *toolView = createToolView(tool);
+				toolViews_ << toolView;
+				layout_->addWidget(toolView);
```
iainworkman: The create tool view seems to have some cases where it can return null, so this here seems to be the source of some "Cannot add null widget to QHBoxView" debug messages. I believe the createToolView needs to either be extended to cover all cases, or if this isn't the intended behaviour this block needs to check if the returned toolView is null before adding it to the layout.

helfrij: Good catch, @iainworkman. I'll push up the change that should fix this.

iainworkman:
Time Estimate: 7 Hours

helfrij:
Time estimate: 5 days

source/ui/dataman/AMScanView.cpp
```diff
@@ -1847,8 +2004,22 @@ void AMScanViewMultiSourcesView::onModelDataChanged(const QModelIndex& topLeft,
 	}
 }
 
+//void AMScanViewMultiSourcesView::onToolSelectionChanged(QList<MPlotAbstractTool*> newSelection)
```
dretrex: Not a big deal, but we missed some commented out code here.

Issue #1460, opened 2015-06-23
Hexapod driver initial framework


iainworkman:
The hexapod driver needs a command line structure setup for command to be run, and for queries to be made to a controller. This will require a console application created which will provide an application user interface for interacting with the controller through the driver.

dretrex:
Time Estimate: 6 hours

dretrex:
Copied by @dretrex
There are a few notes about the application in its current (7da0c24) state:

I had to remove mobility from the config file in acquamanCommon.pri as it seemed to be causing issues. Will look at returning this as we move closer to deploy.
In order for this to build the PI_GCS2 drivers need to be located in ~/beamline/programming/acquaman/contrib/pi_gcs2/lib. Again, closer to deploy this will be brought closer to the standard AM way of including libraries. If anyone needs a copy of the .so file, let me know.
The build requires glibc v2.7 which is not available on SL5.4. We can get access to a virtual opi running SL7.1 for building. For those not using SL5.4 this shouldn't be an issue.
At the moment the AMConsoleApplication stuff I wrote is bundled in the the new AMPIC887DriverServer project. This can (and should) be pulled out in its own project in the fullness of time.
I had to edit the included .h file for the PI GCS2 library. It was missing an include.

Issue #1459, opened 2015-06-23
Hexapod controller communication


iainworkman:
For the hexapod driver to function it must first establish a connection with the controller. This includes getting the controller itself on the SGM VLan, getting the PI_GCS2_DLL driver file into the system and connecting through that libraries connect functions.

dretrex:
Time Estimate: 16 hours

Issue #1458, opened 2015-06-23
Add GCS2 Commands to the Hexapod Server


iainworkman:
The Hexapod controller comes with a set of standard commands. Before any of the more complex work on arbitrary paths can be done we will need to have these initial commands in place and callable from the server.

Note: This was separated from an earlier master issue for the hexapod driver. As such many of the commits were tagged with #1422.

Copied from master issue by @dretrex:

Commands still left to do:
- [x] Move relative
- [x] Set low position soft limit
- [x] Get low position soft limit
- [x] Get on target state
- [x] Set high position soft limit
- [x] Get high position soft limit
- [x] Get position units
- [x] Set cycle time
- [x] Get cycle time
- [x] Set pivot point
- [x] Get pivot point
- [x] Get soft limit status
- [x] Set soft limit status (in but not working)
- [x] Get step size
- [x] Set step size
- [x] Set servo mode
- [x] Get servo mode
- [x] Get min commandable position
- [x] Get max commandable position
- ~~Set closed loop velocity~~ Can only be performed on optional A & B axes (which we're not using)
- ~~Get closed loop velocity~~ Can only be performed on optional A & B axes (which we're not using)
- [x] Set system velocity
- [x] Get system velocity
- [x] Virtual move

iainworkman:
Time Estimate: 66 hours

dretrex:
Done in #1463.  Time estimate?

dretrex:
Ninja'd.

iainworkman:
Hahaha

Issue #1422, opened 2015-06-02
SGM Upgrade - Hexapod Driver


iainworkman:
This master issue was eventually split into 5 smaller issues, for better progress tracking and easier time estimation.

The issues can be found at:
#1460 
#1459 
#1458 
#1461 
#1462 

Edited June 23rd
Pull Request: #1442

iainworkman:
There are a few notes about the application in its current (7da0c24) state:
- I had to remove mobility from the config file in acquamanCommon.pri as it seemed to be causing issued. Will look at returning this as we move closer to deploy.
- In order for this to build the PI_GCS2 drivers need to be located in ~/beamline/programming/acquaman/contrib/pi_gcs2/lib. Again, closer to deploy this will be brought closer to the standard AM way of including libraries.
- The build requires glibc v2.7 which is not available on SL5.4. We can get access to a virtual opi running SL7.1 for building. For those not using SL5.4 this shouldn't be an issue.

dretrex:
Hey @iainworkman, would you mind if I made this a pull request so that it makes code review a little easier?

iainworkman:
Sure, was going to do that, but it slipped my mind.

iainworkman:
Commands still left to do:
- [ ] Move relative
- [ ] Set low position soft limit
- [ ] Get low position soft limit
- [ ] Get on target state
- [ ] Set high position soft limit
- [ ] Get high position soft limit
- [ ] Get position units
- [ ] Set cycle time
- [ ] Get cycle time
- [ ] Set pivot point
- [ ] Get pivot point
- [ ] Set servo mode
- [ ] Get servo mode
- [ ] Get min commandable position
- [ ] Get max commandable position
- [ ] Get number of record tables
- [ ] Set closed loop velocity
- [ ] Get closed loop velocity
- [ ] Set system velocity
- [ ] Get system velocity
- [ ] Virtual move

dretrex:
Since this was split into a bunch of more manageable issues, I'm going to close this one out.

PR #1463, opened 2015-06-23
Pull request for Issue1458: Add GCS2 Commands to the Hexapod Server

iainworkman:
Added the GCS2Commands to the AMPIC887DriverServer
#1458 

This pull request also contains commits from 
#1459 and #1460 as they were all initially part of the larger issue #1422


iainworkman:
- [x] Move relative
- [x] Set low position soft limit
- [x] Get low position soft limit
- [x] Get on target state
- [x] Set high position soft limit
- [x] Get high position soft limit
- [x] Get position units
- [x] Set cycle time
- [x] Get cycle time
- [x] Set pivot point
- [x] Get pivot point
- [x] Get soft limit status
- [x] Set soft limit status (in but not working)
- [x] Get step size
- [x] Set step size
- [x] Set servo mode
- [x] Get servo mode
- [x] Get min commandable position
- [x] Get max commandable position
- ~~Set closed loop velocity~~ Can only be performed on optional A & B axes (which we're not using)
- ~~Get closed loop velocity~~ Can only be performed on optional A & B axes (which we're not using)
- [x] Set system velocity
- [x] Get system velocity
- [x] Virtual move

iainworkman:
Issues identified through testing:
- [ ] DRT? (get data record triggers) needs to include codes in output
- [ ] DRC? (get data record configurations) needs to include codes in output
- [ ] MOV? (get target positions) crashes when called with no axes (backtrace references free())
- [ ] STP (stop) blocks for a long time in the library, then always returns false. The subsequent command also always fails (with a timeout error)
- [ ] HLT (halt) is missing from the help list
- [ ] The subsequent command issued after the HLT (halt) command always fails (with a referencing failed error)
- [ ] SSL still not working. Google turns up nothing, so will attempt to implement as a raw ASCII command.

iainworkman:
Last update didn't fix the error with the stop command. It seems that the time you have to wait until clearing the error message relates to how long it actually takes to stop the axes. The reason I'm thinking this is that altering the system velocity up causes the amount of time that the system has to wait before getting the error message to increase.

iainworkman:
@dretrex I created a small test package to illustrate the problems I was having with the soft limits command, and found I couldn't reproduce it. Spent a good while digging around in my code trying to figure out why it wasn't working before finally tracking the bug down in the get soft limit status command (which I was using to check whether the getter was working). :/
Fixed and tested now.

dretrex:
Okay, I'll try and get to it today.  At least you got it to work though!

acquamanCommon.pri
```diff
@@ -3,7 +3,7 @@
 # ####################################################################
 
 # Video Support: Remove this line if you do not have the multimedia module from QtMobility
-CONFIG *= mobility
+#CONFIG *= mobility
```
dretrex: This will have a negative effect on all the other applications.  Also, even though you comment it out here, you still do `CONFIG -= mobility` in your pro file.

source/application/AMPIC887DriverServer/AMConsoleInputHandler.cpp
```diff
@@ -0,0 +1,64 @@
+#include "AMConsoleInputHandler.h"
+#include "AMAbstractConsoleCommandParser.h"
+
+AMConsoleInputHandler::AMConsoleInputHandler(AMAbstractConsoleCommandParser* commandParser,
+										   QObject* parent)
+	:QObject(parent),
+	  standardOutputStream_(stdout),
```
dretrex: I'm not the biggest fan of initializing member variables in the constructor declaration like this.  Is there a particular reason for doing it this way?  I can also just suck it up if it provides some other inherent benefit (e.g. more efficient, etc.)

source/application/AMPIC887DriverServer/AMGCS2.h
```diff
@@ -0,0 +1,107 @@
...
+	/***************
+	  * Enums
+	  ****************************/
+
+	/*!
+	  * Enumerates all the hexapod axes controlled by a C887.11 controller.
+	  */
+	enum Axis {
+		UnknownAxis = -1,
+		XAxis = 0,
```
dretrex: Although more of an aesthetic comment, you probably don't need to be as pedantic since the enum is `Axis`.  Inside of the `Axis` enum you could just have `X, Y, Z, U, V, W` rather than `XAxis`.  What do you think?

source/application/AMPIC887DriverServer/AMGCS2.h
```diff
@@ -0,0 +1,107 @@
...
+		WAxis
+	};
+
+	/*!
+	  * Enumerates all the axes movement statuses of the hexapod controlled by
+	  * a C887.11 controller.
+	  */
+	enum AxisMovementStatus {
+		UnknownAxisMovementStatus = 1,
+		XAxisIsMoving = 2,
```
dretrex: You could make the same argument here too.

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
```diff
@@ -0,0 +1,497 @@
...
+#include "GCS2Commands/AMGCS2SetStepSizeCommand.h"
+#include "GCS2Commands/AMGCS2GetMinCommandablePositionCommand.h"
+#include "GCS2Commands/AMGCS2GetMaxCommandablePositionCommand.h"
+#include "GCS2Commands/AMGCS2GetSystemVelocityCommand.h"
+#include "GCS2Commands/AMGCS2SetSystemVelocityCommand.h"
+#include "GCS2Commands/AMGCS2VirtualMoveCommand.h"
+
+AMGCS2Command * AMGCS2CommandFactory::buildCommand(const QString &commandString)
+{
+	if(commandString.startsWith("POS?")) {
```
dretrex: There is probably a more efficient way to do this, but this is fine for the time being.  Probably some sort of meta programming could get this done rather simply and efficiently.

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
```diff
@@ -0,0 +1,497 @@
...
+	} else if(commandString.startsWith("VMO?")) {
+		return new AMGCS2VirtualMoveCommand(axesDoublePairFromCommandString(commandString));
+	}
+
+	return 0;
+}
+
+QStringList AMGCS2CommandFactory::commandArguments(const QString& commandString)
+{
+	QStringList argumentsList = commandString.trimmed().split(" ");
```
dretrex: For information purposes only, you could also replace these three lines with:
`return commandString.trimmed().split(" ").mid(1, -1);`

source/application/AMPIC887DriverServer/AMGCS2.h
davidChevrier: I've always struggled with this too. I think that Iain might be right here.
If you want to use this type, then it's AMGCS2::Axis
But if you want to set or compare a value, right now, its AMGCS2::XAxis which seems better than AMGCS2::X
I've taken to being really pedantic with the enum names to avoid the confusion.

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.h
```diff
@@ -0,0 +1,143 @@
...
+	  * string, if the commandString can be parsed, or 0 otherwise.
+	  */
+	static AMGCS2Command* buildCommand(const QString& commandString);
+protected:
+
+	/*!
+	  * Static class, so we protect the contstructor to prevent initialization.
+	  */
+	AMGCS2CommandFactory() {}
+
```
dretrex: No destructor?

source/application/AMPIC887DriverServer/AMGCS2.h
iainworkman: Originally this started as just X, but then once I'd added the movement status and recording source enums, all of them would have contained an AMGCS2::X. In the end I had to be this specific to stop the ambiguity for the compiler as much as the programmer.

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
iainworkman: Yeah, I'd like to set up some sort of registration system, but I've never done any meta programming, and wanted something simple in for the time being.

source/application/AMPIC887DriverServer/AMGCS2Support.h
```diff
@@ -0,0 +1,125 @@
+#ifndef AMGCS2SUPPORT_H
+#define AMGCS2SUPPORT_H
+
+#include <QChar>
+#include <QString>
+#include "AMGCS2.h"
+class AMGCS2Support
```
dretrex: Is there a reason why you went with using a class rather than a namespace?  Since these are all helper functions it seems like it would be a better fit for a namespace.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -0,0 +1,117 @@
...
+		connect(commandParser_, SIGNAL(changeActiveController(QString)), this, SLOT(onActiveControllerChangeIssued(QString)));
+		connect(commandParser_, SIGNAL(GCS2CommandIssued(AMGCS2Command*)), this, SLOT(onGCS2CommandIssued(AMGCS2Command*)));
+
+	} else {
+
+		// Startup failed. Print an error message, then startup a timer to close
+		// the application after a few seconds.
+		consoleInputHandler_->writeLineToStandardError("Could not startup the hexapod driver application.");
+
+		QTimer* exitTimer = new QTimer(this);
```
dretrex: I'm assuming there is a reason for the length of this timer.  What is the criteria for the required time before timing out and quitting the application?

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+void AMPIC887ConsoleApplication::onHelpCommandIssued()
+{
+	consoleInputHandler_->writeLineToStandardOutput(commandParser_->commandList());
+}
+
+void AMPIC887ConsoleApplication::onQuitCommandIssued()
+{
+	consoleInputHandler_->writeLineToStandardOutput("Exiting...");
+	quit();
```
dretrex: Do we not need the timer here too?

source/application/AMPIC887DriverServer/AMConsoleInputHandler.cpp
iainworkman: I can change this. I tend to do this when one of the member variables is stored as a value and has no default parameterless constructor. I can change it, though I'm not sure if it'll involve me switching out the value member variables for references.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+	bool successfulLoad = true;
+
+	successfulLoad &= startupControllers();
+	return successfulLoad;
+}
+
+bool AMPIC887ConsoleApplication::startupControllers()
+{
+	controllerCollection_.append(new AMPIC887Controller("MC1611-401", "10.52.48.26"));
```
dretrex: I know that technically closing the program will get the garbage collector to clean up allocated memory, but is there anything that handles deleting these pointers?

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+	if(controllerCollection_.activeController()) {
```
dretrex: Just because of the simplicity of these statements, you could replace this with `return controller_.activeController();`

source/application/AMPIC887DriverServer/AMPIC887Controller.h
```diff
@@ -0,0 +1,82 @@
...
+  * A class representing a single PI C887.11 Controller. Commands and queries issued
+  * to instances of this class will, in turn, pass them along to the controller
+  * they represent.
+  */
+class AMPIC887Controller
+{
+public:
+	/*!
+	  * Creates a new instance of an C887.11 controller, with the provided hostname
+	  * and id. The controller will not be connected until the connect() call has
```
dretrex: How relevant is the comment _The controller will not be connected until the connect() call has been made._ when you call connect inside the constructor?

acquamanCommon.pri
iainworkman: Including mobility seemed to be causing me compile problems. Not sure if it was because of SL7. I initially tried to remove it from the .pro file of the driver server (to keep the change as limited as possible), but that didn't seem to solve the issue. It was only when I put it here that it finally compiled. I have a checklist of things that I've done to get things working for now, but that will need to be resolved before this is merged into master.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetDataRecorderConfigurationCommand.cpp
```diff
@@ -0,0 +1,76 @@
...
+	} else {
+
+		QStringList axisCharacterList = QString(recordSourceIds.cArray()).split(" \n");
+
+		for (int iRecordConfiguration = 0;
+			 iRecordConfiguration < RECORD_TABLE_COUNT;
+			 ++iRecordConfiguration) {
+
+			dataRecorderConfigurations_.append(
+						new AMPIC887DataRecorderConfiguration(
```
dretrex: Does anything handle these pointers?

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.h
iainworkman: Fixed in bb47bf19cad845886dea807bf8155b78e96a05d7

source/application/AMPIC887DriverServer/AMGCS2Support.h
iainworkman: Fixed in bb47bf19cad845886dea807bf8155b78e96a05d7

source/application/AMPIC887DriverServer/AMConsoleInputHandler.cpp
iainworkman: Done in 67218597d7b459fb6a8b9cd61a35e1683953e73c

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
iainworkman: Fixed in 80a0e13a77c9e61525ca8a9026f74ad52d011827

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
iainworkman: In this case it needs to output to the console. Should require significantly less time than this. I meant to come back later an experiment with lower values, but forgot.

iainworkman: Actually, this one is a relic of a different initial architecture. Neither timers are needed anymore.

iainworkman: Fixed in 8403d4ccdd405dfc8cb9431185f92b8e27468962

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
iainworkman: The controller collection class takes ownership of them.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
iainworkman: Fixed in 70f6b5b990209ef72968dad25b3fb293dd672513

source/application/AMPIC887DriverServer/AMPIC887Controller.h
iainworkman: Fixed in 937aa6dc1b4127cfa1956fa9dcbf87d7a5194a2f

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetDataRecorderConfigurationCommand.cpp
iainworkman: Fixed in 937aa6dc1b4127cfa1956fa9dcbf87d7a5194a2f

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2SetRecordTriggerSourceCommand.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+	return true;
+}
+
+bool AMGCS2SetRecordTriggerSourceCommand::runImplementation()
+{
+	int tableTriggersCount = tableTriggers_.count();
+	AMCArrayHandler<int> recordTableIds(tableTriggersCount);
+	AMCArrayHandler<int> triggerSources(tableTriggersCount);
+	// Library requires dummy values to be included in the call...for some reason.
```
dretrex: Hahaha.

acquamanCommon.pri
dretrex: Ah, alright.

source/application/AMPIC887DriverServer/AMGCS2.h
dretrex: No problem, then.

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
dretrex: @davidChevrier, be a buddy and help a brother out ;)  I'm sure you have plenty of free time.

dretrex:
Overall, this looks great, @iainworkman.  Due to the lack of experience on this from everybody, this code is probably some of the more bulletproof code in the framework.  Most, if not all, of the failure conditions are tackled and dealt with and it is extremely well documented.

If you're done with the couple small things that have been tagged I can merge this guy in.  Let me know.

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
davidChevrier: See:
https://github.com/acquaman/AcquamanDataServer/blob/master/ClientRequest/AMDSClientRequestSupport.h
https://github.com/acquaman/AcquamanDataServer/blob/master/ClientRequest/AMDSClientRequestSupport.cpp

and

https://github.com/acquaman/AcquamanDataServer/blob/master/AMDSTcpDataServer.cpp

iainworkman:
I think the last thing I have to do is figure out that QtMobility thing in the pro file.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2SetRecordTriggerSourceCommand.cpp
iainworkman: At least the manual was kind enough to document this arbitrary insanity.

source/application/AMPIC887DriverServer/AMGCS2.h
iainworkman: Thanks chaps. 
I think what I might do is make a note of this, and get to it at a later date. It will certainly improve efficiency, but feel it can be something done down the road.

If nobody minds I'll create an issue for it, but not include it in the Hexapod milestone (as it isn't fundamental to its workings). This way I won't forget to get to it later.

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
iainworkman: Thanks chaps. 
I think what I might do is make a note of this, and get to it at a later date. It will certainly improve efficiency, but feel it can be something done down the road.

If nobody minds I'll create an issue for it, but not include it in the Hexapod milestone (as it isn't fundamental to its workings). This way I won't forget to get to it later.
@dretrex @davidChevrier Does this seem okay?

iainworkman:
@dretrex Okay, 'last' thing done. Just altered the ordering in the internal .pro file for the driver server, and uncommented out the mobility line in acquamanCommon.pri (to put it back to how it was before).

PR #1472, opened 2015-07-06
Pull Request for Issue1461: Hexapod driver controller abstraction

iainworkman:
Currently this pull request includes the following:
- Implementation of asynchronous commands (Move, MoveRelative and ReferenceMove)
  - Includes  a change in the state machine of the command classes, from a simple wasSuccessful flag to a NotStarted, Running, Successful and Failed system). Only async commands are ever in the Running state, and also emit signals for started(), succeeded() and failed().
  - The error clearing system has been altered, so that it is now the job of the controller class to clear its error message, and set an isBusy_ flag while doing so. Commands will not run on the controller while this process is under way, and async commands will pause checking their done status.
- Change to how commands are obtained from the console (Now the parser interprets them all and emits signals. The application class catches these signals and runs the required function on the controller)
- Implemented caching for all the data of the controller. This is initialized on start up, and then updated whenever a command is issued which alters the state of the controller 
- The system through which the asynchronous move commands update the movement statuses of the axes in the cache has been implemented. This is done by means of a type of semaphore system, which counts the number of moves issued vs failed/succeeded.
- The command factory class has been deprecated (will be removed)
- Functions for all commands has been added to the controller class.

@dretrex This is now ready for review. 
Estimate: 8


dretrex:
Want to throw an estimate for this guy @iainworkman?

source/application/AMPIC887DriverServer/AMGCS2CommandFactory.cpp
```diff
@@ -52,91 +52,93 @@
```
iainworkman: This whole class can be removed. It isn't used anymore.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,298 @@
+#ifndef AMPIC887AXISMAP_H
+#define AMPIC887AXISMAP_H
+
+#include <QMap>
+#include <QHash> //TEMP
```
iainworkman: This include can be removed.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -70,23 +66,517 @@ void AMPIC887ConsoleApplication::onActiveControllerChangeIssued(const QString &c
 	}
 }
 
-void AMPIC887ConsoleApplication::onGCS2CommandIssued(AMGCS2Command* command)
```
iainworkman: This function can be removed.

dretrex:
Hey @iainworkman, were you wanting to do some modifications before I jump into this guy or should I start now?

iainworkman:
As soon as it's built I'll push up the changes and send you a message.

iainworkman:
@dretrex Okay, I've pushed up those changes.

dretrex:
Alright.  I'll start on it then.

source/application/AMPIC887DriverServer/AMPIC887AxisCollection.h
```diff
@@ -0,0 +1,77 @@
...
+  * A class representing a collection of axes controlled by a PI C887.11 Controller
+  */
+class AMPIC887AxisCollection : public QList<AMGCS2::Axis>
+{
+public:
+	// Enumerates the different states the collection can be initialized in.
+	enum InitializationState {
+		EmptyCollection,	// Collection is initialized with no contained axes
+		AllAxes,			// Collection is initialized with all axes
+		LinearAxes			// Collection is initialized with linear axes (X, Y, Z)
```
dretrex: Is it not possible to initialize with rotational axes (u, v, w)?

source/application/AMPIC887DriverServer/AMPIC887AxisCollection.h
```diff
@@ -0,0 +1,77 @@
...
+	};
+
+	// Enumerates the different valid states which the colleciton can be in
+	enum ValidState {
+		Valid,					// Collection is valid
+		ContainsUnknownAxis,	// Collection contains an unknown axis
+		ContainsDuplicateAxes	// Collection contains duplicates
```
dretrex: Is there a general `Invalid` state? Or does unknown act as a catch all here?

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+  * A class which maps Axes of a PI C887.11 controller to a type T.
+  */
+template <typename T>
+class AMPIC887AxisMap : public QMap<AMGCS2::Axis, T>
+{
+public:
+	/*!
+	  * Creates an instance of an axis map, which is empty.
+	  */
+	AMPIC887AxisMap() {}
```
dretrex: Do you not use the default QMap constructor?

```c++
AMPIC887AxisMap()
     : QMap<AMGC2::Axis, T>()
{
}
```

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+
+	/*!
+	  * Returns a collection of the axes which this map contains.
+	  */
+	AMPIC887AxisCollection axes() const
+	{
+		AMPIC887AxisCollection returnCollection(AMPIC887AxisCollection::EmptyCollection);
+		QList<AMGCS2::Axis> axisEntries = this->keys();
```
dretrex: You probably don't need the _this_ keyword here.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+
+		foreach(AMGCS2::Axis currentAxisEntry, axisEntries) {
```
dretrex: Are you not able to just set this via the constructor?  `return AMPIC887AxisCollection(keys());`

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+		// Implemented for known types in partial specializations below.
+		return QString();
+	}
+
+	/*!
+	  * Whether this map contains an entry for the UnknownAxis.
+	  */
+	bool containsUnknownAxis() const
+	{
+		return this->contains(AMGCS2::UnknownAxis);
```
dretrex: Probably don't need _this_ here either.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+		}
+
+		return returnCollection;
+	}
+
+	/*!
+	  * Outputs the map values as a human readable string, if it can parse the
+	  * type into a string, otherwise returns the empty string.
+	  */
+	QString toString(const QString& = "Data") const
```
dretrex: Does this need to be virtual or are you just overloading it in other places?

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+template <>
+class AMPIC887AxisMap<bool> : public QMap<AMGCS2::Axis, bool>
+{
+public:
+	/*!
+	  * Creates an instance of an axis map, which is empty.
+	  */
+	AMPIC887AxisMap() {}
```
dretrex: Same comment here about the constructor.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+
+	/*!
+	  * Returns a collection of the axes which this map contains.
+	  */
+	AMPIC887AxisCollection axes() const
+	{
+		AMPIC887AxisCollection returnCollection(AMPIC887AxisCollection::EmptyCollection);
```
dretrex: Same comment here.  I'll stop repeating myself if I find other instances.  However, how often do you call `axes()`?  If it's really often, could this be done in such a way that you don't have to generate a fresh list every time you called it?  I could image a situation where you have to repeatedly call this function which would be much faster as some sort of cached set of axes.  Comments?

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+	  */
+	QString toString(const QString& header = "Data") const
+	{
+		QString returnString = QString("Axis\t\t%1\n").arg(header);
+
+		QList<AMGCS2::Axis> axes = this->keys();
+		foreach(AMGCS2::Axis currentAxis, axes) {
+			QString axisValue;
+			if(this->value(currentAxis)) {
+				axisValue = "Yes";
```
dretrex: This whole loop could probably be reduced to a single line of code.  You don't have to make the modification, but I thought I'd bring it up.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
```diff
@@ -0,0 +1,297 @@
...
+				returnValues.insert(currentAxis, this->value(currentAxis));
+			}
+		}
+
+		return returnValues;
+	}
+};
+
+template <>
+class AMPIC887AxisMap<double> : public QMap<AMGCS2::Axis, double>
```
dretrex: Repeat comments for all equivalent code in the following classes.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -70,23 +66,507 @@ void AMPIC887ConsoleApplication::onActiveControllerChangeIssued(const QString &c
...
+void AMPIC887ConsoleApplication::onControllerErrorEncountered(const QString &errorMessage)
+{
+	consoleInputHandler_->writeLineToStandardError(errorMessage);
+}
+
+void AMPIC887ConsoleApplication::onSetCommandLevelCommandIssued(AMGCS2::ControllerCommandLevel commandLevel,
+																const QString &password)
+{
+	AMPIC887Controller* activeController = controllerCollection_.activeController();
+	if(!activeController) {
```
dretrex: I believe I mentioned somewhere else recently.  I prefer positive conditional formatting that doesn't use returns to stop bad input/state access.  Therefore this method would be rewritten as:

```c++
AMPIC887Controller *activeController = controllerCollection_.activeController();

if (activeController && !activeController->setCommandLevel(commandLevel, password))
     consoleInputHandler_->writeLineToStandardError(activeController->lastError());
```

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -70,23 +66,507 @@ void AMPIC887ConsoleApplication::onActiveControllerChangeIssued(const QString &c
...
+	}
+
+	if(!activeController->setSystemVelocity(systemVelocity)) {
+		consoleInputHandler_->writeLineToStandardError(activeController->lastError());
+	}
+}
+
+void AMPIC887ConsoleApplication::onMoveCommandIssued(const AMPIC887AxisMap<double> &axisPositions)
+{
+	if(!controllerCollection_.activeController() && !controllerCollection_.activeController()->isInitialized()) {
```
dretrex: This is probably supposed to be an OR?  If you use the style I mentioned earlier then you could just DeMorgan this conditional and it would be fine.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -70,23 +66,507 @@ void AMPIC887ConsoleApplication::onActiveControllerChangeIssued(const QString &c
...
+		return;
+	}
+
+	consoleInputHandler_->writeLineToStandardError(
+				controllerCollection_.activeController()->limitSwitchStatuses(axes).toString("Limit Switch Active"));
+}
+
+void AMPIC887ConsoleApplication::onMinPositionCommandIssued(const AMPIC887AxisCollection &axes)
```
dretrex: Not a big deal, but there is really no need to abbreviate Min and Max.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
```diff
@@ -99,10 +579,80 @@ bool AMPIC887ConsoleApplication::startup()
...
 
 	return (controllerCollection_.activeController() != 0);
 }
 
+void AMPIC887ConsoleApplication::makeConnections()
+{
+	connect(commandParser_, SIGNAL(help()), this, SLOT(onHelpCommandIssued()));
+	connect(commandParser_, SIGNAL(quit()), this, SLOT(onQuitCommandIssued()));
+	connect(commandParser_, SIGNAL(status()), this, SLOT(onStatusCommandIssued()));
+	connect(commandParser_, SIGNAL(changeActiveController(QString)), this, SLOT(onActiveControllerChangeIssued(QString)));
```
dretrex: To be consistent with every other signal/slot combo you made this could probably be `activeControllerChanged(QString)` and `onActiveControlChanged(QString)`.

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -1,28 +1,75 @@
...
+QString AMPIC887Controller::lastError() const
+{
+	return lastError_;
+}
+
+bool AMPIC887Controller::connectToController()
 {
 	id_ = PI_ConnectTCPIP(hostname_.toStdString().c_str(), CONTROLLER_PORT);
-	return connectionEstablished();
+	bool success = connectionEstablished();
```
dretrex: Unless there is more to come, it would be simpler for `return connectionEstablished()'`.

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+
+	return setRecordConfigs(singleValue);
+}
+
+QHash<int, AMPIC887DataRecorderConfiguration> AMPIC887Controller::recordConfigs() const
+{
+	return controllerState_->dataRecorderState()->recordConfigs();
+}
+
+bool AMPIC887Controller::setRecordConfigs(const QHash<int, AMPIC887DataRecorderConfiguration> recordConfigs)
```
dretrex: `const QHash<> &`

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+	}
+
+	controllerState_->dataRecorderState()->setRecordConfigs(recordConfigs);
+
+	return true;
+}
+
+QFlags<AMGCS2::AxisMovementStatus> AMPIC887Controller::movementStatus() const
+{
+	QFlags<AMGCS2::AxisMovementStatus> movements = QFlags<AMGCS2::AxisMovementStatus>();
```
dretrex: Although not necessary, you probably could have made the Qt equivalent of a typedef for these flags and just used `AMGCS2::AxisMovementStatuses`

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+	if(wMotions_ > 0) {
+		movements |= AMGCS2::WAxisIsMoving;
+	}
+
+	return movements;
+}
+
+bool AMPIC887Controller::isMoving() const
+{
+	return xMotions_ > 0 ||
```
dretrex: You could use `QFlags::testFlag()` here if you mapped these to full QFlags rather than just using an enum.

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+{
+	return controllerState_->hexapodState()->maxCommandablePositions().values(axes);
+}
+
+bool AMPIC887Controller::onTargetState(AMGCS2::Axis axis)
+{
+	double currentAxisPosition = currentPosition(axis);
+	double currentTargetPosition = targetPosition(axis);
+
+	double epsilon = 0.0001;
```
dretrex: Is this a hardcoded number because it seems to work/you read it somewhere or is it just a placeholder?

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+		return false;
+	}
+
+	controllerState_->hexapodState()->setStepSizes(stepSizes);
+	return true;
+}
+
+double AMPIC887Controller::systemVelocity() const
+{
+	return controllerState_->hexapodState()->velocity();
```
dretrex: I've noticed that for a lot of these getters you aren't very concerned with making sure you're not accidentally dereferencing bad pointers whereas you're very pedantic in checking in other areas.  Is there a reason why you elected to not be as pedantic here?

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+
+	runCommand(asyncReferenceMoveCommand);
+}
+
+void AMPIC887Controller::onAsyncMoveStarted(AMGCS2AsyncCommand *command)
+{
+	currentPositionRefreshRequired_ = true;
+	AMGCS2AsyncMoveCommand* moveCommand = qobject_cast<AMGCS2AsyncMoveCommand*>(command);
+
+	if(moveCommand) {
```
dretrex: What happens if this fails?  Is it possible this can cause the system to hang?

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+			this, SLOT(onAsyncReferenceMoveStarted(AMGCS2AsyncCommand*)));
+	connect(asyncReferenceMoveCommand, SIGNAL(failed(AMGCS2AsyncCommand*)),
+			this, SLOT(onAsyncReferenceMoveFailed(AMGCS2AsyncCommand*)));
+	connect(asyncReferenceMoveCommand, SIGNAL(succeeded(AMGCS2AsyncCommand*)),
+			this, SLOT(onAsyncReferenceMoveSucceeded(AMGCS2AsyncCommand*)));
+
+	runCommand(asyncReferenceMoveCommand);
+}
+
+void AMPIC887Controller::onAsyncMoveStarted(AMGCS2AsyncCommand *command)
```
dretrex: Is this command used for when composite moves (say x and w or x, y, z, u, v) are done?

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
```diff
@@ -61,10 +114,954 @@ QString AMPIC887Controller::status() const
...
+
+void AMPIC887Controller::onAsyncReferenceMoveFailed(AMGCS2AsyncCommand *command)
+{
+	--xMotions_;
+	--yMotions_;
+	--zMotions_;
+	--uMotions_;
+	--vMotions_;
+	--wMotions_;
+
```
dretrex: Is anything required to be set to the hexapod state if it fails?  Or does it handle that itself?

source/application/AMPIC887DriverServer/AMPIC887ControllerState.cpp
```diff
@@ -0,0 +1,67 @@
...
+AMPIC887DataRecorderState * AMPIC887ControllerState::dataRecorderState() const
+{
+	return dataRecorderState_;
+}
+
+void AMPIC887ControllerState::initialize(AMGCS2::ControllerCommandLevel commandLevel,
+										 const QString &availableParameters,
+										 const QString& identification)
+{
+	currentCommandLevel_ = commandLevel;
```
dretrex: Is it possible that these available parameters and identification can be not valid causing this to not actually become initialized?

source/application/AMPIC887DriverServer/AMPIC887ControllerState.cpp
```diff
@@ -0,0 +1,67 @@
...
+	availableParameters_ = availableParameters;
+	identificationString_ = identification;
+	isInitialized_ = true;
+}
+
+bool AMPIC887ControllerState::isAllInitialized() const
+{
+	return hexapodState_->isInitialized() &&
```
dretrex: Is this initialized different from the `isInitialized_`?  Is there a reason why it's not included?

source/application/AMPIC887DriverServer/AMPIC887DataRecorderState.cpp
```diff
@@ -0,0 +1,92 @@
...
+bool AMPIC887DataRecorderState::isInitialized() const
+{
+	return isInitialized_;
+}
+
+void AMPIC887DataRecorderState::initialize(AMGCS2::DataRecordTrigger recordTrigger,
+										   const QString& availableRecordParameters,
+										   const QHash<int, AMPIC887DataRecorderConfiguration>& recordConfigs)
+{
+	recordTrigger_ = recordTrigger;
```
dretrex: Same initialization questions applies to this class too.

source/application/AMPIC887DriverServer/AMPIC887HexapodState.cpp
```diff
@@ -0,0 +1,454 @@
...
+	AMPIC887AxisCollection axes = currentPositions.axes();
+	foreach(AMGCS2::Axis currentAxis, axes) {
+
+		currentPositions_.insert(currentAxis, currentPositions.value(currentAxis));
+	}
+}
+
+double AMPIC887HexapodState::targetPosition(AMGCS2::Axis axis) const
+{
+	return targetPositions_.value(axis, 0);
```
dretrex: Will it cause any errors if you somehow returned 0 when you didn't want to because the axis wasn't valid?

source/application/AMPIC887DriverServer/AMPIC887HexapodState.cpp
```diff
@@ -0,0 +1,454 @@
...
+void AMPIC887HexapodState::setPivotPoint(AMGCS2::Axis axis, double pivotPoint)
+{
+	if(axis == AMGCS2::UnknownAxis ||
+			axis == AMGCS2::UAxis ||
+			axis == AMGCS2::VAxis ||
+			axis == AMGCS2::WAxis) {
+		return;
+	}
+
+	pivotPoints_.insert(axis, pivotPoint);
```
dretrex: Weren't there more rules for setting pivot points?  Something like u, v, and w all had to be 0 as well?

source/application/AMPIC887DriverServer/AMPIC887HexapodState.h
```diff
@@ -0,0 +1,367 @@
...
+	  * NOTE: This does not initialize the data for the contained axis states. In
+	  * order for isAllInitialized() to return true, all the axis states must also
+	  * be initialized with data.
+	  * \param servoMode ~ Whether the axes of the hexapod are in servo mode.
+	  * \param cycleTime ~ The cycle time for the hexapod.
+	  * \param velocity ~ The current axis velocity.
+	  */
+	void initialize(bool servoMode, double cycleTime, double velocity,
+					AMPIC887AxisMap<bool> referencedStates,
+					AMPIC887AxisMap<double> currentPositions,
```
dretrex: Missing `const X &` in argument definition?

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2AsyncCommand.cpp
```diff
@@ -0,0 +1,85 @@
...
+		return;
+	}
+
+	if(startTime_.msecsTo(QTime::currentTime()) >= timeoutMs_) {
+		runningState_ = Failed;
+		lastError_ = "Async command timer timed out";
+		killTimer(timerId_);
+		emit failed(this);
+	} else {
+		isFinishedImplementation();
```
dretrex: Is it okay for you to be calling `isFinishedImplementation()` here even though you could still be running?

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2AsyncMoveCommand.cpp
```diff
@@ -0,0 +1,157 @@
...
+		lastError_ = "Could not obtain moving status of axes";
+		runningState_ = Failed;
+		return;
+	}
+
+	QFlags<AMGCS2::AxisMovementStatus> movementStatuses =
+			movingStatusCommand.movementStatuses();
+
+	bool allStillMoving = true;
+	foreach(AMGCS2::Axis currentAxis, axesMoved) {
```
dretrex: Just so I know you're sure, if you were doing a move with multiple axes (say x, y, u, v).  This will will become false if any of the axes becomes false but not all of them.  For example, allStillMoving will be false in the situation where x = false, y = true, u = true, and v = true.  From what I can tell from reading the rest of this function, you only want to continue if all the axes have stopped, not when only one has stopped.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2AsyncMoveCommand.cpp
```diff
@@ -0,0 +1,157 @@
...
+			targetPositionCommand.axisTargetPositions();
+
+	AMPIC887AxisMap<double> finalPositions = currentPositionCommand.axisPositions();
+
+	foreach (AMGCS2::Axis currentAxis, axesMoved) {
+		double destination = targetPositions_.value(currentAxis);
+		double finalPosition = finalPositions.value(currentAxis);
+		double currentTargetPosition = currentTargetPositions.value(currentAxis);
+
+		double epsilon = 0.0001;
```
dretrex: Same question about this epsilon from earlier.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2AsyncMoveRelativeCommand.cpp
```diff
@@ -0,0 +1,177 @@
...
+		return;
+	}
+
+	QFlags<AMGCS2::AxisMovementStatus> movementStatuses =
+			movingStatusCommand.movementStatuses();
+
+	bool allStillMoving = true;
+	foreach(AMGCS2::Axis currentAxis, axesMoved) {
+		switch(currentAxis) {
+		case AMGCS2::XAxis:
```
dretrex: Same deal here.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetCurrentPositionCommand.h
```diff
@@ -30,13 +31,8 @@ class AMGCS2GetCurrentPositionCommand : public AMGCS2Command
 	  * the last time the command was run. If the command has not been run, or was
 	  * not successful when last run, then an empty map is returned.
 	  */
-	QHash<AMGCS2::Axis, double> axisPositions();
+	AMPIC887AxisMap<double> axisPositions();
```
dretrex: If you're expecting to call this really often, you could probably make this a `const AMPIC887AxisMap<double> &axisPositions() const;` method since `axisPositions_` is a member of this class.  That way you don't make copies every time you call it.  If this class isn't long lived then it's not a good idea, but just thought I'd throw it out there.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetLowSoftLimitsCommand.cpp
```diff
@@ -2,50 +2,34 @@
...
 }
 
-QHash<AMGCS2::Axis, double> AMGCS2GetLowSoftLimitsCommand::axesLowSoftLimits() const
+AMPIC887AxisMap<double> AMGCS2GetLowSoftLimitsCommand::axesLowSoftLimits() const
 {
 	return axesLowSoftLimits_;
 }
 
-QString AMGCS2GetLowSoftLimitsCommand::outputString() const
+bool AMGCS2GetLowSoftLimitsCommand::validateArguments()
```
dretrex: It looks like this validate arguments method is exactly the same in all subclasses.  There is probably a way to simplify that code so that it's not repeated over and over again.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetMinCommandablePositionCommand.cpp
```diff
@@ -51,56 +38,29 @@ bool AMGCS2GetMinCommandablePositionCommand::validateArguments()
 
 bool AMGCS2GetMinCommandablePositionCommand::runImplementation()
```
dretrex: Honestly, it almost looks like if you could somehow pass the handler for each command you wouldn't even need to have these subclassed/implemented runImplementations().

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2InitializeControllerStateCommand.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+		lastError_ = "Cannot load data for a controller while its hexapod is in motion";
+		return false;
+	}
+
+	return true;
+}
+
+
+bool AMGCS2InitializeControllerStateCommand::runImplementation()
```
dretrex: Is this the method that checks to make sure that the hexapod initializes correctly (the comment X,000 lines of code ago)....

dretrex:
Okay, first pass completed.  Overall this looks great @iainworkman.  The main comments are stylistic/syntactic or they are clarification questions that may/may not require follow up based on the responses.
Time estimate: 3 hours.

source/application/AMPIC887DriverServer/AMPIC887AxisCollection.h
iainworkman: Theoretically I could add this for completeness, but there are commands which require only linear axes (set pivot point) but no commands which require only rotational axes.

source/application/AMPIC887DriverServer/AMPIC887AxisCollection.h
iainworkman: True, that would probably be a good addition. I enumerated all the invalid states I'd yet encountered, but a general case would be nice.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
iainworkman: This one took me a long time to track down. Turns out templates aren't just confusing to the programmer, but the compiler as well. It needs the 'this' pointer.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
iainworkman: I'm 'overloading' it through template specialization. This case catches the general case of types T which aren't bool, double or position units.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
iainworkman: Yes, I've been wondering about this myself. It is incredibly useful here to have the map return its keys in the form of a collection. I could do this with a constructor in the collection which takes a QList<AMGCS2::Axis> I believe.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2AsyncCommand.cpp
iainworkman: I realize now that's ambiguously named
semantically it means
is finished?
not
is finished.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetMinCommandablePositionCommand.cpp
iainworkman: Yes. I have been thinking about that a lot. There are certainly at least a few families of these commands, with the only difference in each family being the eventual call to the PI library. Will give this some though, because not maintaining all those command classes sounds nice to me.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
iainworkman: The same result will be produced either way 
http://stackoverflow.com/questions/5780093/c-should-i-call-the-base-class-default-constructor-in-the-initialization-li
I've added it in anyway for clarity.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
iainworkman: Okay, did a bit more research into this.
It turns out that in order to call a base class function, the compiler needs to know which base it is from. In the case of a template with unspecified type T there is no way of determining this. The compiler doesn't yet know the type of T which is specified, and it doesn't assume that it will necessarily be the same as the instantiated template type from which this is called. As such with yet unknown template types you have to specify 'this' so that it knows to call the keys() function for the same template type.

tldr. I can remove the this pointer from all the specialization classes below, but not the top one for type T.

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
iainworkman: This one confused me a bit. 
You're right, the code you posted works, but what's confusing me is that there is no constructor in AMPIC887AxisColleciton which takes a type QList<AMGCS::Axis>.

Does this mean that you can always construct a derived class by passing a base instance to its constructor?

iainworkman: Okay, must have screwed up somewhere. Now it's not working. 
I'll add that constructor to the axis collection.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
iainworkman: Looking at all these setters (which all suffer from the same issue) I could also do with outputting an error message if no active controller was found. Would you be okay with:
```C++
AMPIC887Controller* activeController = controllerCollection_.activeController();

if(!activeController) {
    consoleInputHandler_->writeLineToStandardError("No active controller to issue command to.");
} else if(!activeController->setCommandLevel(commandLevel, password) {
    consoleInputHandler_->writeLineToStandardError(activeController->lastError()
}
```
???

source/application/AMPIC887DriverServer/AMPIC887AxisMap.h
dretrex: Okay cool.

source/application/AMPIC887DriverServer/AMPIC887ConsoleApplication.cpp
dretrex: That would also definitely be okay.  Feel free to make those changes (... so many changes...)

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2AsyncCommand.cpp
dretrex: Oh.... to make it easier on future us, could you rename that method to more accurately describe what it's doing?  isFinishedImplementation has a tone of finality to it that can lead to misunderstandings.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetMinCommandablePositionCommand.cpp
dretrex: I agree.

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
iainworkman: The controllerState_ is a value member, so that can't ever be null, and while the hexapodState() is a reference, it's a reference whose lifetime is tied to that of the controllerState_.

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
iainworkman: It worked in testing, though that was a while ago.

The thing with this value is that it represents a limit of certainty on the position of the axes from this point higher. The position will always be +/- this epsilon.

With the importance of that fact in mind I've altered these to be a 
``` C++
#define AXIS_POSITION_TOLERANCE 0.0001
```
in the controller class, and then changed all these vague references to epsilon to AXIS_POSITION_TOLERANCE. That way, when I later come to test this value to see how far I can push it I only ever have to change it in one place, and it should remain consistent across the whole system.

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
iainworkman: If the command fails? No, it won't hang the system. The failed() signal will be emitted after a timeout value has been exceeded, and there's nothing blocking about the async commands when running.

dretrex: No, if the moveCommand doesn't qobject_cast.  Would the timer save us in that case too?

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
iainworkman: All the referenced states of the axes are taken care of by setting their referenced states to false when the reference move begins. 

I've added a call to updateStateOnStop() which refreshes the current position values, and sets the target position to the current position.

source/application/AMPIC887DriverServer/AMPIC887HexapodState.cpp
iainworkman: At the moment none of these singular versions are used anywhere. I left them in because I'm pretty certain I'm going to need them later on. I've specified the behaviour in the comment docs for the functions, so will have to make sure I read that, and check for that condition when I come to use them.

source/application/AMPIC887DriverServer/AMPIC887HexapodState.cpp
iainworkman: This class just stores the current pivot points. The actual check should (read doesn't) happen in the AMPIC887Controller::setPivotPoints class. I'll add that in.

source/application/AMPIC887DriverServer/AMPIC887ControllerState.cpp
iainworkman: I've added checks for these now.

source/application/AMPIC887DriverServer/AMPIC887DataRecorderState.cpp
iainworkman: Again, added checks for these now.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2GetCurrentPositionCommand.h
iainworkman: This (and in fact all the get commands) are run only once in the initialization of the controller. All other times the values are obtained from the cache (and the caches are updated each time a set command is called).

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
iainworkman: This would be a bit of a weird one, really. If the command isn't a move command, then the moveCommand variable will end up null after the cast, and nothing will be incremented. This is technically correct in as far as effecting these states is concerned, as if it isn't a movement command we shouldn't be altering the movement states. 

However if it isn't something really funky is happening - someone is connecting the started() signal of a command which is not a move command to the onMoveCommandStarted slot(). It's impossible to say what effect this would have without knowing whether this mystery command was also connected to its correct slots (in case all would be well), and if it wasn't what command it actually is.

The only way I could really imagine of coping with this would be to emit a error signal from the controller which would output to the console a description of what was happening.

source/application/AMPIC887DriverServer/GCS2Commands/AMGCS2InitializeControllerStateCommand.cpp
iainworkman: Technically it's a function which loads all the current state data for the controller into the its data cache (this happens when the controller is constructed).

I should really add a validate function to the controller state, and think about where the checks for that should be.

source/application/AMPIC887DriverServer/AMPIC887Controller.cpp
dretrex: If you think it is highly implausible then I wouldn't worry about it.

iainworkman:
Take 2 TODOs:
- [ ] Create a validate controller function, and add checks for valid state into relevant places
- [ ] Break down commands into families based on arguments, and figure out how to get the PI library call into them (function pointer?)

iainworkman:
@dretrex This is ready for review now.

The diff from the last commit at the point the first review was done, to this commit can be found [here](https://github.com/acquaman/acquaman/compare/cba4c3e...8868cad)

iainworkman:
Tom and Teak are done using the hexapod, so I got some time to do testing this morning. Fixing up some stuff as I go. Will alter the label for this once I'm done.

iainworkman:
@dretrex  This is ready for review again.

dretrex:
Okay.  If that's everything, I think we can merge this in.  Any further comments, @iainworkman?

iainworkman:
Nope, looks good to go.

dretrex:
Time estimate for all your changes?

iainworkman:
Time Estimate: 17.5 Hours

PR #1473, opened 2015-07-07
Pull Request for Issue1435: Creating scan-based energy calibration routine

helfrij:


helfrij:
I think this pull request may be at a point were a preliminary review might be helpful. Progress so far:
- refactored the beamline and XAS scan-related classes to group together functionality that Side and Main have in common. This should help expand the functionality of Main considerably and reduce the number of errors caused by minor differences between Side and Main.
- created new BioXASSSRLMonochromatorEnergyCalibrationXASScan family of classes that include some basic energy calibration specific information.
- created new BioXASSSRLMonochromatorEnergyCalibrationView that provides an interface for loading and manipulating XAS scan information, and calibrating the mono.

@dretrex, would a review at this point be okay? Or is it better to wait until all work is complete?

dretrex:
A review here would totally be fine. I'll go over it soon. @iainworkman this might be a good opportunity for you to have a look at some scan controller code. Would you review it as well?

Sent from my iPhone

On Jul 22, 2015, at 9:36 AM, helfrij <notifications@github.com<mailto:notifications@github.com>> wrote:


I think this pull request may be at a point were a preliminary review might be helpful. Progress so far:

  *   refactored the beamline and XAS scan-related classes to group together functionality that Side and Main have in common. This should help expand the functionality of Main considerably and reduce the number of errors caused by minor differences between Side and Main.
  *   created new BioXASSSRLMonochromatorEnergyCalibrationXASScan family of classes that include some basic energy calibration specific information.
  *   created new BioXASSSRLMonochromatorEnergyCalibrationView that provides an interface for loading and manipulating XAS scan information, and calibrating the mono.

@dretrex<https://github.com/dretrex>, would a review at this point be okay? Or is it better to wait until all work is complete?

-

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
```diff
@@ -0,0 +1,380 @@
...
+		if (braggMotor) {
+			monoInitialization = new AMSequentialListAction3(new AMSequentialListActionInfo3("BioXAS Monochromator Initialization", "BioXAS Monochromator Initialization"));
+			monoInitialization->addSubAction(braggMotor->createPowerAction(CLSMAXvMotor::PowerOn));
+		}
+	}
+
+	// Initialize the standards wheel.
+
+	AMAction3* standardsWheelInitialization = 0;
+//	CLSStandardsWheel *standardsWheel = BioXASBeamline::bioXAS()->standardsWheel();
```
dretrex: Have these just been disabled for the time being or was there something wrong with them?

source/acquaman/BioXAS/BioXASXASScanConfiguration.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+void BioXASXASScanConfiguration::setupDefaultEXAFSRegions()
+{
+	// Clear previous regions.
+
+	clearRegions();
+
+	// Create new regions and add to the list of regions for the first scan axis.
+
+	if (scanAxisAt(0)) {
```
dretrex: If there is no valid scan axis, would it make sense to create and then setup the regions regardless?

source/acquaman/BioXAS/BioXASXASScanConfiguration.h
```diff
@@ -0,0 +1,79 @@
...
+	/// Clears all regions.
+	void clearRegions();
+	/// Sets up the default XANES regions.
+	void setupDefaultXANESRegions();
+	/// Sets up the default EXAFS regions.
+	void setupDefaultEXAFSRegions();
+
+protected slots:
+	/// Inserts the given region into the given scan axis index position, and makes the appropriate connections.
+	void addRegion(int scanAxisIndex, int regionIndex, AMScanAxisRegion *region);
```
dretrex: Would it make sense to have these in AMGenericStepScanConfiguration?  They could be virtual if you needed them to be, but I wonder if these could live in a more general location so that other scans could take advantage of it.

source/beamline/AMPseudoMotorControl.h
```diff
@@ -65,6 +70,8 @@ public slots:
 	virtual FailureExplanation move(double setpoint);
 	/// Stops the control, by stopping all children.
 	virtual bool stop();
+	/// Calibrates the control such that the old value becomes the new value. Fails if calibration has not been implemented for this control.
+	virtual FailureExplanation calibrate(double oldValue, double newValue);
```
dretrex: This is a cool addition.  Good thinking.

source/beamline/BioXAS/BioXASSideBeamlineUtilities.h
```diff
@@ -0,0 +1,126 @@
...
+	// Flow switches.
+
+
+
+	// Temperature monitors.
+
+protected:
+	// Pressure controls
+
+	AMControl *ccgSide1_;
```
dretrex: All of these `AMControl *`'s could be their specific `AMXYZControl *`.  This would simplify some of your state checking code in BioXASBeamlineUtilities class.

dretrex:
This is a really robust refactor and is incredibly thorough.  Great work, @helfrij!  This is awesome.  I'll just wait on the couple responses, but other than that this code looks solid to me.

dretrex:
Time Estimate: 50 minutes

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
helfrij: The standards wheel is disabled on the beamline temporarily. As far as I know, there is no problem with the device itself. A better thing to do might be to comment out the creation of the standards wheel in BioXASSideBeamline, instead of the initialization actions here.

source/acquaman/BioXAS/BioXASXASScanConfiguration.h
helfrij: It's tough to tell sometimes in comments--do you mean the add/connect/remove/disconnectRegion slots?

source/acquaman/BioXAS/BioXASXASScanConfiguration.cpp
helfrij: Hmmm I'm not sure. I think a scan axis is required in order to add a region to a configuration, as in addRegion below.

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
dretrex: Yeah, that would probably be a better idea.  Then when we bring it back it will already be working in the various scans.

source/acquaman/BioXAS/BioXASXASScanConfiguration.cpp
dretrex: Sorry, I missed the last piece of my thought in that comment.  What I meant was, would it make sense to create a scan axis if one doesn't exist and then add regions like you would expect?

source/acquaman/BioXAS/BioXASXASScanConfiguration.h
dretrex: Yup, those are the ones.

source/beamline/BioXAS/BioXASSideBeamlineUtilities.h
helfrij: I've modified the variable types to match their constructed types, but I'm not sure where I see a simplification can be made. Can you explain further?

source/acquaman/BioXAS/BioXASXASScanConfiguration.cpp
helfrij: Ah! Okay, let's see. Scan axes are created when a scan control for that axis is set, in AMGenericStepScanConfiguration::setControl. If there is no valid scan axis, that means that a control hasn't been set, but might be set in the future. Regions are actually preserved when a control is set. If it's acceptable to potentially have regions set when there is no control to scan, I can definitely make that change.

source/acquaman/BioXAS/BioXASXASScanConfiguration.h
helfrij: Good idea, will move them.

source/beamline/BioXAS/BioXASSideBeamlineUtilities.h
dretrex: You should now not need to cast to cast the control when you use it in your "on error" methods and need to check to see if it casts correctly.

source/acquaman/BioXAS/BioXASXASScanConfiguration.cpp
dretrex: Ah, I see.  I think in that case then, it is better to leave it as it is.

dretrex: Or it's okay if we want to change the paradigm, but we would need extra validation in that case though.  What do you think would be a better future-proof way to go?

source/beamline/BioXAS/BioXASSideBeamlineUtilities.h
helfrij: I think that cast may be necessary, as AMControlSet always returns AMControl pointers when asked for a control at a particular index.

source/acquaman/BioXAS/BioXASXASScanConfiguration.cpp
helfrij: It's tough to say. I kind of sidestep this issue on BioXAS by initializing all XAS scan configurations in the app controller--there's never an XAS configuration displayed that doesn't have an appropriate control and regions already provided. It might be useful to always have regions available, even if they're default "nonsense" regions. However, I think it makes more conceptual sense to say that regions are tied more closely to the scanned control than they are inherent properties of scan configurations. But it's kind of a toss-up either way.

dretrex:
Okay, all merged up.  @helfrij, could you add a time estimate for the changes?

helfrij:
Thanks for your comments, @dretrex. Will do.

helfrij:
Time Estimate: 1 hour.

PR #1475, opened 2015-07-11
Pull Request for Issue1457: Integrate Scaler Continuous Mode

helfrij:
I've created a new control CLSSIS3820ScalerModeControl that manages the current scaler mode. It uses the scan count, number of scans, and start scan pvs provided by the scaler.

The new control has been added to CLSSIS3820Scaler and seems to work as intended when changing values through the ScalerControlsView interface. Have yet to test in a scan. Knowing whether or not this control implementation works more reliably than the pv version will be tricky, as we won't know until an error situation crops up.


source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -227,7 +282,7 @@ void AMPseudoMotorControl::onMoveFailed(QObject *action)
 	emit moveFailed(AMControl::OtherFailure);
 }
 
-void AMPseudoMotorControl::onMoveSucceeded(QObject *action)
+void AMPseudoMotorControl:: onMoveSucceeded(QObject *action)
```
dretrex: It probably doesn't matter, but there probably shouldn't be a space here.

source/beamline/CLS/CLSSIS3820Scaler.cpp
```diff
@@ -328,7 +333,16 @@ void CLSSIS3820Scaler::onScanningToggleChanged(){
 	emit scanningChanged(startToggle_->withinTolerance(1));
 }
 
-void CLSSIS3820Scaler::onContinuousToggleChanged(){
+void CLSSIS3820Scaler::onContinuousToggleChanged()
+{
+	qDebug() << "\nScaler mode changed.";
```
dretrex: We can keep this until we run some tests on the beamline with scans, but we should remove this before merging it in.

source/beamline/CLS/CLSSIS3820ScalerModeControl.cpp
```diff
@@ -0,0 +1,270 @@
...
+				scanCountControl_ && scanCountControl_->isConnected() &&
+				numberOfScansPerBufferControl_ && numberOfScansPerBufferControl_->isConnected() &&
+				startScanControl_ && startScanControl_->isConnected()
+				);
+
+	setConnected(isConnected);
+
+	// If we weren't connected before but we are now, set mode to Continuous.
+
+	if (!wasConnected && isConnected)
```
dretrex: Is there a reason why we are specifically putting it into continuous?  It should probably be set to whatever mode it was in last.

source/beamline/CLS/CLSSIS3820ScalerModeControl.cpp
```diff
@@ -0,0 +1,270 @@
...
+					startScanControl_->isMoving()
+					);
+
+		setIsMoving(isMoving);
+	}
+}
+
+void CLSSIS3820ScalerModeControl::onMoveSucceeded(QObject *action)
+{
+	AMPseudoMotorControl::onMoveSucceeded(action);
```
dretrex: Is the order of these two important?  Could there be issues of the control thinking it has succeeded before the setpoint is set?  Just curious.

dretrex:
Other than the couple comments, it looks good.  Lets put this on the side endstation and run some scan tests to verify that it works correctly before merging it in.

source/beamline/CLS/CLSSIS3820ScalerModeControl.cpp
helfrij: This code serves to put the scaler into Continuous mode after the child controls are set and report as connected. Though perhaps this should be done a done a different way?

source/beamline/CLS/CLSSIS3820ScalerModeControl.cpp
helfrij: I hadn't thought of that--it's probably better to be safe and reverse the order.

helfrij:
The control seems to work as would be expected in a scan.

The qDebugs() that were added to CLSSIS3820Scaler::onContinuousToggleChanged() show that the scaler mode is reporting a mode change for every move in an XAS scan. The mode control shouldn't be reporting that the value has changed when it really hasn't (new mode is always reported as SingleShot).

helfrij:
One problem came up that I hadn't thought about--running multiple instances of Acquaman. I just opened an instance of BioXAS Side on the Imaging OPI, while the other instance was running test scans on the Side user area OPI. Starting up the second instance caused the scaler pvs to assume 'Continuous' mode values for both instances, which effectively halted the scans happening on Side (without warnings or alerts). The Side instance scaler still reported being in SingleShot mode, which is still locally accurate but it no longer reflected the actual scaler functioning.

source/beamline/CLS/CLSSIS3820ScalerModeControl.cpp
dretrex: I would say that they probably shouldn't be changed at all.  If it's connected, it should check all the PVs to determine which state it is in, rather than forcing a new state.  This would also deal with your comment below about multiple instances of AM.

In general, I would say that nothing should change state on startup.  It should read what it's states are and try its best to tell you where it is.

helfrij:
Implemented @dretrex's recommendations, the scaler mode control appears to behave as expected using the scaler view and in scans. If there's nothing further, I think this pull request is ready to merge.

dretrex:
Time estimate?

dretrex:
Time Estimate: 15 minutes

PR #1485, opened 2015-07-16
Pull Request for Issue 1481: Scaler hangs

helfrij:
- Added extra step to set scaler to NotScanning in createMoveToContinuousModeAction().
- Moved existing step to set scaler to NotScanning to first step in createMoveToSingleShotAction()
- Removed some XAS scan initialization steps related to scaler setup--they don't appear to be necessary.


dretrex:
Hey @helfrij, yes I think that this is a viable solution.  The 'scaler-punches' might not be necessary anymore since I think the main reason why they were added in the first place was to deal with the continuous mode of the scaler.  They might not even be needed at all anymore.

helfrij:
Time estimate: 30 minutes.

dretrex:
Time estimate: 10 minutes.

Issue #1486, opened 2015-07-16
Votex Detector data issue duing EXAFS scan


XuetaoLiu-CLS:
On the SolidState endstation on SXRMB there appears to be some issues with the data on Votex Detector when doing EXAFS scan.

dretrex:
Can you be a little more explicit in the description of the issue?

XuetaoLiu-CLS:
Time Estimate: 4 days

PR #1488, opened 2015-07-17
Pull Request for Issue1486: Add dataReady check for the 4-element Votex Detector

XuetaoLiu-CLS:


dretrex:
@XuetaoLiu-CLS, have you been able to test this at all?  Have you heard any more complaints that would require more immediate action on this issue?

XuetaoLiu-CLS:
@dretrex This PR is still in work progress. I didn't have an opportunity to play with the Votex yet. I talked with SXRMB staffs yesterday and looks like I didn't have any opportunity in the near future as well.

Once we have everything tested, those commented codes should be removed.

dretrex:
Okay.

XuetaoLiu-CLS:

source/beamline/SXRMB/SXRMBFourElementVortexDetector.cpp
```diff
@@ -55,6 +76,75 @@ QString SXRMBFourElementVortexDetector::synchronizedDwellKey() const
...
+	detectorCleanupListActions->addSubAction(AMXRFDetector::createCleanupActions());
+
+	AMListAction3 *scanPVCleanupListActions = new AMListAction3(new AMListActionInfo3("Votex Detector Scan PV Cleanup", "Votex Detector Scan PV Cleanup Actions"), AMListAction3::Parallel);
+	detectorCleanupListActions->addSubAction(scanPVCleanupListActions);
+
+	AMListAction3 *scanPVCleanupActions = 0;
+
+	// add the MCA cleanup actions
+	scanPVCleanupActions = new AMListAction3(new AMListActionInfo3("Votex MCA cleanup", "Votex MCA cleanup"), AMListAction3::Sequential);
+	scanPVCleanupActions->addSubAction(AMActionSupport::buildControlMoveAction(mcaRatePVControl_, mcaRatePVControl_->value()));
```
helfrij: It looks like the control move actions here move a control to its current position. Just checking if that's intentional.

helfrij:
One question. Looks good @XuetaoLiu-CLS.

Time estimate: 15 minutes.

source/beamline/SXRMB/SXRMBFourElementVortexDetector.cpp
XuetaoLiu-CLS: @helfrij yes, we will create a cleanup action here, which will restore the PV to the value it was before the scan. This function will be executed before the scan. In this way, when we use pv->value(), we will keep the current value for the clean up action. When the scan is done, it will move the PV to that value.

helfrij: Ah, okay. It's possible for that setpoint to be variable then? I mean, not the same every time you execute this method?

XuetaoLiu-CLS: For scan cleanup, we suppose to set the setpoint to whatever it was before the scan. That's the reason we use the pv->value() here instead of a variable.

source/application/SXRMB/SXRMB.h
```diff
@@ -43,7 +43,8 @@ namespace SXRMB {
 	{
 		NoXRFDetector = 0,
 		BrukerDetector = 1,
-		FourElementDetector = 2
+		FourElementDetector = 2,
+		BrukerAndFourElementDetector = 3
```
dretrex: By having these as QFlags, you don't need to add this enum like this.  Instead you can do:

```c++
QFlag flag = detectorFlag();

if (flag.testFlag(SXRMB::FourElement) && flag.testFlag(SXRMB::Bruker))
   // Do stuff
```

Then, all you need to do is make sure your enum is always a binary sequence then you can use logic + testFlag() for everything.

dretrex:
One comment.  Otherwise the data ready aspect looks good.

Time Estimate: 5 minutes

XuetaoLiu-CLS:
https://github.com/acquaman/acquaman/issues/1486

Issue #1487, opened 2015-07-16
Generic Motor Group/View For 1-6 degrees of motion


iainworkman:
When setting up the new SGM myself and Darren noticed that the motor group and corresponding view copes well with up to 3 axes, but must be extended to do more (see AM4DMotorGroup, SGMSampleManipulatorMotorGroup). It would be nice if the group and corresponding view coped well with anything from 1 to 6 degrees of freedom across the 3 axes.

iainworkman:
I have a proposal for this. It would mean one new class, and changes to the existing AMMotorGroupObject, MotorGroup and MotorGroupView.

## AMMotorGroupAxis
Designation::AxisDesignation (enum values XAxis, YAxis, ZAxis, Theta, Psi, Phi)
Units::QString (eg. "mm", "deg" etc.)
Control::AMControl*
Name::QString (eg. "Z", "Y", "Up", "Yaw", "In", etc.)

## AMMotorGroupObject
Name::QString (eg. "Sample Stage" etc.)
Axes:: QMap<AMMotorGroupAxis::AxisDesignation, AMMotorGroupAxis*>

##AMMotorGroup
GroupObjects::QMap<QString, AMMotorGroupObject*>

Formatting the view would take a bit more logic, but could be fairly easily worked out if implicit relationships were assumed between translational and rotational axes (ie. alpha rotates around z, gamma rotates around y and beta rotates around x)

dretrex:
Looks really good to me.  If you want, you could potentially use pitch, roll, and yaw right in the enum since I think those terms are conceptually broader than their greek counterparts.

iainworkman:
Time Estimate: 29.5 hours

PR #1496, opened 2015-07-22
Pull Request for Issue1487: Generic Motor Group/View For 1-6 degrees of motion

iainworkman:
#1487 
Alters the structure of the motor group to more generically cope with 1-6 degrees of motion.
Redesign of the motor group view to better display the new motor group structure.
Done:
- [x] Implement new structure of motor group
- [x] Refactor existing instantiations of motor groups in the beamlines to meet new requirements.
- [x] Add all widgets to new motor group view
- [x] Get widgets to reflect the data of the motor group, with updates when required.
- [x] Get widgets in motor group view to alter the data of the controls when clicked/interacted with.
- [x] Refactor CLSPseudoMotorGroupView

Estimate: 3


iainworkman:
Okay, there is only a certain amount of this I can test on my machine, so I could do with at least viewing it on VESPERS (mostly for the compact space it requires and use of pseudo motors) and making sure it issues commands properly to the relevant controls.

dretrex:
We'll do the test on Monday.

source/application/SGM/SGMAppController.cpp
```diff
@@ -1489,10 +1489,18 @@ bool SGMAppController::setupSGMPeriodicTable(){
 	return success;
 }
 
+#include "ui/AMMotorGroupView.h"
```
dretrex: I know this was just for testing, but could you move this to the top of the file with the rest of the includes?

source/beamline/AMMotorGroup.cpp
```diff
@@ -25,286 +25,779 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
...
-// AMMotorGroupObject
-////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+// Motor Group Motion
+////////////////////////////////////////////////////////////////////////////////
 
- AMMotorGroupObject::~AMMotorGroupObject(){}
-AMMotorGroupObject::AMMotorGroupObject(const QString &name, const QString &prefix, const QString &units, AMControl *control, Orientation orientation, MotionType motionType, QObject *parent)
-	: QObject(parent)
+AMMotorGroupMotion::AMMotorGroupMotion(const QString& name,
```
dretrex: This has really weird spacing.

source/beamline/AMMotorGroup.cpp
```diff
@@ -25,286 +25,779 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
...
+									   AMControl* motor)
+	: QObject(0)
```
dretrex: Is there a reason for switching `parent` with `0`?

source/beamline/AMMotorGroup.cpp
```diff
@@ -25,286 +25,779 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
...
 
-int AMMotorGroupObject::verticalIndex() const
+
+////////////////////////////////////////////////////////////////////////////////
+// Motor Group Axis
+////////////////////////////////////////////////////////////////////////////////
+
+AMMotorGroupAxis::AMMotorGroupAxis(AMMotorGroupMotion* translationalMotion,
+								   AMMotorGroupMotion* rotationalMotion)
+	: QObject(0)
```
dretrex: Why do we not include parent?

source/beamline/AMMotorGroup.cpp
```diff
@@ -312,17 +805,59 @@ AMMotorGroup::AMMotorGroup(QObject *parent)
...
+void AMMotorGroup::addMotorGroupObject(AMMotorGroupObject *object)
 {
-	infoMap_.insert(name, object);
+	if(object == 0) {
+		return;
+	}
+
+	AMMotorGroupObject* currentObject = motorGroupObject(object->name());
+
+	if(currentObject == object) {
```
dretrex: I think if you want to do something like this, you should let the caller deal with the memory.  I would change this to `bool` function and if it returns false, the caller will know that they need to deal with the object.

source/beamline/AMMotorGroup.h
```diff
@@ -27,153 +27,831 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
...
+	  * Sets the override values for the position units for the rotational motions.
+	  * \param positionUnits ~ The position units to associate with the rotational
+	  * movements.
+	  */
+	void setRotationPositionUnits(const QString& positionUnits);
+
+	/*!
+	  * Stops motions of the translation motor.
+	  */
+	void stopTranslation();
```
dretrex: In general, all of the action items for a motor group object should be slots.  That way (in particular with stop) there is a very straight forward "button click -> mapped directly to stop".  It also makes it more convenient for programmatic calls more succinct rather than having to make their own slot just call stop or move.  For all versions of move/stop at all levels, they should likely be modified into slots.

source/beamline/AMMotorGroup.h
```diff
@@ -27,153 +27,831 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
...
+	/*!
+	  * Creates an action which will stop all translational and rotational movement
+	  * across all associated axes within the group object.
+	  */
+	AMAction3* createStopAllAction();
+
+	/*!
+	  * Instructs all axes within the group object to cease their motions.
+	  */
+	void stopAll();
```
dretrex: Just flagging this as part of previous comment about stop as a slot.

source/ui/AMMotorGroupView.cpp
```diff
@@ -17,520 +17,1042 @@ GNU General Public License for more details.
...
+AMMotorGroupView::AMMotorGroupView(AMMotorGroup* motorGroup,
+								   ViewMode viewMode,
+								   QWidget* parent)
+	: QWidget(parent)
 {
-	if (viewMode_ == Exclusive){
+	motorGroup_ = motorGroup;
 
-		menu->addAction("Switch to Multiple View");
+	if(!motorGroup_) {
```
dretrex: Aesthetically, I'm not a big fan of returns in constructors.  I think it would be "better" if it was:

```
if (motorGroup_){
     ...
}
```

source/ui/AMMotorGroupView.cpp
```diff
@@ -17,520 +17,1042 @@ GNU General Public License for more details.
...
 
-		QStringList visibleNames;
+	// Set up the stuff we are going to use.
+	QVBoxLayout* mainLayout = new QVBoxLayout();
 
-		foreach (AMMotorGroupObjectView *view, motorGroupViews_.values())
-			if (view->isVisible())
-				visibleNames << view->motorGroupObject()->name();
+	groupObjectSelector_ = new QComboBox();
+	groupObjectStack_ = new QStackedWidget();
```
dretrex: I like this as a solution rather than my old one where I just hid them.  Nice change up.

source/util/SGM/SGMSettings.cpp
```diff
@@ -44,7 +44,7 @@ void SGMSettings::load(){
 #ifdef Q_WS_MAC
 	SGMDataFolder_ = settings.value("SGMDataFolder", "/Users/fawkes/Documents/CLS/SGM/ACQUAMANDATA").toString();
 #else
-	SGMDataFolder_ = settings.value("SGMDataFolder", "/home/sgm/beamline/programming").toString();
+	SGMDataFolder_ = settings.value("SGMDataFolder", "/home/workmai/beamline/programming").toString();
```
dretrex: Is this okay that it's only tied to your account or do we have to worry about this at a later date since you've replaced the main sgm account in the path?

dretrex:
Overall, this looks really good.  The only issue I'd like to bring up is the parenting issue.  I think there is a discontinuity in the parenting even though sub objects are parented using `setParent(this)` I don't think all the objects were accounted for.  Although not the end of the world, I'd prefer if all of the objects were parented properly so that we can continue to utilize the QObject garbage clean up.

Time estimate: 1.5 hours.

iainworkman:
Yeah, if there was a part of the grouping that I didn't account for that was a mistake. The intention was that, because there was a fairly explicit hierarchy of structure from motion >> axis >> object >> group I made it so that groups just automatically took ownership of objects, which in turn owned their axes, which owned their motions. I'll double check and make sure that everything is accounted for (as well as fix up the few other issues).

I noted a few issues when building the full suite that cropped up due to the refactoring, which I'm currently working on.

source/util/SGM/SGMSettings.cpp
iainworkman: Yeah, this is one of those changes that I have to make from my machine each time I pull down SGM and wish to run it. I usually revert this file before I push it back up to the remote. Must have forgotten this time. It'll get changed back before merging.

source/beamline/AMMotorGroup.cpp
iainworkman: Okay, here's how I'm thinking this will go for resource ownership, in that case:
- AMMotorGroup takes no ownership of it's contained AMMotorGroupObjects. The above addMotorGroupObject(...) function now returns a reference to an AMMotorGroupObect that has been replaced, or 0 if no replacement took place.
- AMMotorGroupObjectss still take ownership of their three AMMotorGroupAxes. This class is so closely tied to the motor group object that it would be an odious overhead to have the user manually keep track of these resources separate from the group object.
- Similarly AMMotorGroupAxes take ownership of their AMMotorGroupMotions.
- None of these classes takes any ownership of the contained AMControl's. After all these are just convenience organizational classes.

iainworkman:
@dretrex Are we still on for doing those tests today?

dretrex:
Yes, that's the plan.

Sent from my iPhone

On Jul 31, 2015, at 9:47 AM, iainworkman <notifications@github.com<mailto:notifications@github.com>> wrote:


@dretrex<https://github.com/dretrex> Are we still on for doing those tests today?

-

dretrex:
Is this all done now, @iainworkman?

iainworkman:
No, I need to look at what the best control for the situation is. AMExtendedMotorControl would be perfect, but the header that it automatically displays makes it too big for the space sensitive cases. I'm thinking it might be best to make some changes to the extended motor control so it doesn't extend group box, as that's what's making it take up too much room.

-----Original Message-----

iainworkman:
@dretrex This is now ready for a last check over.

dretrex:
I've done the merge @iainworkman.  Feel free to add merge master into SGMUpgrade to get it more up-to-date.

iainworkman:
Time Estimate: 17.5 Hours

PR #1503, opened 2015-07-27
Pull Request for Issue1494: Display the y value of a clicked position in AMScanView.

helfrij:


helfrij:
This pull request is being closed because it is no longer needed. The issue it is meant to address is handled in #1963.

PR #1506, opened 2015-07-29
Pull Request for Issue1491: VESPERS: Needs a way to have the SMAK exporter available for non-auto-exported scans

dretrex:
Stored a SMAK exporter to the database, but held off registering the exporter.  I think this should work to get around this bug.


dretrex:
Can you look at this pull request with your eyes for me @iainworkman?

iainworkman:
Looks good to me. I'm assuming that the boolean variables passed are correct for the exporter type based on the commented out code sharing the same values.

For the commented out code, I assume you're going to leave that in in hopes of a future general fix for the issue? If not it could be removed.

dretrex:
You are correct on both accounts.

dretrex:
Took a fair amount of testing to isolate that I needed to add a check.  Original idea worked though, so that's promising.

Time estimate: 1.5 hours

PR #1510, opened 2015-07-31
Pull Request for Issue1417: XAS scans can(not) run backwards

epengr:
Setting min value for all spinboxes to 0 instead of -10,000.  The will
prevent backwards XAS scans, and other silly behaviours.


dretrex:
Time estimate: 1 minute.

PR #1515, opened 2015-08-05
Pull Request for Issue1465: Crash in AMStepScanActionController

dretrex:
Ensured that the file writer and file writer threads are properly nulled so that they can be checked against to make sure that they don't get dereferenced at a later time.


dretrex:
Although bad form, I'll do this pull request.
Time estimate: 1 minute

PR #1516, opened 2015-08-05
Pull Request for Issue1514: Open up other databases in read-only mode

dretrex:
This allows the opening of other databases inside of an instance of a running application. It also contains a large amount of small changes to AMDatabase and AMDbObject to facilitate general database access and tweaking some classes to introduce a read-only database loading.

@davidChevrier and @iainworkman, could both of you go through this pull request since it has a very large amount of database tweaks?

List of major changes:
- Added another means of using the read-only flag in AMDatabase.
- Removed the static keyword from execQuery().  This required the modification of some const functions.
- Added the ability to load databases inside of AMDatamanAppController.
- Changed AMDatabase and AMDbObject to not save changes to databases opened in read-only mode.
- Fixed a bug in AMLightweightScanInfo where it was calling the user database even though it had a valid database pointer.
- Modified the loadData() method in AMScan to reflect the database it is being loaded from rather than using the AMUserSettings::userDataFolder.
- Added a shutdown method to AMDatamanAppControllerForActions3 to delete the actions and scanActions database.
- Some general code cleanup I saw along the way.

### DIFF
--- a/source/dataman/database/AMDatabase.cpp
+++ b/source/dataman/database/AMDatabase.cpp
@@ -39,24 +39,32 @@ QMutex AMDatabase::databaseLookupMutex_(QMutex::Recursive);
 
 // This constructor is protected. Only access is through AMDatabase::createDatabase().
  AMDatabase::~AMDatabase(){}
-AMDatabase::AMDatabase(const QString& connectionName, const QString& dbAccessString) :
-	QObject(),
+AMDatabase::AMDatabase(const QString& connectionName, const QString& dbAccessString, bool openAsReadOnly)
+	: QObject(),
 	connectionName_(connectionName),
 	dbAccessString_(dbAccessString),
 	qdbMutex_(QMutex::Recursive)
 {
-	QFileInfo accessInfo(dbAccessString_);
-	if(accessInfo.exists())
-		isReadOnly_ = !accessInfo.isWritable();
-	else
-		isReadOnly_ = false; // Probably not correct. If this the call that creates a new database, when does the .db file actually get created?
+
+	if (openAsReadOnly)
+		isReadOnly_ = true;
+
+	else {
+
+		QFileInfo accessInfo(dbAccessString_);
+		if(accessInfo.exists())
+			isReadOnly_ = !accessInfo.isWritable();
+		else
+			isReadOnly_ = false; // Probably not correct. If this the call that creates a new database, when does the .db file actually get created?
+	}
 
 	// Make sure the database is initialized in the creating thread:
 	qdb();
 
 }
 
-AMDatabase* AMDatabase::createDatabase(const QString &connectionName, const QString &dbAccessString) {
+AMDatabase* AMDatabase::createDatabase(const QString &connectionName, const QString &dbAccessString, bool openAsReadOnly)
+{
 	QMutexLocker ml(&databaseLookupMutex_);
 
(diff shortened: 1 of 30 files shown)


source/dataman/database/AMDatabase.cpp
```diff
@@ -850,9 +859,15 @@ bool AMDatabase::supportsTransactions() const
 	return qdb().driver()->hasFeature(QSqlDriver::Transactions);
 }
 
-
+#include <QDebug>
```
iainworkman: Does this qDebug still need to be here?

source/dataman/database/AMDatabase.cpp
```diff
@@ -39,24 +39,32 @@ QMutex AMDatabase::databaseLookupMutex_(QMutex::Recursive);
...
 {
-	QFileInfo accessInfo(dbAccessString_);
-	if(accessInfo.exists())
-		isReadOnly_ = !accessInfo.isWritable();
-	else
-		isReadOnly_ = false; // Probably not correct. If this the call that creates a new database, when does the .db file actually get created?
+
+	if (openAsReadOnly)
+		isReadOnly_ = true;
+
```
iainworkman: This is a really minor quibble, but it seems odd to have the if() .. missing the braces but have the  else include them. Would be easy to screw up editing this section when maintaining this in future.

iainworkman:
Couple of issues that I've already discussed with @dretrex, but recording them here for reference:
- What happens when a db which is being opened needs an upgrade?
- There seems to be a few current paradigms for passing around/accessing databases:
    - Through AMDatabase::database("hardcodedStringLiteral");
    - Passing the name of the database to the function/class.
    - Passing a reference to the database in question
```C++
void AMClass::doSomething()
{
     AMDatabase* userDatabase = AMDatabase::database("user");
    // Use userDatabase
}
```

```C++
void AMClass::doSomething(const QString& databaseName)
{
    AMDatabase* database = AMDatabase::database(databaseName);
    // Use database
}
```

```C++
void AMClass::doSomething(AMDatabase* database)
{
    // Use database
}
```

Now, the first one scares me a little bit, but I can see it at least being functionally separate from the other two (in that it covers cases where a class will always use one specific database, like the actions or scanAcitons database). The second two seem pretty much equivalent, though so it would be nice to adopt a standard means of doing this.

dretrex:
I also need to modify the scan selection dialog used inside of the generic scan editor.  It currently only looks at the user database.

dretrex:
After looking into it, I am not going to do anything about the upgrades for the database because if a load failed because the database was not upgraded, it would fail to load.  This will work for the time being since the upgrades situation is a messy one to say the least.

iainworkman:
Just to make sure, the upgrades are written in a manner which means a failed upgrade isn't going to put them in some kind of broken state?

dretrex:
No, because at it's core, even if something tried to update them, they would need to make non-select/pragma statements which is strictly disallowed by execQuery for read-only databases.

iainworkman:
Okay, sounds good then.

dretrex:
Okay, I have modified the open scan dialog.  This _should_ be the last change to this pull request aside from comments.  I have already done the couple small things you've noted already, @iainworkman, but if you could go over these other changes too that would be cool.

iainworkman:
Looks good to me.
Time Estimate: 1.5 hours

dretrex:
Time estimate: 8 hours.

PR #1519, opened 2015-08-06
Pull Request for Issue1517: Generic scan config view not showing correct start and end values

iainworkman:
Added a check in the constructor of AMGenericStepScanConfigurationView which checks if a valid value already exists for each axis, and only sets defaults if they don't

Estimate: 1

### DIFF
--- a/source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
+++ b/source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
@@ -106,8 +106,15 @@ AMGenericStepScanConfigurationView::AMGenericStepScanConfigurationView(AMGeneric
 
 	positionsBox->setLayout(positionLayout);
 
-	onAxisControlChoice1Changed();
-	onAxisControlChoice2Changed();
+	// Set initial spinbox values in cases where we don't yet have axes:
+	if(configuration_->scanAxes().count() == 0) {
+		// Set defaults for axis 1 and 2
+		onAxisControlChoice1Changed();
+		onAxisControlChoice2Changed();
+	} else if(configuration_->scanAxes().count() == 1) {
+		// Set defaults just for axis 2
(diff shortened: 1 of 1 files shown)


dretrex:
Don't forget to provide an estimate and assign it to someone.

iainworkman:
I tried using a curl script that David provided me for setting the estimates, but it complained about certificates. It's about a 1. I'll try and remember to set it once I get home tonight.

dretrex:
I set it for you, no worries.

source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
```diff
@@ -106,8 +106,15 @@ AMGenericStepScanConfigurationView::AMGenericStepScanConfigurationView(AMGeneric
 
 	positionsBox->setLayout(positionLayout);
 
-	onAxisControlChoice1Changed();
-	onAxisControlChoice2Changed();
+	// Set initial spinbox values in cases where we don't yet have axes:
+	if(configuration_->scanAxes().count() == 0) {
```
dretrex: My only question with this is whether or not it properly displays and enables everything.  If someone had a 2D scan would both axes show up enabled?

dretrex:
Aside from the one comment it looks good.
Time estimate: 5 minutes

source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
iainworkman: I've tested the same change on the SGMUpgrade branch, and it behaves correctly with new scans with no controls yet selected, loaded 1d scan configurations and loaded 2d scan configurations.

dretrex: Okay, great.

PR #1528, opened 2015-08-11
Pull Request for Issue1423: Creating Hexapod Driver to AM Epics Interface

iainworkman:
Conceptually the most basic motions of the hexapod. It provides an interface with the hexapod from Acquaman which treats each of the axes of the hexapod as a separate motor which is to be controlled individually. Coordinated motions of the hexapod are to follow.
### Driver Side
- Created an epics db file which contains standard records for the X, Y, Z, U, V and W axes (feedback, setpoint and status), a stop all PV and system velocity (setpoint and feedback) PVs.
- Created a new class AMPIC887EpicsCoordinator which initializes the PV values with the controller values and startup, and then acts as an adapter to pass setpoint from PV to controller, and changed values from controller to PV.
- Added signals in the AMPIC887Controller class which fire on updates to motion status (start, complete, failed), changes to the system velocity and a timer controller periodic update of the current position when a motion is active (so the coordinator can push these to the PVs).

### SGM Application Side
- Added AMPVwStatusControls to the SGMBeamline for the hexapod PVs.
- Added the hexapod position PVs to the beamline's motor group
- Added to hexapod position PVs to the beamline's exposed controls
- Added an AMExtendedControlEditor for the system velocity to the SGMPersistentView
- Added a check to ensure that only my username and the sgm upgrade username can run the new version of Acquaman (stops users from running it and borking their database)


iainworkman:
Okay, now I've merged in SGMUpgrade branch to fix the merge conflicts this is ready for review @dretrex 
(It also took it from 8000+ lines of addition to just under 1000)

source/application/SGM/SGMAppController.cpp
```diff
@@ -30,6 +30,9 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 #include "ui/CLS/CLSSIS3820ScalerView.h"
 #include "acquaman/AMGenericStepScanConfiguration.h"
 #include "ui/acquaman/AMGenericStepScanConfigurationView.h"
+#include "util/AMErrorMonitor.h"
+
+#include <stdlib.h> // Used for obtaining username to prevent users other than iain (for dev) or SGM-Upgrade (for commissioning). Remove for deploy.
```
dretrex: I'm assuming this will stay for the time being @iainworkman?

dretrex:
Looks good.  After your response to my comment I'll merge this in.
Time estimate: 30 minutes

source/application/SGM/SGMAppController.cpp
iainworkman: Yeah, it's just there for the getenv("USER") function, which acts as a check to stop someone, somehow managing to run the development version on an account which has old acquaman data.

iainworkman:
I'll fix up these merge conflicts

iainworkman:
Time Estimate: 1 hour

PR #1530, opened 2015-08-12
Pull Request for IainBuild

iainworkman:
This is simply a branch which alters some of the common pri files to get Acquaman to build on my laptop. 
I pulled this down on an OPI and tested the build there and everything works fine.

Estimate: 1


dretrex:
Time estimate: 1 minute

Issue #1435, opened 2015-06-08
Implement scan-based mono calibration procedure


helfrij:
The BioXAS team would like to be able to perform an XAS scan and then calibrate the mono based on the scan results.

helfrij:
Time Estimate: 3 months

PR #1537, opened 2015-08-18
Pull Request for Issue1435DbUpgrade

helfrij:
Here is the first draft of the database upgrade needed to make sure data in old BioXASSide- and BioXASMainConfigurations are still accessible. These classes are functionally replaced with BioXASXASScanConfiguration in #1435.

### DIFF
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -3,6 +3,8 @@
 #include "beamline/BioXAS/BioXASBeamline.h"
 #include "beamline/CLS/CLSStorageRing.h"
 
+#include "dataman/BioXAS/BioXASDbUpgrade1Pt1.h"
+
 #include <QDebug>
 
 BioXASAppController::BioXASAppController(QObject *parent) :
@@ -12,6 +14,11 @@ BioXASAppController::BioXASAppController(QObject *parent) :
 
 	userConfiguration_ = new BioXASUserConfiguration(this);
 	setDefaultUseLocalStorage(true);
+
+	// Database upgrades.
+
+	AMDbUpgrade *bioxas1pt1 = new BioXASDbUpgrade1Pt1("user", this);
+	appendDatabaseUpgrade(bioxas1pt1);
 
 	// Initialize member variables.
 
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -17,8 +17,10 @@ BioXASAppController::BioXASAppController(QObject *parent) :
 
 	// Database upgrades.
 
-	AMDbUpgrade *bioxas1pt1 = new BioXASDbUpgrade1Pt1("user", this);
-	appendDatabaseUpgrade(bioxas1pt1);
+	AMDbUpgrade *bioxas1pt1UserDb = new BioXASDbUpgrade1Pt1("user", this);
+	appendDatabaseUpgrade(bioxas1pt1UserDb);
+	AMDbUpgrade *bioxas1pt1ActionsDb = new BioXASDbUpgrade1Pt1("actions", this);
+	appendDatabaseUpgrade(bioxas1pt1ActionsDb);
 
 	// Initialize member variables.
 
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -434,7 +434,7 @@ AMScanConfigurationViewHolder3* BioXASAppController::createScanConfigurationView
 
 			AMGenericStepScanConfiguration *stepScanConfiguration = qobject_cast<AMGenericStepScanConfiguration*>(configuration);
 			if (stepScanConfiguration) {
-				connect( stepScanConfiguration, SIGNAL(totalTimeChanged()), view, SLOT(updateOverallScanTime(double)) );
+				connect( stepScanConfiguration, SIGNAL(totalTimeChanged(double)), view, SLOT(updateOverallScanTime(double)) );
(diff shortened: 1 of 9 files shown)


source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -13,6 +15,11 @@ BioXASAppController::BioXASAppController(QObject *parent) :
 	userConfiguration_ = new BioXASUserConfiguration(this);
 	setDefaultUseLocalStorage(true);
 
+	// Database upgrades.
+
+	AMDbUpgrade *bioxas1pt1 = new BioXASDbUpgrade1Pt1("user", this);
+	appendDatabaseUpgrade(bioxas1pt1);
```
dretrex: Not a big deal for initial testing, but you will need to do this with the actions database too.  Due to old design constraints, the user and actions databases need to be the same (since they both can recall scan configurations).

source/dataman/BioXAS/BioXASDbUpgrade1Pt1.cpp
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+
+	// Create and populate BioXASXASScanConfiguration_table, if it doesn't already exist.
+
+	if (!databaseToUpgrade_->tableExists("BioXASXASScanConfiguration_table")) {
+		QStringList names = QStringList() << "AMDbObjectType" << "thumbnailCount" << "thumbnailFirstId" << "name" << "detectorConfigurations" << "axisControlInfos" << "configurationDbObject" << "header";
+		QStringList types = QStringList() << "TEXT" << "INTEGER" << "INTEGER" << "TEXT" << "TEXT" << "TEXT" << "TEXT" << "TEXT";
+
+		if (!databaseToUpgrade_->ensureTable("BioXASXASScanConfiguration", names, types)) {
```
dretrex: I would use the `AMDbUpgradeSupport::addTable()` method instead since it handles all the support table changes as well.

source/dataman/BioXAS/BioXASDbUpgrade1Pt1.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+		if (databaseToUpgrade_->insertOrUpdate(0, "AMDbObjectTypes_visibleColumns", names, types)) {
+			databaseToUpgrade_->rollbackTransaction();
+			AMErrorMon::alert(this, BIOXASDBUPGRADE1PT1_COULD_NOT_INSERT_OR_UPDATE_TABLE, "Could not add BioXASXASScanConfiguration entry to AMDbObjectType_visibleColumns.");
+			return false;
+		}
+
+		if (databaseToUpgrade_->insertOrUpdate(0, "AMDbObjectTypes_loadColumns", names, types)) {
+			databaseToUpgrade_->rollbackTransaction();
+			AMErrorMon::alert(this, BIOXASDBUPGRADE1PT1_COULD_NOT_INSERT_OR_UPDATE_TABLE, "Could not add BioXASXASScanConfiguration entry to AMDbObjectType_loadColumns.");
```
dretrex: Missing a `return false`.

dretrex:
As a first impression, this looks really good!  Great work, @helfrij!  Now we just need to do some thorough testing to make sure there isn't anything weird and we should be good to go!

helfrij:
Hey @dretrex, is this what you meant by using the AMDbObjectSupport::addTable method, instead of ensureTable?

dretrex:
Yup, that's what I meant.

helfrij:
@dretrex 
At the end of the day yesterday, some scans were loaded correctly and some weren't. Today I noticed that the ones that weren't loading were trying to find data in the wrong place:

Serious: "AMCDFDataStore: Could not open file '/Users/helfrij/SideMachine/AcquamanSideData/users/99-9999/userData//2015/07/Mon_07_13_13_53_16_8208.cdf' for the CDF data store. It's possible that the file does not exist or is corrupted. The CDF data store is now invalid. (code -101)" 
Serious: "Could not open the CDF file at '/Users/helfrij/SideMachine/AcquamanSideData/users/99-9999/userData//2015/07/Mon_07_13_13_53_16_8208.cdf'. It's possible that the file has been moved, or is corrupted. (code 627801)" 
Debug: "in [AMXASScan]: AMScan: Loading data failed. (code -483)" 
Debug: "[AMDbObjectSupport] Could not load the object with ID 76 from the table 'AMScan_table': loadFromDb() failed. Please report this bug to the Acquaman developers (code -92)" 

I added a check in AMCDFv1FileLoaderPlugin that makes sure the userData filename doesn't have a '/' character at the end of it, and it seemed to fix the problem with loading data. Is there any reason to worry that this change will mess up file loading for anyone else?

helfrij:
I believe all issues with the db upgrade that have been discovered have been addressed. The Side application can access previous data (open scans in a scan editor, and show configurations) that was collected on both Side and Main applications. Same with Main.

helfrij:
Starting Acquaman from scratch with a new database seems to work as expected. Also able to successfully load and open scans from a preexisting database.

helfrij:
Every time I open I scan to be edited, even though the scan is loaded successfully, I get an error:

Debug: "in [AMSamplePre2013]: Could not load from database, the database id is invalid. Please report this problem to the Acquaman developers. (code -277004)"

I notice that this error appears on versions of the BioXAS application that do not include this database upgrade. Does this error appear on other beamline applications? Or would this be uniquely BioXAS?

@acquaman/acquaman-working-group

davidChevrier:
Hi @helfrij 
I bet it's fairly common. Could you breakpoint and send me a stack trace to make sure I'm thinking about this correctly.

iainworkman:
It appears on SGM too, at least with the test data which I use.

PR #1542, opened 2015-08-19
Pull Request for Issue1540: yPoints inside of SXRMB2DScanActionController doesn't use the same axis for calculation

dretrex:
I not only fixed the typo, but went through the whole framework and made use of the  AMScanAxis::numberOfPoints() method in all the places that made sense to make the code cleaner.


iainworkman:
Looks good. Way clearer than what was there before.

dretrex:
Cool.  Thanks @iainworkman.  Time estimate?

iainworkman:
Time Estimate: 30 Minutes.

PR #1558, opened 2015-08-27
Pull Request for Issue1547: Upgrade IDEAS to AMChooseDataFolderDialog and local data caching

dretrex:
IDEAS now uses the modern data file selection dialog rather than the old school one.


iainworkman:
Looks good.
Time Estimate: 5 minutes

PR #1580, opened 2015-09-01
Pull Request for Issue1571: Change axisValues() from AMNumber * to double *.

dretrex:
Changed the optimized getter for the axis values to be a `double *` to match the rest of the bulk data requests.  I also optimized some of the code inside the methods as well to take advantage of the fact that they were now doubles and not AMNumbers.


source/dataman/datasource/AMDataSourceSeriesData.cpp
```diff
@@ -156,7 +156,7 @@ void  AMDataSourceSeriesData::onDataChanged(const AMnDIndex &start, const AMnDIn
 
 void AMDataSourceSeriesData::onAxisValuesChanged()
 {
-	QVector<AMNumber> axisData = QVector<AMNumber>(axisSize_, 0);
+    QVector<double> axisData = QVector<double>(axisSize_, 0);
 	source_->axisValues(0, 0, axisSize_-1, axisData.data());
```
iainworkman: Wait! You can get access to the underlying data in array form like this? I went through the rigmarole of creating a whole template class to safely handle arrays. :(
I'll certainly use this in future.

iainworkman:
Looks good. 
Time Estimate: 15 Minutes

PR #1582, opened 2015-09-02
Pull Request for Issue1574: Implement caching in commonly used 3D data sources and analysis blocks

dretrex:
Implemented caching and it appears to work as intended.

### DIFF
--- a/source/analysis/AM3DAdditionAB.cpp
+++ b/source/analysis/AM3DAdditionAB.cpp
@@ -24,6 +24,7 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 AM3DAdditionAB::AM3DAdditionAB(const QString &outputName, QObject *parent)
 	: AMStandardAnalysisBlock(outputName, parent)
 {
+    cacheUpdateRequired_ = false;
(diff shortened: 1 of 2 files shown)


source/analysis/AM3DAdditionAB.cpp
```diff
@@ -252,5 +242,29 @@ void AM3DAdditionAB::reviewState()
...
+{
+    AMnDIndex start = AMnDIndex(0, 0, 0);
+    AMnDIndex end = size();
+    int totalSize = start.totalPointsTo(end);
+
+    QVector<double> data = QVector<double>(totalSize);
+    sources_.at(0)->values(start, end, data.data());
+
+    // Do the first data source separately to initialize the values.
+    cachedData_ = data;
```
iainworkman: I assume it is guaranteed that there will always be at least one data source?

iainworkman:
Assuming that guarantee of at least one data source is true, then this looks good.
Time Estimate: 10 minutes

source/analysis/AM3DAdditionAB.cpp
dretrex: It is guaranteed since it is only called in `value()` and `values()` and it has to pass the `isValid()` check which does check against empty lists.

PR #1583, opened 2015-09-02
Pull Request for Issue1464: Arbitrary motions of the hexapod

iainworkman:
Added:
- AM3DCoordinatedSystemControl ~ An abstract class which allows for coordination of three motors in some 3d space.
- AM3DRotatedSystemControl ~ Extends coordinated system to allow rotations of the coordinate system
- SGMHexapodTransformedAxis ~ Extends AM3DRotatedSystemControl to interact with the trajectory moves of the hexapod as if they were simple single axes motions.

Tweaked:
- Made changes at both the AM level of the hexapod and the driver level in order to make fine motions more accurate.

General AM Changes:
- Added a new match type (MatchWithinTolerance) to the AMControlWaitAction, and added the ability to (optionally) specify the match type when calling AMAcitonSupport::buildControlWaitAction().


iainworkman:
Created it accidentally to merge into master, and there seems to be no way to change to merge to branch after the fact. Will recreate with the correct branch.

Issue #638, opened 2014-06-09
REIXS Would like to be able to adjust XES integration time during a scan


epengr:
Need to add functionality to bottom bar similar to loop number changes, and change REIXSXESScanActionController to handle. Maybe more?

epengr:
Time Estimate: 4 hours

PR #1585, opened 2015-09-02
Pull Request for Issue638 - adjust XES integration time during a scan

epengr:
Created a REIXS specific bottom bar subclass that is XES scan aware and is able to add to the dwell time for a currently running MCPDetector acquisition.

I hope that this is a good way to approach this, and not too crude of a hack.

#638


dretrex:
Looks good.
Time Estimate: 5 minutes

PR #1588, opened 2015-09-03
Pull Request for Issue1573: Implement caching in commonly used 2D data sources and analysis blocks

dretrex:
Implemented caching on the commonly used analysis blocks as well as the MPlot-AM data handlers.  

@iainworkman, you will notice that AM3DAdditionAB has some changes in it here as well.  It isn't that we missed anything in the original pull request, but certain behaviours only became apparent when updating the data on a 1 point at a time basis, which is untestable for data sources of rank 3.

### DIFF
--- a/source/analysis/AM1DCalibrationAB.cpp
+++ b/source/analysis/AM1DCalibrationAB.cpp
@@ -18,7 +18,6 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 
-#include <QDebug>
 #include <QtCore/qmath.h>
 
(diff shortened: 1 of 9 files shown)


iainworkman:
Looks good to me. I was wondering if there would be a way to bring more of the commonalities into a single 2D AB kind of class, but it seems that they are so subtly different that this would be of much benefit.
Time Estimate: 30 minutes.

dretrex:
No, there are actually A LOT of things that could be merged together.  Unfortunately, I think there is a non-trivial database upgrade that gets in the way. :(

PR #1593, opened 2015-09-04
Pull Request for Issue1572: Implement caching in commonly used 1D data sources and analysis blocks

dretrex:
Implemented the caching on the 1D sources.  These were a lot easier because flat arrays are really straight-forward for 1D data sets.

### DIFF
--- a/source/analysis/AM1DCalibrationAB.cpp
+++ b/source/analysis/AM1DCalibrationAB.cpp
@@ -405,7 +405,7 @@ bool AM1DCalibrationAB::values(const AMnDIndex &indexStart, const AMnDIndex &ind
         computeCachedValues();
 
     int totalSize = indexStart.totalPointsTo(indexEnd);
-    memcpy(outputValues, cachedData_.constData()+indexStart.i()*size(1), totalSize*sizeof(double));
+    memcpy(outputValues, cachedData_.constData()+indexStart.i(), totalSize*sizeof(double));
 
 	return true;
 }
--- a/source/analysis/AM1DCalibrationAB.cpp
+++ b/source/analysis/AM1DCalibrationAB.cpp
@@ -353,14 +353,14 @@ void AM1DCalibrationAB::computeCachedValues() const
     cacheUpdateRequired_ = false;
 }
 
-bool AM1DCalibrationAB::canAnalyze(const QString &dataName, const QString &NormalizationName) const
+bool AM1DCalibrationAB::canAnalyze(const QString &dataName, const QString &normalizationName) const
 {
 	// Can always analyze two 1D data sources.
 	if (sources_.count() == 2)
 		return true;
 
-	// The first data source is reserved for the data.
-	if (indexOfInputSource(dataName) >= 0 && indexOfInputSource(NormalizationName))
+	// Both data and normalization names must correspond to valid data sources.
+	if (indexOfInputSource(dataName) >= 0 && indexOfInputSource(normalizationName))
 		return true;
 
 	return false;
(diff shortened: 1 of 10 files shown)


source/analysis/AM1DCalibrationAB.cpp
```diff
@@ -208,113 +219,171 @@ void AM1DCalibrationAB::setInputSources()
...
+
+            for (int i = 0; i < totalSize; i++)
+                cachedData_[i] /= scale;
+        }
+    }
+
+    cacheUpdateRequired_ = false;
 }
 
+bool AM1DCalibrationAB::canAnalyze(const QString &dataName, const QString &NormalizationName) const
```
iainworkman: camel case in ```NormalizationName```.

source/analysis/AM1DCalibrationAB.cpp
```diff
@@ -208,113 +219,171 @@ void AM1DCalibrationAB::setInputSources()
...
+{
+	// Can always analyze two 1D data sources.
+	if (sources_.count() == 2)
+		return true;
 
+	// The first data source is reserved for the data.
+	if (indexOfInputSource(dataName) >= 0 && indexOfInputSource(NormalizationName))
+		return true;
```
iainworkman: Bit confused by this here, given the comment.
I'm assuming that the comment is saying that index 0 must be the raw data (ie the one with the key ```dataName```), is this correct? If so why is this 
```C++
if(indexOfInputSource(dataName) >= 0 ...
```
and not
```C++
if(indexOfInputSource(dataName) == 0 ...
```

Is there some case I haven't considered?

dretrex: Unfortunately, that is a case of unclear commenting.  The logic is correct with the ` >= 0` because the _data_ source can be any data source that has been provided (i.e.: `0` to `size()-1`).

source/analysis/AM1DDerivativeAB.cpp
```diff
@@ -153,10 +161,73 @@ void AM1DDerivativeAB::setInputSource()
...
+
+        if (axis.at(totalSize-1) == axis.at(totalSize-2))
+            badIndices.append(totalSize-1);
+
+        // Compute all the values
+        cachedData_[0] = (data.at(1)-data.at(0))/(axis.at(1)-axis.at(0));
+        cachedData_[totalSize-1] = (data.at(totalSize-1)-data.at(totalSize-2))/(axis.at(totalSize-1)-axis.at(totalSize-2));
+
+        for (int i = 1, count = totalSize-1; i < count; i++)
+            cachedData_[i] = (data.at(i+1)-data.at(i-1))/(2*(axis.at(i+1)-axis.at(i-1)));
```
iainworkman: My calculus is pretty rusty, but isn't this making a note of the bad indices, and then ignoring them when it actually comes to do the calculation for a given ```i```?

I feel like this should be:
```C++
for (int i = 1, count = totalSize-1; i < count; i++) {
    if(!badIndices.contains(i))
        cachedData_[i] = (data.at(i+1)-data.at(i-1))/(2*(axis.at(i+1)-axis.at(i-1)));
}
```
Then the following loop would go through and fill in all the values which were missed because of bad indices. That said I would imagine the ```contains()``` method is probably O(numberOfBadIndices), and the chances of two values being a precise double match is slim - in which case maybe the badIndices should just be manually set as 0 and size -1, and then manually added?

dretrex: Yeah, since the problem with dividing by zero and the like only causes issues when trying to use it (since it'll just be _nan_), I elected to go this way because then we have a O(n) + O(m < n) rather than O(n*m).

iainworkman:
Other than the couple of comments, this looks good.

PR #1594, opened 2015-09-08
Pull Request for Issue1424: Visualizing the Hexapod

iainworkman:
Added a main window pane for the hexapod and a simply widget for rotating the hexapod's coordinate system. At the moment it is not super clear what the end goal is for the hexapod user interface, so this is going to be a first rev, which will be iterated on with feedback from Tom, Teak and users.

Estimate: 2


dretrex:
Looks good.
Time Estimate: 10 minutes

PR #1599, opened 2015-09-10
Pull Request for Issue1598 - REXIS XES AB - apply bin with other AB parameter

epengr:
Added the new binning level to the apply-to-other-scans feature, which I neglected to do...


dretrex:
Time Estimate: 1 minute

PR #1600, opened 2015-09-11
Pull Request for Issue1575: Implement caching in all commonly used data sources and analysis blocks that change rank

dretrex:
Implemented  the rank reduction data sources that are commonly used.  Validated by testing, I figured out a more efficient manner in which to calculate the data.  Based on the tests I did there appears to be no adverse effects due to the new algorithm.


source/analysis/AM2DSummingAB.cpp
```diff
@@ -315,40 +317,92 @@ bool AM2DSummingAB::axisValues(int axisNumber, int startIndex, int endIndex, dou
...
-	if (startIndex >= inputSource_->axisInfoAt(otherAxis).size || endIndex >= inputSource_->axisInfoAt(otherAxis).size)
+	if (startIndex >= inputSource_->size(otherAxis) || endIndex >= inputSource_->size(otherAxis))
 		return false;
 
 	return inputSource_->axisValues(otherAxis, startIndex, endIndex, outputValues);
 }
 
+void AM2DSummingAB::setSumAxis(int sumAxis)
+{
+	if((unsigned)sumAxis >= 2)
```
iainworkman: I believe this introduces a rare corner case if sumAxis is passed as INT_MIN or INT_MIN + 1.
If we were to take a 32 bit int, with INT_MIN = -4294967296. The cast would add 2^32 in order to produce a valid unsigned value, and produce 0 for INT_MIN or 1 for INT_MIN + 1, which would then fail this if block and drop through to the rest of the code.

Obviously the chances of this are very minimal, but it would probably be more straight forward to just go with:
```C++
if(sumAxis < 0 || sumAxis >=2)
    return;
```

source/analysis/AM3DBinningAB.cpp
```diff
@@ -430,10 +383,85 @@ bool AM3DBinningAB::axisValues(int axisNumber, int startIndex, int endIndex, dou
...
+	if (startIndex >= inputSource_->size(actualAxis) || endIndex >= inputSource_->size(actualAxis))
 		return false;
 
-	return inputSource_->axisValues(actualAxis, startIndex, endIndex, outputValues);
+    return inputSource_->axisValues(actualAxis, startIndex, endIndex, outputValues);
+}
+
+void AM3DBinningAB::setSumAxis(int sumAxis)
+{
+    if((unsigned)sumAxis >= 2)
```
iainworkman: Same as above re: cast to unsigned.

source/analysis/AM4DBinningAB.h
```diff
@@ -24,8 +24,6 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "analysis/AMStandardAnalysisBlock.h"
 
-#define AM3DMAGICNUMBER -1.493920384020
```
iainworkman: That is the best constant name I've ever seen.

iainworkman:
Everything looks good apart from the comment regarding use of unsigned. I'd like to pop over this morning and just go through the algorithm for computing the values, just because it is extremely elegant but also difficult to pick through coming from cold (so it might benefit from a few more comments regarding its workings).

source/analysis/AM2DSummingAB.cpp
dretrex: Good catch.  That is a legacy bug as I mostly just moved these functions from the header to the cpp file.

dretrex:
I'll be in IDEAS for the majority of the day.  Come find me any time.

davidChevrier:
Time Estimate: 1.5 Hours

Issue #1520, opened 2015-08-07
AMScan View in GSE not sizing properly


epengr:
The new data tools are taking up 50% of the screen real estate.

![screenshot](https://cloud.githubusercontent.com/assets/677939/9145431/8f9acd82-3d11-11e5-938f-f819ca7ed974.png)

epengr:
I have a theory about this one - REIXS is the open deployment of AM that running on a 2560*1440 desktop...

epengr:
I stole this issue, and tracked it down to an addStretch() in the scanView vertical layout (there wasn't one in the 2DScanView).
Completed in #1607
Time Estimate: 2 hours

PR #1607, opened 2015-09-16
Pull Request for Issue1520: Removed stretch from GSE layout

epengr:
Stretch in the scanView in the vertical layout was keeping from using available screen space on higher resolution monitors.  

#1520


dretrex:
Time Estimate: 1 minute

epengr:
Oh, @dretrex, I couldn't find the source of that little bit of dead space at the top of the right side bar that you pointed out...  Not worth spending hours on , I think.

dretrex:
Agreed.  Thanks for looking though.

PR #1608, opened 2015-09-16
Pull Request for Issue1605: Initial XES scan axis values incorrect

epengr:
Moved the initialization of the binning values before the axis values are set invalid in the constructor so that the first computeCachedAxisValues call uses the correct bin level.

### DIFF
--- a/source/analysis/REIXS/REIXSXESImageInterpolationAB.cpp
+++ b/source/analysis/REIXS/REIXSXESImageInterpolationAB.cpp
@@ -47,6 +47,8 @@ REIXSXESImageInterpolationAB::REIXSXESImageInterpolationAB(const QString &output
 	energyCalibrationOffset_ = 0;
 	tiltCalibrationOffset_ = 0;
 
+	interpolationLevel_ = 10;
+	binningLevel_ = 2;
 
 	inputSource_ = 0;
 	cacheInvalid_ = true;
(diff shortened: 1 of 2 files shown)


epengr:
@dretrex, does this make sense?  I'm not 100% clear on what's triggering the first axis value computation, but it seems that if the binning level is initialized before the axisValuesCacheInvalid is set true, then is should be computed correctly whenever it is triggered.

I can't test this right now as there are users on the endstation.

dretrex:
No, that should have no effect on your issue.  It will be related to the fact that something isn't looking at your binning level on the first pass, but is captured correctly after you change the value of the spin box.

source/analysis/REIXS/REIXSXESImageInterpolationAB.cpp
```diff
@@ -303,6 +303,7 @@ void REIXSXESImageInterpolationAB::setInputDataSourcesImplementation(const QList
 	cachedValues_ = QVector<double>(axes_.at(0).size);
 	axisValueCacheInvalid_ = true;
 	cachedAxisValues_ = QVector<double>(axes_.at(0).size);
+	onInputSourceSizeChanged();
```
epengr: @dretrex, this fixes it, but I don't think it's the right way to do it. In fact, i don't think that it should be needed, since when a source is connected ```inputSource_->signalSource()/sizeChanged(int)``` is connected to ```onInputSourceSizeChanged()``` (line 297 above) and on line 309 below the ```emitSizeChanged(0)``` is called which should trigger this anyways, if things are all connected up properly, shouldn't it?

epengr: @dretrex ,
The connection on line 297 seems to be bad.  I have found in the past that the argument mis-match can cause a silent connection failure.  However, when I tried to make ```onInputSourceSizeChanged()``` into ```onInputSourceSizeChanged(int)``` so that they do match, it still doesn't work.  ```emitSizeChanged(0)``` is calling the ```AMDatasource::emitSizeChanged(int)```, which is calling ```AMDataSignalSource::emitSizeChanged(int)```, but the chain stops there.  It never reaches ```onInputSourceSizeChanged()``` or ```onInputSourceSizeChanged(int)```   (and ```onInputSourceSizeChanged()``` never paid attention to the argument, and only looks at axis 0.

dretrex: That may be part of the problem, but in general, a slot having less arguments than the signal is okay provided that if there are parameters that they follow the same order (from first to last).  I would have been very surprised if you changing the slot to having an int would have made a difference.

epengr:
@dretrex,  I can't get the sizeChanged signal to connect no matter what I do.  The stateChange connection made at the same time also doesn't workhowever valuesChanged does.  

Can we merge in this change (just calling onInputSourceSizeChanged() directly on new data source) so I can deploy a new version for a group using the endstation next week?

The Issue could be left open and I'll continue to work on it as I have time (less now that the endstation has arrived...)

epengr:
Time Estimate: 4 hours

dretrex:
I would like to look into it before we merge it ourselves.  Would you be able to sit down with me on Monday afternoon sometime?  I'd really like to have a look at it because this behaviour seems a little off to me.

epengr:
Yep, Monday works.   

Thanks.

dretrex:
Time Estimate: 40 minutes

PR #1620, opened 2015-09-21
Pull Request for Issue1560: Create support classes for energy control calculations

iainworkman:
- Added an AMValidator class which provides a simple structure for managing validity checks for an instance of an object.
- Added support namespaces for mono energy calculations related to the undulator (SGMUndulatorSupport) and grating (SGMGratingSupport).
- Added SGMMonochromatorInfo - A class which stores the state of a monochromator and assists in performing the optimization calculations, targeting energies etc.
- Added very simple test suite which includes a basic test to ensure that the mono info class is coming up with the right values, as well as a test view class which ensures that the relationships between the components is valid.


source/beamline/SGM/monochromator/SGMGratingSupport.h
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	/*!
+	  * Correction value used in the calculation of the energy produced by a given
+	  * grating translation.
+	  * \param translationSelection ~ The grating whose correction value is to be
+	  * returned.
+	  */
+	inline static double c1(GratingTranslation translationSelection)
```
dretrex: Are there more verbose names you could give these guys?  I know they were probably called c1 and c2, but if they have a contextual name, I think that would benefit everyone (including our future selves when coming back to this code in the future).

source/beamline/SGM/monochromator/SGMGratingSupport.h
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	/*!
+	  * Correction value used in the calculation of the energy produced by a given
+	  * grating translation.
+	  * \param translationSelection ~ The grating whose correction value is to be
+	  * returned.
+	  */
+	inline static double thetaM(GratingTranslation translationSelection)
```
dretrex: Things like this might be better suited to `monochromatorAngle` now that we don't have to worry about weird names and the like.

source/beamline/SGM/monochromator/SGMGratingSupport.h
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	/*!
+	  * Correction value based on the radius of the grating for each of the grating
+	  * translations, used in the calculation of the energy produced.
+	  * \param translationSelection ~ The translation whose radius curvature correction
+	  * is to be returned.
+	  */
+	inline static double radiusCurvatureOffset(GratingTranslation translationSelection)
```
dretrex: Like this, this is really nice.

source/beamline/SGM/monochromator/SGMGratingSupport.h
```diff
@@ -0,0 +1,140 @@
...
+    /*!
+     * Checks whether the given energy is valid for the provided grating translation.
+     * \param translationSelection ~ The grating translation to check against.
+     * \param energy ~ The energy value to check the validity of for the given grating translation.
+     */
+    inline static bool validEnergy(GratingTranslation translationSelection, double energy)
+    {
+        switch(translationSelection) {
+        case LowGrating:
+            return (energy > 9);
```
dretrex: These checks seem overly simplistic.  Are we sure these greater than checks are sufficient?

source/beamline/SGM/monochromator/SGMGratingSupport.h
iainworkman: Yes! I have talked about this with Tom and he agrees. I need to get hold of
him for long enough for us to figure these out. In some cases, however,
these are just mathematical constructs (eg c1 and c2 are just correction
values).

source/beamline/SGM/monochromator/SGMMonochromatorInfo.cpp
```diff
@@ -0,0 +1,382 @@
...
+    return (-1/0.14295709668) * log( (1/36.00511212946)*((1737.41045746644/(energy/int(undulatorHarmonic))) -1)) + undulatorOffset;
+}
+
+double SGMMonochromatorInfo::optimizedExitSlitPosition(SGMGratingSupport::GratingTranslation gratingTranslationSelection, double energy) const
+{
+	double spacing = SGMGratingSupport::spacing(gratingTranslationSelection);
+	double thetaM = SGMGratingSupport::thetaM(gratingTranslationSelection);
+	double xOffset = 4546; // Need to research this value. Seems to have been altered over time.
+	double lambda = (1239.842 / energy) * 1.0e-9;
+	double thetaI = asin((lambda / (2*spacing)) / cos(thetaM/2)) + (thetaM/2);
```
dretrex: Can these I's, M's, and D's be given explicit names?

source/beamline/SGM/monochromator/SGMMonochromatorInfo.h
```diff
@@ -0,0 +1,384 @@
...
+
+protected:
+	/*!
+	  * Helper method which returns the best fit grating translation for a given
+	  * energy and optimization mode.
+	  * \param requestedEnergy ~ The energy to optimize the grating translation for.
+	  * \param optimizationMode ~ The optimization criteria for selecting a grating
+	  * translation.
+	  */
+	SGMGratingSupport::GratingTranslation optimizedGrating(double requestedEnergy,
```
dretrex: Would this not be a `const` function?

source/beamline/SGM/monochromator/SGMUndulatorSupport.h
```diff
@@ -0,0 +1,84 @@
...
+ * \param undulatorHarmonic ~ The undulator to check against.
+ * \param energy ~ The energy value to check the validity of for the given harmonic.
+ */
+inline static bool validEnergy(UndulatorHarmonic undulatorHarmonic, double energy)
+{
+    switch(undulatorHarmonic) {
+    case FirstHarmonic:
+        return (energy > 47 && energy < 1738);
+    case ThirdHarmonic:
+        return (energy > 141 && energy < 5213);
```
dretrex: Is this right?  SGM doesn't go higher than 2000 eV, this check seems VERY high.

source/tests/SGM/SGMMonochromatorInfoTestView.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+        errorsTextEdit_->setText(errorString);
+    }
+}
+
+void SGMMonochromatorInfoTestView::onGratingAngleSpinBoxChanged()
+{
+    double value = gratingAngleSpinBox_->value();
+    if(qAbs(value - sgmMonochromatorInfo_->gratingAngle()) > 0.001 ||
+            sgmMonochromatorInfo_->hasErrors()) {
```
dretrex: Is this logically correct?  You want to set a new value to the info if the value is different (seems fine) _OR_ there are errors?

source/tests/SGM/SGMMonochromatorInfoTestView.cpp
```diff
@@ -0,0 +1,428 @@
...
+    if(sgmMonochromatorInfo_->isUndulatorTracking() != value) {
+        sgmMonochromatorInfo_->setUndulatorTracking(value);
+    }
+}
+
+void SGMMonochromatorInfoTestView::onUndulatorPositionSpinBoxChanged()
+{
+    double value = undulatorPositionSpinBox_->value();
+    if(qAbs(sgmMonochromatorInfo_->undulatorPosition() - value) > 0.001 ||
+            sgmMonochromatorInfo_->hasErrors()) {
```
dretrex: Same deal with these guys.  If this is on purpose and intended then you can disregard.

source/tests/SGM/SGMMonochromatorInfoTestView.h
```diff
@@ -0,0 +1,75 @@
...
+    Q_OBJECT
+public:
+    explicit SGMMonochromatorInfoTestView(SGMMonochromatorInfo* mono, QWidget *parent = 0);
+
+signals:
+
+public slots:
+
+protected slots:
+    void onMonoGratingAngleChanged(double gratingAngleSteps);
```
dretrex: A couple comments for these guys would be a nice addition.  Nothing crazy since they are fairly straight-forward.

source/beamline/SGM/monochromator/SGMGratingSupport.h
iainworkman: I imagine these will evolve as time goes on, and I can pin down Tom on what constraints he would like (which, again, I imagine will be an evolving process).

This (and the other function for valid energy on the undulator below) are for the moment based upon those value which, if entered into the formula for for position based on energy, will result in NaN.

source/tests/SGM/TestSGMMonochromatorInfo.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    // Calcium L Edge:
+    testIndividualRequestEnergy("Calcium L Edge Start", 345.0, SGMGratingSupport::MediumGrating, -590024.88, 1834.184986, 15.308185, -123722.20);
+    testIndividualRequestEnergy("Calcium L Edge Middle", 355.0, SGMGratingSupport::MediumGrating, -573404.46, 1642.027233, 15.55847741, -120766.24);
+    testIndividualRequestEnergy("Calcium L Edge End", 365.0, SGMGratingSupport::MediumGrating, -557694.75, 1473.127714, 15.80358308, -117871.54);
+
+    // Titanium L Edge:
+    testIndividualRequestEnergy("Titanium L Edge Start", 450.0, SGMGratingSupport::LowGrating, -247476.56, 34.56543835, 17.71524808, -95294.76);
+    testIndividualRequestEnergy("Titanium L Edge Middle", 465.0, SGMGratingSupport::LowGrating, -239493.45, 53.21604579, 18.02659692, -91617.73);
+    testIndividualRequestEnergy("Titanium L Edge Start", 480.0, SGMGratingSupport::LowGrating, -232009.28, 75.02669983, 18.33163518, -88015.23);
```
dretrex: Titanium L Edge End?

source/beamline/SGM/monochromator/SGMUndulatorSupport.h
iainworkman: As the line note above states, for the moment these checks are simply based on value which result in NaN.

dretrex:
A general comment is that it appears you have different tab settings on the different machines you're working on.  Nothing deal breaking, but you may want to look into making sure that your projects use the same spacing.

source/beamline/SGM/monochromator/SGMGratingSupport.h
dretrex: I agree that they may just be correction values, but what are they correcting?  Are they angle corrections?  Translation corrections?  Even something minor like that would be helpful.

dretrex:
Time Estimate: 25 minutes

iainworkman:
With regard to spacing, I've noticed that a lot of the ways I format code (breaking function arguments onto multiple lines, for instance) get mangled when going through git, which can make pull requests a nightmare. 
I'm going to look at whether this is down to Qt indenting options and see if I can pin down something that works.

iainworkman:
A follow up to that:
The mangling of code from Qt to git on my machine was due to a setting on my OPI in Qt Creator
Align Continuation Lines: With Spaces
when it should have been
Align Continuation Lines: With Regular Indent

dretrex:
Ah, yeah.  That'll do it.

iainworkman:
Okay, I've fixed up most of the smaller things. There are two things remaining:
- Working out error/warning states which makes sense for SGM
- Sitting down with Tom and figuring out what all the c1, c2, thetaM etc variable names relate to.

For the second issue I've fired off an email to Tom, so we can pin down a time.

iainworkman:
Time Estimate: 2 Hours

dretrex:
Should me make those other two things issues and merge this in or would you prefer to wait to make these changes?

iainworkman:
I'd say wrap this up, and certainly do the error/warning states in another issue. The variable names will get cleaned up within other issues, I'd imagine, pending a response from Tom.

dretrex:
Done.  Would you mind creating the other issues?

iainworkman:
Will do.

PR #1624, opened 2015-09-22
Pull request for Issue1623: Rename variables used in SGM energy calculation

iainworkman:
- Merged c1 and c2 into curveFitCorrection variable (=c1*c2) as they were only ever used together.
- Renamed the variables based on what they represent from information gathered from Tom and the SGM Wiki.


dretrex:
I like it.
Time Estimate: 2 minutes

PR #1629, opened 2015-09-28
Pull Request for Issue1562: Continuous Motion Support

iainworkman:
Added classes to represent an energy trajectory which calculates an approximated coordinated movement of all the components of the monochromator (currently the approximation is linear).

(Also includes test classes which visualize the trajectory and coordination).


source/application/SGMTests/SGMTestsWindow.cpp
```diff
@@ -40,7 +43,52 @@ void SGMTestsWindow::onShowMonoViewClicked()
...
 
-    SGMMonochromatorInfoTestView* testView = new SGMMonochromatorInfoTestView(testMonoInfo);
-    testView->show();
+    monoInfoTestView_ = new SGMMonochromatorInfoTestView(testMonoInfo);
+    monoInfoTestView_->show();
+}
+
+void SGMTestsWindow::onShowTrajectoryViewClicked()
+{
+    if(energyTestView_ != 0) {
```
dretrex: I know this is just in your tests, but would it make more sense to do the inverse of this?  If you don't have one, then make one, but every subsequent call you just bring up the one that has been built already?

source/beamline/SGM/monochromator/SGMEnergyTrajectory.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    startMonoInfo_ = new SGMMonochromatorInfo(startEnergy, gratingTranslation);
+    startEnergy_ = startEnergy;
+    endMonoInfo_ = new SGMMonochromatorInfo(endEnergy, gratingTranslation);
+    endEnergy_ = endEnergy;
+
+    // ensure both undulator harmonics are the same.
+    if(startMonoInfo_->undulatorHarmonic() != endMonoInfo_->undulatorHarmonic()) {
+
+        // If there is a conflict, just use the first harmonic.
```
dretrex: Design choice?

source/beamline/SGM/monochromator/SGMEnergyTrajectory.cpp
```diff
@@ -0,0 +1,199 @@
...
+        startMonoInfo_->setAutoDetectUndulatorHarmonic(false);
+        endMonoInfo_->setAutoDetectUndulatorHarmonic(false);
+
+        startMonoInfo_->setUndulatorHarmonic(SGMUndulatorSupport::FirstHarmonic);
+        endMonoInfo_->setUndulatorHarmonic(SGMUndulatorSupport::ThirdHarmonic);
```
dretrex: If we are just using the first harmonic like your previous comment mentions, shouldn't this line also be `SGMUndulatorSupport::FirstHarmonic`?

source/beamline/SGM/monochromator/SGMEnergyTrajectory.cpp
```diff
@@ -0,0 +1,199 @@
+#include "SGMEnergyTrajectory.h"
+
+SGMEnergyTrajectory::SGMEnergyTrajectory(double startEnergy,
+                                         double endEnergy,
+                                         double time,
+                                         SGMGratingSupport::GratingTranslation gratingTranslation)
+{
+
+    startMonoInfo_ = new SGMMonochromatorInfo(startEnergy, gratingTranslation);
```
dretrex: Is there a design reason for new'ing the mono info's rather than just using them as stack variables?

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+        int numberOfItems = int((endValue - startValue) / increment);
+        SGMMonochromatorInfo monoInfo(200, gratingTranslation);
+
+        QVector<qreal> xValues = QVector<qreal>(numberOfItems);
+        QVector<qreal> gratingAngleYValues = QVector<qreal>(numberOfItems);
+        QVector<qreal> undulatorPositionYValues = QVector<qreal>(numberOfItems);
+        QVector<qreal> exitSlitPositionYValues = QVector<qreal>(numberOfItems);
+
+        xValues.clear();
```
dretrex: These `clear()` methods are unnecessary.  They are empty by virtue of the constructor you used.

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+#include "beamline/SGM/monochromator/SGMMonochromatorInfo.h"
+#include "beamline/SGM/monochromator/SGMEnergyTrajectory.h"
+
+#include <QHBoxLayout>
+#include <QVBoxLayout>
+#include <QGridLayout>
+#include <QLabel>
+
+#define GRATING_ANGLE_PLOT_INDEX 0
```
dretrex: These defines don't appear to be used anywhere.

source/util/AMTrapezoidVelocityProfile.h
```diff
@@ -0,0 +1,82 @@
...
+/*!
+ * \brief A class which represents the velocity profile of a some motion which
+ * follows a trapezoidal structure ie. it begins at some base velocity, accelerating
+ * at the initialAcceleration towards some target velocity; when it reached the
+ * point at which it must decelerate to return to the base velocity by the end
+ * of the motion changing velocity at the rate defined by returnAcceleration it
+ * will begin its deceleration.
+ * NOTE 1: If the time of the motion is not sufficient to accelerate up to the
+ * target velocity, the profile will resemble a triangle, rather than a trapezoid.
+ * NOTE 2: The velocity profile might describe a motion in which the target velocity
```
dretrex: Is this realistic?  This seems a bit weird to me.

dretrex:
Overall, I think this looks good.  I didn't see anything wrong with your implementation, @iainworkman.  Maybe you can mention what was bothering you about it again here just in case I missed what it was you were concerned with.

Time Estimate: 25 minutes

source/beamline/SGM/monochromator/SGMEnergyTrajectory.cpp
iainworkman: Thanks. Good catch.

source/beamline/SGM/monochromator/SGMEnergyTrajectory.cpp
iainworkman: I'm going to speak with Tom regarding this. It's a fairly arbitrary decision, but there is probably some more complex logic which could be used to determine which harmonic to use in cases of conflict.

source/beamline/SGM/monochromator/SGMEnergyTrajectory.cpp
iainworkman: There wasn't really a good state to initialize the mono info in for the case of the default constructor.  The calculations which follow (when setting the value of a given component, or requesting an energy for example) all tend to prompt a series of follow up calculations, and having the info initialized in an invalid state was causing a lot of corner cases in which these calculations failed (came out with NaN usually).

To get around this I made sure that there wasn't a default constructor for the info, and forced the setting of the values on initialization. This, however, means that I can't have one as a class member variable, without it being a pointer reference or using member initialization lists.

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
iainworkman: That's odd. When I put them into the plot, or stopped on the debugger they always seemed to be initialized with a single value of 0 when calling this constructor. Calling clear immediately after fixed the bug.
Will look into this more closely.

source/util/AMTrapezoidVelocityProfile.h
iainworkman: Where the target velocity is less than the base? Well I wanted to make it fairly general, but I could also imagine the case where you're at position 20 and you would like to move to position 10 in 5 seconds; in which case your base velocity will be 0 and your target -2.

iainworkman:
I dunno, just felt like I was getting off track a bit, specifically the AMTrapezoidalTrajectory. I wasn't sure whether it would fulfil all the requirements of it in the future.

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
iainworkman: Set this up in a test project, and 
```
QVector<double> testVector = QVector<double>(10);
```
creates a vector with ten elements, each initialized to zero.

source/beamline/SGM/monochromator/SGMEnergyTrajectory.cpp
dretrex: Sounds good.

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
dretrex: That is the behaviour I have come to expect.  Is that not what you were expecting?

source/util/AMTrapezoidVelocityProfile.h
dretrex: Oh, as in going backwards?  Yeah, that would make sense.  I was only thinking of going in one direction, but you are quite right.

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
iainworkman: I was expecting it to allocate the memory to make room for the elements, but not actually add them (thus preventing the worst case complexity of having to reallocate space when the underlying array has run out of space, and copy over values). I could rewrite this to explicitly overwrite elements at position i, instead of appending.

dretrex: Overwriting the elements is how I do 100% of my QVector use.  Appending is quite bad as each append is O(n) rather than doing it all at once then overwriting.

dretrex:
Was that the last of your changes @iainworkman?  If so, I'll merge this in.

iainworkman:
Yes, that's the last of my changes.

iainworkman:
Time Estimate: 3 Hours

PR #1633, opened 2015-09-29
Pull Request for Issue1632: Add command line argument to exclude the 13-element detector

epengr:
Added command line argument for 13-element detector
"--Ge13" is required in the command line or the 13-element detector is
excluded through the application.

Tested and found expected behavior without --Ge13.  AM complains the same as it did before with --Ge13.  I believe this is adequate testing for now, but it may need to be revisited at a future date when the 13-element detector electronics are online on IDEAS.

### DIFF
--- a/source/beamline/IDEAS/IDEASBeamline.cpp
+++ b/source/beamline/IDEAS/IDEASBeamline.cpp
@@ -98,13 +98,10 @@ void IDEASBeamline::setupDetectors()
 	ketek_ = new IDEASKETEKDetector("KETEK", "Single Element XRF Detector", this);
 	addSynchronizedXRFDetector(ketek_);
 
-	if (QApplication::instance()->arguments().contains("--Ge13"))
-	{
 	ge13Element_ = new IDEAS13ElementGeDetector("13-el Ge", "The thirteen element Germanium Detector", this);
 	addSynchronizedXRFDetector(ge13Element_);
 	ge13ElementRealTimeControl_ = new AMReadOnlyPVControl("13-el Ge Real Time", "dxp1608-B21-13:ElapsedReal", this);
 	ge13ElementRealTime_ = new AMBasicControlDetectorEmulator("13E_dwellTime", "13-element Ge dwell time", ge13ElementRealTimeControl_, 0, 0, 0, AMDetectorDefinitions::ImmediateRead, this);
-	}
 
 	ketekPeakingTime_ = new AMPVControl("XRF1E Peaking Time","dxp1608-1002:dxp1:PeakingTime_RBV","dxp1608-1002:dxp1:PeakingTime", QString(), this, AMCONTROL_TOLERANCE_DONT_CARE);
 	ketekTriggerLevel_ = new AMPVControl("XRF1E Trigger Level","dxp1608-1002:dxp1:TriggerThreshold_RBV","dxp1608-1002:dxp1:TriggerThreshold", QString(), this, AMCONTROL_TOLERANCE_DONT_CARE);
@@ -281,7 +278,7 @@ AMXRFDetector *IDEASBeamline::xrfDetector(IDEAS::FluorescenceDetectors detectorT
 	if (detectorType.testFlag(IDEAS::Ketek))
 		XRFDetector = IDEASBeamline::ideas()->ketek();
 
-	else if (detectorType.testFlag(IDEAS::Ge13Element) && QApplication::instance()->arguments().contains("--Ge13"))
+	else if (detectorType.testFlag(IDEAS::Ge13Element) && IDEASBeamline::ideas()->ge13Element()->isConnected())
 		XRFDetector = IDEASBeamline::ideas()->ge13Element();
 
 	return XRFDetector;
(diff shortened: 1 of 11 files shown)


dretrex:
So I'm not the biggest fan of this kind of implementation.  It is too hard coded and doesn't scale well.  I also don't like that a script has to be modified when the the detector changes.  I'm not super concerned with you doing, but the comfortableness of the person making changes like that is also not scalable.  We may need to invest in making a simple program launcher that permeates changes into the application without having hardcoded arguments like this.  If you would like to have this as a temporary solution I would be willing to merge it in and we make a new issue to handle this properly.  Let me know what you think.  @acquaman/acquaman-working-group, does anyone else have any thoughts on this?

dretrex:
Time Estimate: 10 minutes

helfrij:
What happens if the detector object is created but not connected? An alternate solution might be to handle setup/adding to scan configurations when the detector reports as connected, ignoring it otherwise. Does that make sense?

dretrex:
That is probably a more sensible solution, @helfrij.

epengr:
If the detector is not connected, errorMon spits out a bunch of complaints, so it's clearly being identified already.  I don't know that it's early enough to modify the contents of panels that have already spawned or not...

My thought for end use would be to create multiple scripts to go along with this, joe-end-user just has to click AM-with-Ge13 ot AM-without-Ge13 and their off...

This can either be merged as-is (I'm happy with it), or I can investigate @helfrij 's suggestion if you'd prefer.

epengr:
To clarify -  I believe that it requires a certain level of savvy to set the 13-Ge up, or move it from one beamline to another, and changing that script would just be part of that process.  Someone involved in that process has to be able to handle that.  It can even be added as a step in the CID IDEAS wiki for setting up the detector.

dretrex:
I know that functionally they would operate in a similar fashion, but I really don't like having users having to click on different scripts.  I would prefer something inside the app controller that would save something persistently and upon loading could take action.  

Also, in your current implementation, I would still change it to minimize the number of places you are referencing the command line arguments.  For instance, the _ONLY_ place I think that should be referenced is IDEASBeamline where you instantiate the detector.  Something like,

```c++
if (arg == "--Ge13")
    detector_ = new Detector();

else
    detector_ = 0;
```

Then, in all of your views and whatnot, you would simply check if the returned detector from the beamline was valid (i.e.: not 0).

epengr:
Hmmm, that does make a lot more sense.  I think I see how to make that work.

I don't envision this as a users choice "are you planning to us the Ge13, if so click here" type of thing.  I'm envisioning removing that argument when the detector is on VESPERS and adding it when the detector is set up on IDEAS.  So it would be me, or possible CID, making the change when the detector is physically moved.

dretrex:
I see what you're saying.  But it's not you who I'm concerned about.  It's everyone else.

epengr:
@dretrex, I took a crack at implementing @helfrij 's suggestion.  Is this what you had in mind?

I think it will work.  I testing is by swapping in the ketek's connected signal anywhere I used the ge13Element's signal and when I did, all the GUI elements for the ge13Element were created correctly.

source/ui/IDEAS/IDEAS2DScanConfigurationView.cpp
```diff
@@ -15,6 +15,7 @@
 #include <QSpinBox>
 #include <QMenu>
 #include <QStringBuilder>
+#include <QApplication>
```
dretrex: Probably don't need this one since you have one in the cpp file as well.

dretrex: Er, not in the cpp file, not being used.

dretrex:
It looks like the scoping and grabbing it is okay, but I don't see where you use the command line argument anywhere.  Did it get missed?

epengr:
As you suggested, it no longer uses a command line argument - if the detector connects then the GUI elements spawn.  If not doesn't they don't.

I thought I"d removed all the QApplication includes since they are no longer used.  I'll fix that one.

epengr:
Sorry...  That was as @helfrij suggested  :)

source/beamline/IDEAS/IDEASBeamline.cpp
```diff
@@ -27,6 +27,9 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "beamline/AMAdvancedControlDetectorEmulator.h"
 
+#include <QApplication>
```
dretrex: If we're going to be pedantic on the QApplication includes, there is one here too.

source/beamline/IDEAS/IDEASBeamline.h
```diff
@@ -40,6 +40,9 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 #include "beamline/IDEAS/IDEASKETEKDetector.h"
 #include "beamline/IDEAS/IDEAS13ElementGeDetector.h"
 
+#include <QApplication>
```
dretrex: And here.

epengr:
Argh, thank you.  Got it.

PR #1635, opened 2015-09-30
Pull Request for Issue1576:  Implement a range feature for all data sources

dretrex:
This adds a cached data range to all the analysis blocks that implement my computeCachedValues routine.  It definitely shows another example where AMAnalysisBlock/AMDataSource could be refactored, but I'll leave that for another day.

It may appear that the rangeFinder() method may be expensive, but it is okay due to QVector implementing implicit sharing and there is no writing happening inside of rangeFinder(), minimumFinder(), and maximumFinder().

Would you like to tackle this one @helfrij?

@epengr, could you have a quick peak or test that there are no issues with the REIXS analysis blocks?  I noticed that they were missing an axisValues() implementation so I added one.  Whether it is before this gets merged in or not isn't particularly important as it should be a 100% transparent change.


buildTests/AMDetectorSetTest/AMDetectorSetTest.pro
```diff
@@ -2,5 +2,5 @@ include ( $$PATH_TO_AM/compositeCommon/AMDetectorSet.pri )
 
 TARGET = AMDetectorSetTest
 
-SOURCES += \
+SOURCES *= \
```
helfrij: Just curious, what's the difference between += and *=?

dretrex: += can be read as "always add" and *= can be read as "add if you haven't yet".  *= is preferred because if you doubly add a file into source and headers and they are not properly header scoped (#define CLASS_NAME_H) then you will get linker errors.

source/analysis/AM1DNormalizationAB.h
```diff
@@ -87,7 +87,10 @@ class AM1DNormalizationAB : public AMStandardAnalysisBlock
 	/// When the independent values along an axis is not simply the axis index, this returns the independent value along an axis (specified by axis number and index)
 	virtual AMNumber axisValue(int axisNumber, int index) const;
 	/// Performance optimization of axisValue():  instead of a single value, copies a block of values from \c startIndex to \c endIndex in \c outputValues.  The provided pointer must contain enough space for all the requested values.
-    virtual bool axisValues(int axisNumber, int startIndex, int endIndex, double *outputValues) const;
+	virtual bool axisValues(int axisNumber, int startIndex, int endIndex, double *outputValues) const;
+
+	/// Returns the cached range of the data contained within the data source.  This is always valid because it is always recomputed when the data is recomputed.
+	virtual AMRange dataRange() const { return cachedDataRange_; }
```
helfrij: I notice that this getter is repeated for every AB class that implements data caching (?). Would it make sense to promote this to AMStandardAnalysisBlock?

source/analysis/AM2DAdditionAB.cpp
```diff
@@ -250,5 +254,6 @@ void AM2DAdditionAB::computeCachedValues() const
             cachedData_[j] += data.at(j);
     }
 
+    cachedDataRange_ = AMUtility::rangeFinder(cachedData_);
```
helfrij: I like the idea of collecting together 'misc' static functions into this common AMUtility class. NIce!

source/analysis/AM2DAdditionAB.h
```diff
@@ -77,13 +80,15 @@ protected slots:
...
-    /// Flag for knowing whether we need to compute the values.
-    mutable bool cacheUpdateRequired_;
-    /// The vector holding the data.
-    mutable QVector<double> cachedData_;
+	/// Flag for knowing whether we need to compute the values.
+	mutable bool cacheUpdateRequired_;
+	/// The vector holding the data.
+	mutable QVector<double> cachedData_;
+	/// Holds the cached data range.
+	mutable AMRange cachedDataRange_;
```
helfrij: Could some of these properties exist in an AMCachedDataAB that the blocks that use data caching inherit from?

source/analysis/AM2DNormalizationAB.cpp
```diff
@@ -30,7 +32,9 @@ AM2DNormalizationAB::AM2DNormalizationAB(const QString &outputName, QObject *par
 	canAnalyze_ = false;
 	dataName_ = "";
 	normalizationName_ = "";
-    cacheUpdateRequired_ = false;
+	cacheUpdateRequired_ = false;
```
helfrij: Some AB's seem to intitialize cacheUpdateRequired_ to true and some to false in the AB constructor. Is it implementation dependent, or should there be a standard way of initializing this value?

source/util/AMUtility.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+			if (value > maximum)
+				maximum = value;
+		}
+	}
+
+	return maximum;
+}
+
+AMRange AMUtility::rangeFinder(const QVector<double> &data, double valueToIgnore)
```
helfrij: Just double-checking, does the valueToIgnore argument allow ranges to be identified for datasets with a specific value for invalid data?

source/analysis/AM3DAdditionAB.cpp
```diff
@@ -253,5 +257,6 @@ void AM3DAdditionAB::computeCachedValues() const
             cachedData_[j] += data.at(j);
     }
 
+    cachedDataRange_ = AMUtility::rangeFinder(cachedData_);
```
helfrij: Should this call to rangeFinder() include a valueToIgnore argument?

buildTests/AMDetectorSetTest/AMDetectorSetTest.pro
helfrij: Ah, okay. Cool!

helfrij:
I had a couple of questions about initializing constructor values, when to use valueToIgnore argument, and how useful inheritance would be to help simplify things. Other than that, looks good @dretrex.

Time Estimate: 30 minutes

source/analysis/AM2DAdditionAB.h
dretrex: Yes, they certainly could.  Unfortunately, that involves making a database upgrade since every single analysis block would need to be redone.  I'd love to do some AMDataSource/AMAnalysisBlock refactoring, but at the moment it seems like a rather large undertaking.

helfrij: Ah, okay.

source/analysis/AM1DNormalizationAB.h
dretrex: You are quite right, this is 100% code duplication.  I thought about putting it inside AMStandardAnalysisBlock but didn't add it in because then every AB would need to implement and fill the `cachedDataRange_`, not just the ones that do caching.

source/analysis/AM2DNormalizationAB.cpp
dretrex: They should all be false to start with I believe.  You should only say you have to update the data after you have sources to get data from.  I can look into these again to make sure they are consistent.  Good catch!

source/analysis/AM1DNormalizationAB.h
helfrij: Okay, makes sense. I was wondering if it could be like AMControl::move(), where the interface provides a way for a subclass to move and a default implementation, but it's up to the subclass to give the real implementation if they actually want to move. Could providing this kind of interface be a different way to skirt having to do a huge database upgrade? Or would it needlessly complicate AMStandardAnalysisBlock?

source/util/AMUtility.cpp
dretrex: Yes, that's the intended idea.

source/analysis/AM3DAdditionAB.cpp
dretrex: Probably.  I'd need to modify the caching to ensure a single invalid value is used as I believe it is not just -1 like the rest of them.  I probably need to do something to ensure that if the value is -1 then the data is always -1.

source/analysis/AM1DNormalizationAB.h
dretrex: Well, it kind does that already.  Unfortunately, AMStandardAnalysisBlock is subclassed by more AB than just what we could call an AMStandardCachedAB where it would be able to take advantage of that kind of paradigm.  Now that someone else has some database upgrade experience, maybe the time do some refactoring is coming closer than maybe originally thought? ;)

source/analysis/AM3DAdditionAB.cpp
helfrij: I wasn't sure if invalid values were something that were implemented by default; if there's no invalid value, then there's no need to have that argument. Gotcha.

dretrex: Unfortunately not :(  A default that was implemented by default would be great, but we've mostly just been getting along by trying to assume that -1 is "not too bad".  Far from ideal but mostly a legacy design decision/hack.

dretrex:
Fixed initialization inconsistency.  Other comments would be done in a nice, proper refactor of the data source system.  Thanks, @helfrij!
Time Estimate: 20 minutes

Issue #1577, opened 2015-09-01
Update MPlot to take advantage of the range in a data source.


dretrex:
After #1576 is completed, MPlot should be able to be improved to make use of the range feature.  This should minimize it has to ask it's model for information - whether it's for painting, finding the range, or populating the colormap.  This will likely involve modifying all the drawing and model classes.

dretrex:
As it turns out, with the new caching of the other data sources, it doesn't appear that there are any changes to MPlot since all the changes happen inside of AMDataSourceSeriesData, AMDataSourceImageData, and AMDataSourceImageDatawDefault (which is technically a model class for MPlot).

dretrex:
Time Estimate: 1 hour

dretrex:
Done in #1639.

PR #1639, opened 2015-10-01
Pull Request for Issue1577: Update MPlot to take advantage of the range in a data source.

dretrex:
From #1577.
> As it turns out, with the new caching of the other data sources, it doesn't appear that there are any changes to MPlot since all the changes happen inside of AMDataSourceSeriesData, AMDataSourceImageData, and AMDataSourceImageDatawDefault (which is technically a model class for MPlot).

Therefore, this was actually much more straight forward as all the work was done inside the previous issues (and using the new AMUtility class again).

Assigned to @helfrij.

### DIFF
--- a/source/dataman/datasource/AMDataSourceImageData.cpp
+++ b/source/dataman/datasource/AMDataSourceImageData.cpp
@@ -66,7 +66,15 @@ void AMDataSourceImageData::setDataSource(const AMDataSource* dataSource) {
 	}
 
     onSizeChanged();
-	onDataChanged(AMnDIndex(0,0), AMnDIndex(xSize_-1, ySize_-1));
+    onDataChanged(AMnDIndex(0,0), AMnDIndex(xSize_-1, ySize_-1));
+}
+
+MPlotRange AMDataSourceImageData::range() const
+{
+	if (cacheUpdateRequired_)
+		updateCachedValues();
+
+	return range_;
 }
 
 double AMDataSourceImageData::x(int index) const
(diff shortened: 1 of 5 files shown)


source/dataman/datasource/AMDataSourceImageData.cpp
```diff
@@ -176,36 +186,15 @@ void AMDataSourceImageData::onDataSourceDeleted()
...
 
-            if (newValue < rangeMinimum)
-                rangeMinimum = newValue;
-        }
+	if (source_->values(start, end, cachedData_.data())){
 
-        range_ = MPlotRange(rangeMinimum, rangeMaximum);
-		updateCacheRequired_ = false;
+		AMRange dataSourceRange = AMUtility::rangeFinder(cachedData_);
+		range_ = MPlotRange(dataSourceRange.minimum(), dataSourceRange.maximum());
```
helfrij: Just curious, what's the difference between MPlotRange and AMRange?

helfrij:
Looks good, @dretrex.

Time Estimate: 10 minutes.

source/dataman/datasource/AMDataSourceImageData.cpp
dretrex: AMRange is a class I made a long time to encapsulate a range and tries to be fairly full fledged by offering things like isNull(), isValid(), etc.
MPlotRange is a typedef of a QPointF that was introduced at the beginning of MPlot because it did a decent enough job of providing the necessary functionality.

PR #1650, opened 2015-10-05
Pull Request for Issue1647: Implementing AMControl::hasChild().

helfrij:
Added new method hasChild(AMControl*) to AMControl. 

Tested on BioXAS using the JJ slits and the carbon filter farm:
- controls were correctly able to return true when asked about one of their children, false for not a child.
- controls were correctly able to return true when asked for a child of one of their children.

### DIFF
--- a/source/beamline/AMControl.h
+++ b/source/beamline/AMControl.h
@@ -308,6 +308,8 @@ public:
 			return children_.at(index);
 		return NULL;
 	}
+	/// Returns true if this control has children and if one of them (or one of their children) matches the given control. Returns false otherwise.
+	bool hasChildControl(AMControl *control);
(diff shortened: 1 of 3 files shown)


source/beamline/AMControl.h
```diff
@@ -308,6 +308,8 @@ class AMControl : public QObject {
 			return children_.at(index);
 		return NULL;
 	}
+	/// Returns true if this control has children and if one of them (or one of their children) matches the given control. Returns false otherwise.
+	bool hasChildControl(AMControl *control);
```
dretrex: This would be a `const' function.

dretrex:
Only the one comment.  Looks good otherwise.
Time Estimate: 2 minute

source/beamline/AMControl.h
helfrij: You're right--added the 'const' keyword.

PR #1651, opened 2015-10-05
Pull Request for Issue1630a: Creating valves control for all BioXAS valves

helfrij:
Task list:
- [x] BioXASValves - creates instances of each specific valve control, and the master valve control.
- [x] BioXASValveControl - reports on the current 'valves open' state, can open valves but not close them.
- [x] Add BioXASValves to BioXASBeamline.


dretrex:
Hey @helfrij, can you add an estimate for this?

helfrij:
Ooops, I forgot it. Thanks @dretrex.

source/application/BioXAS/BioXAS.h
```diff
@@ -8,12 +8,16 @@ namespace BioXAS
...
 	/// Handles the XRF detector choice.  Available choices are None, KETEK, and 13-element Ge.
 	enum FluorescenceDetector {
-
 		NoneXRFDetector = 0,
 		FourElementVortexDetector = 1,
 		Ge13ElementDetector = 2
 	};
 
+	/// Enum to describe each of the three BioXAS beamlines.
+	enum Beamline {
```
dretrex: Purely a naming thing, but is Beamline the best description for this enum?  I don't have a better one off the top of my head, but I thought I'd bring it up anyway.

source/beamline/AMPseudoMotorControl.h
```diff
@@ -100,7 +100,7 @@ protected slots:
 	/// Updates the current value.
 	virtual void updateValue() = 0;
 	/// Updates the moving state.
-	virtual void updateMoving() = 0;
+	virtual void updateMoving() { return; }
```
dretrex: `return;` statement unnecessary.

source/beamline/BioXAS/BioXASValves.h
```diff
@@ -0,0 +1,157 @@
...
+	CLSBiStateControl *vvr3_;
+	/// The fourth front end vacuum valve control.
+	CLSBiStateControl *vvr4_;
+	/// The fifth front end vacuum valve control.
+	CLSBiStateControl *vvr5_;
+
+	/// The set of Side beamline vacuum valves.
+	AMControlSet *sideValves_;
+	/// The first Side beamline vacuum valve.
+	AMControl *vvrSide1_;
```
dretrex: Reason for switching from `CLSBiStateControl` to `AMControl` from here on out?

source/beamline/BioXAS/BioXASValvesControl.cpp
```diff
@@ -0,0 +1,421 @@
...
+		result = imagingValveSet_;
+		break;
+	default:
+		break;
+	}
+
+	return result;
+}
+
+bool BioXASValvesControl::isOpen() const
```
dretrex: From a high level perspective, I agree if any set of valves is closed then the overall valves control is not open, however, I would not agree that if not all valves are closed then the valves control is not closed unless you want to provide some kind of mixed state.  It is quite possible that the front end valves are closed and side open, what should be the overall state displayed to a user?  Something to consider.  What are your thoughts on how the overall state should be be displayed to users?

source/beamline/BioXAS/BioXASValvesControl.h
```diff
@@ -0,0 +1,121 @@
...
+	/// Returns true if this control can move right now.
+	virtual bool canMove() const;
+
+	/// Returns true if the given value is a valid value for this control. False otherwise.
+	virtual bool validValue(double value) const;
+	/// Returns true if the given value is a valid setpoint for this control. False otherwise.
+	virtual bool validSetpoint(double value) const;
+
+	/// Returns the valve set for the named beamline.
+	AMControlSet* valveSet(BioXAS::Beamline beamline);
```
dretrex: Should be a const function.

source/beamline/BioXAS/BioXASValvesControl.h
```diff
@@ -0,0 +1,121 @@
...
+	/// Returns the Imaging valve set.
+	AMControlSet* imagingValveSet() const { return imagingValveSet_; }
+
+	/// Returns true if this control's value is Open, false otherwise.
+	bool isOpen() const;
+	/// Returns true if this control's value is Closed, false otherwise.
+	bool isClosed() const;
+
+	/// Returns a string representation of the given value.
+	QString valueToString(BioXASValvesControl::Value value);
```
dretrex: Same with this one.

dretrex:
Aside from a few comments, this looks good to me.
Time Estimate: 10 minutes

source/application/BioXAS/BioXAS.h
helfrij: I agree. I think Beamline or Endstation is probably more appropriate when considering only Side, Main, and Imaging. I include the FrontEnd that all three share in common here. Not sure what the best name would be.

source/beamline/AMPseudoMotorControl.h
helfrij: Removed.

source/beamline/BioXAS/BioXASValves.h
helfrij: No reason. Just copy-paste error, I think.

source/beamline/BioXAS/BioXASValvesControl.cpp
helfrij: This valves control is going to be used in the beam on/off controls for BioXAS, and the requirement for 'beam on' for the front end is that all valves on Side, Main, and Imaging are all open. Any other state would be be indicative of 'beam off'. I was thinking about this control from that perspective, though I see it can certainly be made to be more flexible. Is your recommendation to: 
- add additional 'Between' state
- consider only two states: Open, NotOpen

source/beamline/BioXAS/BioXASValves.h
helfrij: Fixed.

source/beamline/BioXAS/BioXASValvesControl.h
helfrij: Added const.

source/beamline/BioXAS/BioXASValvesControl.h
helfrij: Added const.

source/beamline/BioXAS/BioXASValvesControl.cpp
dretrex: I think that is too restrictive.  The side, main, and imaging valves should not be connected to the overall shutter status since it is very conceivable that Imaging valves could be closed and users running on Main.  Additionally, I believe the closed state should be more inclusive.  If any valve is closed, the valve group should be closed.

Am I under the wrong impression when it comes to the location of these valves?  Do all valves need to be open for any specific beamline to be running?

source/application/BioXAS/BioXAS.h
dretrex: Okay.  It's not the end of the world or anything, if we don't come up with something better before we merge it in, not a big deal.

source/beamline/BioXAS/BioXASValvesControl.cpp
helfrij: I think this is actually a machine protection thing: no beamline can be operational when any vacuum valve on any beamline is closed. But I will definitely double-check with David B on this.

helfrij: Ah, I see what you mean about the Closed state. Those should be ORs in BioXASValvesControl::isClosed() instead of ANDs. I've made that change.

helfrij: Yup, the confirmed requirement is: no beamline is operational when any vacuum valve on any beamline is closed. In this respect, the beamlines act like a unit and I've been assured that there will never (never!) arise a situation where a valve would be closed on one beamline and another would still be running. To get this to work, the pvs themselves would have to be circumvented; Acquaman wouldn't know the difference either way.

dretrex: Okay, then.  Sounds good.  I was not aware of that requirement.

dretrex:
We are now good with these latest changes right?  Or was there anything else you wanted to change before I merge this in?

helfrij:
I think that's everything. Thanks @dretrex.

dretrex:
Time estimate for extra changes?

helfrij:
Time Estimate: 10 minutes

PR #1652, opened 2015-10-06
Pull Request for Issue1630b: Creating front-end beam status control

helfrij:
This is the first of the beam status controls to be developed for BioXAS and will be used to control the front-end beam status. All shutters and valves considered are before any beamline optical equipment. This control maps the states of valves and shutters to a beam "On" or "Off" state.
- listens to photon shutter 1, vacuum valve, fast valve, photon shutter 2, safety shutter, beamline valves.
- is "On" when all valves and shutters are open.
- is "Off" when any one of the valves and shutters are closed.
- "turns on" by opening all valves and shutters in sequence: fast valve, vacuum valve, photon shutter 1, beamline valves, safety shutter, photon shutter 2.
- "turns off" by closing the photon shutter 2 and safety shutter.

Implementation of this control is broken down into the following classes:
- [x] Create BioXASFrontEndShutters - manages instances of front end shutter controls
- [x] Add instance of BioXASFrontEndShutters to BioXASBeamline.
- [x] Create BioXASBeamStatusControl - abstract class for all beam status controls to inherit from
- [x] Create BioXASFrontEndBeamStatusControl - reports front end ON/OFF status, according to the shutters, valves
- [x] Add instance of BioXASFrontEndBeamStatusControl to BioXASBeamline.

(Time estimate: 3.5 hours.)


source/beamline/BioXAS/BioXASFrontEndBeamStatusControl.cpp
```diff
@@ -0,0 +1,229 @@
...
+	if (isConnected()) {
+		result = (valves_->isOpen() && safetyShutter_->isOpen() && photonShutterDownstream_->isOpen() && photonShutterUpstream_->isOpen());
+	}
+
+	return result;
+}
+
+bool BioXASFrontEndBeamStatusControl::isOff() const
+{
+	bool result = false;
```
dretrex: This could also just be `return !isOn();` based on the logic used.

source/beamline/BioXAS/BioXASFrontEndBeamStatusControl.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+	return result;
+}
+
+AMAction3* BioXASFrontEndBeamStatusControl::createBeamOnAction()
+{
+	AMListAction3 *beamOn = new AMListAction3(new AMListActionInfo3("Turn beam on", "Turn beam on"), AMListAction3::Sequential);
+	beamOn->addSubAction(AMActionSupport::buildControlMoveAction(valves_, BioXASValvesControl::Open));
+	beamOn->addSubAction(AMActionSupport::buildControlMoveAction(safetyShutter_, CLSBiStateControl::Open));
+	beamOn->addSubAction(AMActionSupport::buildControlMoveAction(photonShutterDownstream_, CLSBiStateControl::Open));
```
dretrex: Is this order backwards?  It's probably supposed to go valves -> safety shutter -> photon shutter upstream -> photon shutter downstream.

dretrex:
Looks good otherwise.
Time Estimate: 8 minutes

source/beamline/BioXAS/BioXASFrontEndBeamStatusControl.cpp
helfrij: I picked this order assuming that an upstream photon shutter couldn't be opened unless all downstream shutters/valves were also open. Is that not the case?

source/beamline/BioXAS/BioXASFrontEndBeamStatusControl.cpp
helfrij: Yup, that's true. Will change.

source/beamline/BioXAS/BioXASFrontEndBeamStatusControl.cpp
dretrex: Yes, but you are mixing up upstream and downstream.  Downstream is closest to the sample, upstream is closer to the storage ring.

helfrij: Hmmm. Sorry, I'm not sure I see the problem. The last shutter that should be opened is the upstream photon shutter, correct? That one appears to be the last action added to the list?

dretrex: The downstream shutter should be the last one opened.  That is the final shutter to let light into the SOE.

helfrij: I think in that order, there could arise a situation where the upstream shutter is open and the downstream shutter is closed. Is that okay? If so, I'll make the change.

dretrex: Yes, that is typically the intended use case.  You have already set up the situation for that to happen with the beam off only closing the downstream shutter.  This is often a case where beamlines will "warm up" the optics by opening up everything before the SOE to let all the optical equipment thermally stabilize.

helfrij: I'll double-check with David B. on what should be done for both "beam on" and "beam off." Maybe what you're pointing out is a different error I've made--the upstream shutter is the one that should be closed for beam off. This control is meant to control the beam in the front-end only--all shutters considered here are before any optical equipment, I believe.

helfrij:
I spoke with David B today about the beam status control. The control should now be correct in terms of the valves/shutters considered (added a front end vacuum valve and a fast valve), and the beam on/off actions.

helfrij:
Time estimate: 1 day

PR #1655, opened 2015-10-09
Pull Request for Issue1565: SGM Energy Pseudomotor

iainworkman:
Added a pseudo motor class for controlling the SGM Energy.

Tracks the grating angle, grating translation, exit slit position and undulator position PVs in order to produce an energy value.
Contains child controls for the grating angle, grating translation, exit slit position and undulator position so these can be interfaced with manually.


iainworkman:
Hey @helfrij. Would you mind having a look over this pull request for me?

source/beamline/SGM/energy/SGMEnergyControl.cpp
```diff
@@ -0,0 +1,363 @@
...
+{
+    return true;
+}
+
+bool SGMEnergyControl::shouldStop() const
+{
+    return true;
+}
+
+bool SGMEnergyControl::canMeasure() const
```
helfrij: Do the features available to the child controls impact whether SGMEnergyControl canMeasure()/canMove()/canStop()? Eg. can the energy control stop if the grating angle control can't stop?

source/beamline/SGM/energy/SGMEnergyControl.cpp
```diff
@@ -0,0 +1,363 @@
...
+		energyPositionController_->blockSignals(false);
+    }
+}
+
+void SGMEnergyControl::onEnergyPositionControllerEnergyChanged(double value)
+{
+    setValue(value);
+}
+
+void SGMEnergyControl::initializeEnergyPositionController()
```
helfrij: Just checking: is the reason that this code in initializeEnergyPositionController() isn't in the constructor is because it should only be executed when SGMEnergyControl is connected for the first time?

source/beamline/SGM/energy/SGMEnergyControl.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+void SGMEnergyControl::updateMoving()
+{
+    setIsMoving(gratingAngleControl_->isMoving() ||
+                gratingTranslationControl_->isMoving() ||
+                undulatorPositionControl_->isMoving() ||
+                exitSlitPositionControl_->isMoving());
+}
+
+void SGMEnergyControl::updateValue()
```
helfrij: I'm guessing that updateValue() doesn't work for you in this context, if it's not implemented here. Is it currently pure virtual in AMPseudoMotorControl? Maybe that should change.

source/beamline/SGM/energy/SGMEnergyControl.cpp
iainworkman: I believe they are fairly simple. If they are connected, then they can measure/stop. I'm not entirely sure about whether it is possible for them to move if already moving.

@davidChevrier Are there any quirks with these controls in this regard?

source/beamline/SGM/energy/SGMEnergyControl.cpp
iainworkman: Yeah, because all the inputs are rigged up to the energyPositionController_, that is the thing which is taking all those input positions and calculating the resulting value. It is thus just easier for me to rig up to that than to perform the calculation here in this method.

It is a pure virtual function in the base class. It could change, though I'm not sure if there will likely be too many cases where this kind of thing is necessary.

source/beamline/SGM/energy/SGMEnergyControl.cpp
iainworkman: Yes. The energyPositionController_ needs those values for its initialization, or else the calculations it contains tend to hit NaN.

source/beamline/SGM/energy/SGMGratingTranslationStepControl.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+	return true;
+}
+
+bool SGMGratingTranslationStepControl::shouldStop() const
+{
+	return true;
+}
+
+bool SGMGratingTranslationStepControl::canMeasure() const
```
helfrij: Same note here with canMeasure()/canMove()/canStop() as for the SGMEnergyControl. If the gratingTranslationStepPV_ can only measure/move/stop sometimes, maybe include a check here: gratingTranslationStepPV_->canMeasure() (etc).

source/beamline/SGM/energy/SGMGratingTranslationStepControl.h
```diff
@@ -0,0 +1,77 @@
+#ifndef SGMGRATINGTRANSLATIONSTEPCONTROL_H
+#define SGMGRATINGTRANSLATIONSTEPCONTROL_H
+
+#include "beamline/AMPseudoMotorControl.h"
+/*!
+  * Pseudo-motor class which acts as a wrapper for the grating translation PV,
+  * which does not update its moving status if asked to move within tolerance.
+  */
+class SGMGratingTranslationStepControl : public AMPseudoMotorControl
```
helfrij: Is it necessary to have AMPseudoMotorControl as an ancestor for this class? It seems like this class gets most of its functionality from asking the gratingStepTranslationPV_ control for information. What additional functionality does this class provide?

source/tests/SGM/SGMEnergyPositionTest.cpp
```diff
@@ -0,0 +1,168 @@
...
+	testIndividualRequestEnergy("Silicon K Edge Start", 1830.0, SGMGratingSupport::HighGrating, -171813.59, 567.5737302, 20.77155529, -59199.75);
+	testIndividualRequestEnergy("Silicon K Edge Middle", 1860.0, SGMGratingSupport::HighGrating, -169042.40, 601.0803091, 20.94762194, -57120.40);
+	testIndividualRequestEnergy("Silicon K Edge End", 1890.0, SGMGratingSupport::HighGrating, -166359.19, 635.0727254, 21.12242885, -55055.93);
+}
+
+void SGMEnergyPositionTest::testResultantEnergy()
+{
+    SGMEnergyPosition testEnergyPosition(240, SGMGratingSupport::LowGrating);
+
+    if(!qFuzzyCompare(testEnergyPosition.resultantEnergy(), 240)) {
```
helfrij: Just curious: what does qFuzzyCompare do?

helfrij:
I had a few (minor) questions about two AMPseudoMotorControl subclasses. Other than that, it looks good. 

Time Estimate: 45 minutes

source/beamline/SGM/energy/SGMEnergyControl.cpp
helfrij: Ah okay.

source/beamline/SGM/energy/SGMGratingTranslationStepControl.h
iainworkman: The grating translation PV is badly behaved. It doesn't alter its status when you ask it to move within tolerance of its current position. It seemed that the easiest way to cope with this on the AM side was just to make it a child of AMPseudoMotorControl, and have that fake the move for me.

source/beamline/SGM/energy/SGMGratingTranslationStepControl.cpp
iainworkman: Will have a look at both of these cases.

source/tests/SGM/SGMEnergyPositionTest.cpp
iainworkman: It does a 'close enough' comparison between two floating point values. I only use it here because its a test class (it has some issues if either value can possibly be zero, which makes it pretty useless in other situations).

this call is essentially equivalent to 

qAbs(testEnergyPosition.resultantEnergy() - 240) < someToleranceValue

(I'm not sure what someToleranceValue qt uses)

source/beamline/SGM/energy/SGMGratingTranslationStepControl.h
helfrij: Ah, okay.

source/tests/SGM/SGMEnergyPositionTest.cpp
helfrij: Cool!

PR #1658, opened 2015-10-14
PullRequest for Issue1616: add AMWaveformBinningSinglePVControl to AMPVControl

XuetaoLiu-CLS:


source/beamline/AMPVControl.cpp
```diff
@@ -209,6 +209,13 @@ AMControl::FailureExplanation AMPVControl::move(double setpoint) {
 	return NoFailure;
 }
 
+// Start a move to the value setpoint:
+AMControl::FailureExplanation AMPVControl::move(const QString &stringSetPoint)
```
dretrex: Could you explain why this is required?  It also is not very safe since it doesn't check anything, nor does it listen to any potential preset parameters.

source/beamline/AMPVControl.cpp
```diff
@@ -672,6 +650,106 @@ void AMPVwStatusAndUnitConversionControl::setUnitConverters(AMAbstractUnitConver
...
+}
+
+double AMWaveformBinningSinglePVControl::value() const{
+	if(attemptDouble_)
+		return readPV_->binFloatingPointValues(lowIndex_, highIndex_);
+	return readPV_->binIntegerValues(lowIndex_, highIndex_);
+}
+
+bool AMWaveformBinningSinglePVControl::values(int size, int *outputValues) const{
+	if(attemptDouble_)
```
dretrex: Why do we ignore floating point values here but not in the read only control?

dretrex: Nevermind, I see what is happening now.  You'd probably be better off having just a single values and just cast the integers to doubles if the type isn't right.

source/beamline/AMPVControl.cpp
```diff
@@ -672,6 +650,106 @@ void AMPVwStatusAndUnitConversionControl::setUnitConverters(AMAbstractUnitConver
...
+//	qDebug() << "Attempting QVector<int>";
+	if(attemptDouble_)
+		return;
+
+//	qDebug() << "Doing QVector<int> on " << writePV_->pvName();
+
+	writePV_->setValues(((dbr_long_t*)(values.data())), values.count());
+}
+
+void AMWaveformBinningSinglePVControl::setValues(const QVector<double> values){
```
dretrex: You should make sure that these vectors are `const QVector<double> &values)` to guarantee that implicit sharing is being used.

dretrex:
I have a couple high level questions.  I'm sure we'll have some back and forth about them.
Time Estimate: 5 minutes

iainworkman:
Hey guys, is there any update on this? I think I'm going to require this PV for my hexapod data recorder interface.

source/beamline/AMPVControl.cpp
XuetaoLiu-CLS: @dretrex in the old Amptek codes, it added a function `setString()` to the AMPVControl, so that the Detector manager can write string to some PVs.
```
void AMPVControl::setString(const QString &string){
	writePV_->setValue(string);
}
```

In stead of name the function `setString`, I overloaded the function `move()` with the ability to write a string

source/beamline/AMPVControl.cpp
XuetaoLiu-CLS: The whole class is from the original Amptek codes.

dretrex:
Okay, I think this is alright.  This class could do with some robustness, I think, but it will do for the time being.

XuetaoLiu-CLS:
Yes, I agree for the for the robustness. So far, it is doing what it should do.

dretrex:
Okay, I'll merge this in.  @iainworkman, feel free to use this now.

PR #1662, opened 2015-10-15
PullRequest for Issue1661: Register SXRMB database informtion

XuetaoLiu-CLS:


XuetaoLiu-CLS:
This doesn't work as expected since we are mapping the facility ID with facility description (AMFacility) by using the order they added to the DB instead of the real ID.

XuetaoLiu-CLS:
This should solve the issue now!

XuetaoLiu-CLS:
Need database upgrade for this.

dretrex:
I'm going to merge this in now since the other aspect of this problem is already an issue.  Thanks for looking into this @XuetaoLiu-CLS!

davidChevrier:
Time Estimate: 2 minutes

PR #1669, opened 2015-10-18
Pull Request for Issue1648: Create action for calibrating controls

helfrij:
Created new action classes for calibrating controls. These classes are based heavily off of existing AMControlMoveAction/Info.
- [x] AMControlCalibrateActionInfo
- [x] AMControlCalibrateAction

Added an inline function to AMActionSupport that creates and returns a calibration action when supplied with a control and two values to calibrate with (oldValue, newValue).

Implemented the change in BioXASSSRLMonochromatorEnergyControl::createCalibrateAction(), where the AMActionSupport version of getting the bragg motor calibration action is now used. 

The goal of testing is to make sure that existing functionality is not affected by the new changes. Making small changes to the bragg motor calibration and the energy control calibration through the BioXASSSRLMonochromatorConfigurationView seem to show that no changes have occurred.

### DIFF
--- a/source/actions3/actions/AMControlCalibrateAction.cpp
+++ b/source/actions3/actions/AMControlCalibrateAction.cpp
@@ -1,0 +1,121 @@
+#include "AMControlCalibrateAction.h"
+#include "beamline/AMControl.h"
+#include "beamline/AMBeamline.h"
+#include "util/AMErrorMonitor.h"
+
+AMControlCalibrateAction::AMControlCalibrateAction(AMControlCalibrateActionInfo *info, AMControl *control, QObject *parent) :
+	AMAction3(info, parent)
+{
+	// Initialize class variables.
+
+	control_ = 0;
+
+	// Current settings.
+
+	setControl(control);
+}
+
+AMControlCalibrateAction::AMControlCalibrateAction(const AMControlCalibrateAction &original) :
+	AMAction3(original)
+{
+	setControl(original.control());
+}
+
+AMControlCalibrateAction::~AMControlCalibrateAction()
+{
+
+}
+
+void AMControlCalibrateAction::setControl(AMControl *control)
+{
+	if (control)
+		control_ = control;
+	else
+		control_ = AMBeamline::bl()->exposedControlByInfo(controlCalibrateInfo()->controlInfo());
+}
+
+void AMControlCalibrateAction::onCalibrateStarted()
+{
+	// Set action as started.
+
+	setStarted();
+
+	// Start progress timer updates.
+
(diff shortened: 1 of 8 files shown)


helfrij:
@iainworkman @XuetaoLiu-CLS
Would one of you mind taking a look at this pull request?

iainworkman:
Looks good to me.
Time Estimate: 10 Minutes

helfrij:
Thanks, @iainworkman!

source/actions3/actions/AMControlCalibrateAction.cpp
```diff
@@ -0,0 +1,121 @@
...
+	if(!control_->canCalibrate()) {
+		QString fundamentalFailureMessage = QString("There was an error calibrating the control '%1', because the control reported it cannot calibrate.").arg(control_->name());
+		AMErrorMon::alert(this, AMCONTROLCALIBRATEACTION_CONTROL_CANT_CALIBRATE, QString("%1. Please report this problem to the beamline staff.").arg(fundamentalFailureMessage));
+		setFailed(fundamentalFailureMessage);
+		return;
+	}
+
+	// Check that the calibration values are valid.
+
+	if (controlCalibrateInfo()->oldValue() == controlCalibrateInfo()->newValue()) {
```
dretrex: Is it desirable to have the calibration fail because it is already calibrated?  Wouldn't it make more sense to have it succeed?  I think this will have bad workflow issues if a calibration action was part of a more complicated workflow.  For example, if this action was inside of a scan's initialization, it would fail and stop the scan.

helfrij:
Can't sneak anything by you. I'll take a look when I get in tomorrow morning.

source/actions3/actions/AMControlCalibrateAction.cpp
helfrij: I think what I meant to do was to prevent a calibration from happening when the old value and the new value haven't been set in the action info. They are initialized to the same value in the action info, and an assumption was made that they would be 'valid' if they were different. A better way to do this may be to use AMNumbers instead of doubles and initialize them to AMNumber::Null or AMNumber::InvalidError.

dretrex: Ah, I see.  Yeah, I think if that is the case, then making the values AMNumbers would be better (or some other kind of test, like hasValidValues() or something).

As it relates to my original comment, it might be preferable to have the action automatically succeed if the values are actually the same.

helfrij: Sounds good, I'll make the change.

PR #1674, opened 2015-10-29
Pull Request for Issue1645: Scans not always added to loop actions when using loop mode

dretrex:
Simple change to the method `createMultipleScans()`.

### DIFF
--- a/source/ui/acquaman/AMScanConfigurationViewHolder3.cpp
+++ b/source/ui/acquaman/AMScanConfigurationViewHolder3.cpp
@@ -122,8 +122,13 @@ void AMScanConfigurationViewHolder3::setEnabled(bool enabled){
 
 AMAction3 * AMScanConfigurationViewHolder3::createScan()
 {
-	if(view_)
-		return new AMScanAction(new AMScanActionInfo(view_->configuration()->createCopy()));
+	if(view_){
+
+		const AMScanConfiguration *config = view_->configuration();
+
+		if (config)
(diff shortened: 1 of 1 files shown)


source/ui/acquaman/AMScanConfigurationViewHolder3.cpp
```diff
@@ -133,15 +138,13 @@ AMAction3* AMScanConfigurationViewHolder3::createMultipleScans()
...
-				loopAction->addSubAction(createScan());
-				return loopAction;
-			} else {
-				return createScan();
-			}
+
+			AMLoopAction3 *loopAction = new AMLoopAction3(new AMLoopActionInfo3(iterationsBox_->value(), config->name(), config->description()));
+			loopAction->addSubAction(createScan());
+
+			return loopAction;
```
helfrij: Just checking that I understand. Before, the view holder created either scan actions or loop actions, depending on the desired number of scans. With these changes, the scan configuration view holder now always produces loop actions for scans--a single scan is now a loop action of one scan?

dretrex: Correct.  If you instantiate the view holder such that it you want it to create loops (`enableLoopAction = true` in the constructor) it will always create loops regardless of the number of scans you select to start.  If `enableLoopAction` is false then it will simply create the single scan like it always has.

helfrij:
Time Estimate: 2 minutes.

PR #1676, opened 2015-10-29
Pull Request for Issue1603: Name not updating in VESPERS endstation motor group view

dretrex:
Turns out some previous iteration of code with what I presume is copy and paste error got the names of the controls screwed up.  I have now fixed them and the system works as it is supposed to.


helfrij:
Looks good, @dretrex.

Time Estimate: 20 seconds.

PR #1699, opened 2015-11-04
Pull Request for Issue1692: Analysis blocks quite slow on initial computation while running in scans

dretrex:
Added a list of indices to be used for updating the cache rather than the entire thing every time.  This is necessary for large scans as going through all the data on every point is unrealistic and time consuming.

### DIFF
--- a/source/analysis/AM1DNormalizationAB.h
+++ b/source/analysis/AM1DNormalizationAB.h
@@ -130,6 +130,8 @@ protected:
 
 	/// Flag for knowing whether we need to compute the values.
 	mutable bool cacheUpdateRequired_;
+	/// A list of indices of values that need to be updated.  These are meant for when scans are running and data is coming in a pixel at a time.
+	mutable QList<AMnDIndex> dirtyIndices_;
 	/// The vector holding the data.
 	mutable QVector<double> cachedData_;
 	/// Holds the cached data range.
--- a/source/analysis/AM4DBinningAB.cpp
+++ b/source/analysis/AM4DBinningAB.cpp
@@ -124,6 +124,7 @@ void AM4DBinningAB::setInputDataSourcesImplementation(const QList<AMDataSource*>
 		}
 
 		cacheUpdateRequired_ = true;
+		dirtyIndices_.clear();
 		cachedData_ = QVector<double>(size().product());
 
 		setDescription(QString("%1 Summed (over %2)")
@@ -204,6 +205,7 @@ void AM4DBinningAB::setInputSource()
 		}
 
 		cacheUpdateRequired_ = true;
+		dirtyIndices_.clear();
 		cachedData_ = QVector<double>(size().product());
 
 		setDescription(QString("%1 Summed (over %2)")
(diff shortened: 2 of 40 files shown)


source/analysis/AM1DNormalizationAB.h
```diff
@@ -130,6 +130,8 @@ protected slots:
 
 	/// Flag for knowing whether we need to compute the values.
 	mutable bool cacheUpdateRequired_;
+	/// A list of indices of values that need to be updated.  These are meant for when scans are running and data is coming in a pixel at a time.
+	mutable QList<AMnDIndex> dirtyIndices_;
```
iainworkman: Are there any analysis blocks which don't use this caching method, and as such don't require the dirty indices and cached data?

It seems that these reappear in a lot of analysis blocks, so it makes sense for them to be in a common base (either AMAnalysisBlock, AMStandardAnalysisBlock or a new AMCachedAnalysisBlock).

source/analysis/AMAdditionAB.cpp
```diff
@@ -104,6 +104,15 @@ bool AMAdditionAB::axisValues(int axisNumber, int startIndex, int endIndex, doub
...
 	cacheUpdateRequired_ = true;
+
+	AMnDIndex scanStart = start;
+	AMnDIndex scanEnd = end;
+	scanStart.setRank(start.rank()-1);
+	scanEnd.setRank(end.rank()-1);
+
+	if (scanStart == scanEnd)
+		dirtyIndices_ << start;
+
```
iainworkman: I feel like I'm missing something here with respect to the dirtyIndices_ only being set when the source values change one at a time. What happens when a range of values changes and start != end?
(this applies to all the onInputSourceValuesChanged(...)  calls)

source/analysis/AM4DBinningAB.cpp
```diff
@@ -254,33 +256,25 @@ void AM4DBinningAB::computeCachedValues() const
...
 
-	case 3:
+	else {
 
-	    start = AMnDIndex(0, 0, 0, sumRangeMin_);
-	    end = AMnDIndex(inputSource_->size(0)-1, inputSource_->size(1)-1, inputSource_->size(2)-1, sumRangeMax_);
-	    break;
+		start = dirtyIndices_.first();
+		end = dirtyIndices_.last();
 	}
```
iainworkman: I notice that when each index is added to the list of dirty indices it is appended to the end, and then we assume the range here to be from .first() to .last(). Is there and case where the onInputSourceValuesChanged(...) signals might be received out of order?

source/analysis/AM1DNormalizationAB.h
dretrex: Ah, you have noticed the same thing @helfrij did in my last big analysis block update.  Yes, you are 100% correct that a lot of this could be done in a subclass.  However, as it stands right now, it would not be able to be in AMStandardAnalysisBlock because not all AB use this caching method (mostly old ones that are there for legacy reasons).  And unfortunately, AMCachedAnalysisBlock that would inherit from AMStandardAnalysisBlock would take more effort than I am willing at the moment because all AB's are AMDbObjects and every AB that has been modified would need a database upgrade (or one massive one that updated the database with the new schema).  I would love, more than anything, to do a code refactor of our analysis blocks.  Remove the old ones, update the current ones, and just make the whole system a little more robust.

source/analysis/AMAdditionAB.cpp
dretrex: At the moment, there are only two ways analysis blocks update.  It is either one value at a time (step scanning) or full repaint (from size change, source change, etc.).

source/analysis/AM4DBinningAB.cpp
dretrex: Not at present, and it technically wouldn't matter since the mapping is absolute.

davidChevrier:
Time Estimate: 2.5 hours

PR #1700, opened 2015-11-04
Pull Request for Issue1618: Add It and I0 to the generic scan configuration

dretrex:
This adds I0 to the generic scan engine that we've been using on BioXAS and SGM.  For the time being it only auto adds normalized channels to regions of interest, but further logic could be added based on desires.

### DIFF
--- a/source/analysis/AMNormalizationAB.h
+++ b/source/analysis/AMNormalizationAB.h
@@ -1,0 +1,100 @@
+#ifndef AMNORMALIZATIONAB_H
+#define AMNORMALIZATIONAB_H
+
+#include "analysis/AMStandardAnalysisBlock.h"
+
+/// This analysis block accepts any number of identically ranked data sources and divides one by the other to normalize one designated as "data".
+class AMNormalizationAB : public AMStandardAnalysisBlock
+{
+	Q_OBJECT
+
+	Q_CLASSINFO("AMDbObject_Attributes", "description=Normalization Analysis Block")
+public:
+	/// Constructor.
+	Q_INVOKABLE AMNormalizationAB(const QString &outputName = "InvalidInput", QObject *parent = 0);
+	/// Destructor.
+	virtual ~AMNormalizationAB();
+
+	/// Description.
+	QString infoDescription() const { return QString(); }
+
+	/// Check if a set of inputs is valid. The empty list (no inputs) must always be valid. For non-empty lists, our specific requirements are...
+	/*!
+	- the rank() of all sources are the same.
+	*/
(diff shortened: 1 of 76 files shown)


source/analysis/AMNormalizationAB.h
```diff
@@ -0,0 +1,100 @@
+#ifndef AMNORMALIZATIONAB_H
+#define AMNORMALIZATIONAB_H
+
+#include "analysis/AMStandardAnalysisBlock.h"
+
+/// This analysis block accepts any number of identically ranked data sources and divides one by the other to normalize one designated as "data".
```
iainworkman: Is the 'any number' part of this comment correct? It looks to me like you have a single data source to analyse and the normalizing data source.

iainworkman:
Okay, I'd just like to see if I've got the logic of this correct in my head.
- The Scan configuration can have a single detector listed as the i0
- The controller will automatically add an analysis block for each detector with the same rank as the i0, which normalizes it against the i0 value.
- The config view now allows the user to choose any of the selected detectors as the i0.

source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
```diff
@@ -483,10 +503,22 @@ void AMGenericStepScanConfigurationView::onDetectorSelectionChanged(QAbstractBut
...
+
 				configuration_->addDetector(detector->toInfo());
-			else
+				i0ComboBox_->addItem(detector->name());
+			}
+
+			else {
+
 				configuration_->removeDetector(detector->toInfo());
+				i0ComboBox_->removeItem(i0ComboBox_->findText(detector->name()));
```
iainworkman: Should there be some logic here as to what the i0 choice will be in the case that the detector which is currently the i0 is unchecked? I'm not 100% sure on the behvaiour of the QComboBox when the current selected item is removed, but I think it just selects the item with the previous index (if there is one) or following (if not). 

Might it be best to select "None" if the selected detector is removed, just so that it's consistent?

iainworkman:
Other than the couple of questions posted, this looks good. 
Time Estimate: 1 hour

source/analysis/AMNormalizationAB.h
dretrex: Yup.  You can pass in a list of data sources and then choose which two you want to analyze.

dretrex:
You are correct with your assessment of what this is supposed to do.

source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
dretrex: That is probably a good idea.  I'll take a look.

dretrex:
Time Estimate: 15 minutes

PR #1702, opened 2015-11-05
Pull Request for Issue1701: Change all scan controllers to use CDF instead of writing to text files

dretrex:
Changed all the scan controllers that used the ascii region file loader to use CDF instead.


iainworkman:
I notice that the XRF scans for various beamlines use the vespers2011XRF format still. Is it intentional that these aren't being switched to CDF?

(There's also the SGMFastScanActionController which uses the format sgm2013Fast, but this is from the old SGMAcquaman so I'm not sure if this really matters).

iainworkman:
Time Estimate: 10 Minutes

dretrex:
I think the XRF scans can be dealt with when we get around to doing #1191.  You are quite right about the SGM fast scan controller, we will ignore it because it is old and we will be using CDF for the new one.

Issue #1689, opened 2015-11-02
19) Diagnostic Ladder Controls


iainworkman:
There are new diagnostic ladders which have been added to the beamline as part of the upgrade. These will need control interfaces (AMPseudoMotors?) which can easily be worked with inside Acquaman.

regiert:
The ladders are all on CLS motors (no feedback).  The should have their
velocity PV's (:divisor) set to 1 so they move as fast as possible - this
has been tested and their is no problem running at these velocities.  All
of these motors have been calibrated using the CCW limit switch as a zero
reference.  This means that the calibration procedure should be that you
run the motors to the CCW and put a 1 to the :isCalib PV.  The calibration
should not have to be run regularily as these motors do not appear to lose
steps during normal operation.  When the motor crate/IOC is power cycled,
the calibration will have to be performed.  Acquaman should have a way of
recognizing that the calibration is void and either notify the operator or
automatically run the calibration procedure.

XAS diagnostic ladder (upstream of KB) SMTR16114I1014- just use centre
positions

0 - Upper limit - Vanadium filter
6700 to 14400 - Photodiode
17600 to 23600 - Au Mesh
31200 to 37200 - Pinhole
43000 to 49500 - Passthrough
68000 - Phosphor

XPS diagnostic ladder (upstream of EA1) SMTR16114I1012

0 - 3400 Upper limit - Photodiode
10000 - 14000 - Vanadium Filter
18000 - 24000 - Au Mesh
26000 and higher - Retracted from beam

Bypass diagnostic ladder SMTR16114I1013

0 - Upper limit - Slit
12000 - Passthrough
27000 - Phosphor

Note that the values provided for the different positions may need to be
adjusted and should be allowed to be reconfigured by beamline staff.  This
would suggest that a configuration screen is required for each ladder that
contains the step values for the different positions and an option to
recalibrate (drive to CCW limit and set :isCalib).


Tom Regier, PhD.
Beamline Scientist, SGM Beamline
Canadian Light Source, Inc.
tom.regier@lightsource.ca
(306) 657 - 3733

helfrij:
Initial implementation completed in #1707.
Time Estimate: 5 hours

Issue #1527, opened 2015-08-11
Refactor the carbon filter farm family of classes


helfrij:
The carbon filter farm classes could use an update, when time is available. I think the code would be simpler to read/change if the classes are structured more like the XIA filters control family, less dependent on enum values and no beamline-specific controls.

helfrij:
Time Estimate: 2 days.

PR #1703, opened 2015-11-05
Pull Request: general ladder-type control

helfrij:
The existing carbon filter farm controls on BioXAS (#1527) and the proposed diagnostic ladder control (#1689) seem to have some functionality in common:
- are pseudomotors -- they translate feedback from a 'real' motor into new information.
- have enumerated values -- the value can be one of a set of discrete value options.

I've made two classes that handle the enumeration needs for the carbon filter farm controls:
- [x] AMEnumeratedControl
- [x] AMSingleEnumeratedControl

The general enumerated control handles managing a list of the available option indices, matching each index to a string representation, and generating enum and move enum states from these lists. Now inheriting from AMEnumeratedControl:
- [x] BioXASCarbonFilterFarmFilterControl

AMSingleEnumeratedControl inherits from AMEnumeratedControl and additionally manages the relationship between the control and a single subcontrol. This is basically a convenience subclass, as I anticipate most AMEnumeratedControl subclasses will depend on just one subcontrol. Now inheriting from AMSingleEnumeratedControl:
- [x] BioXASCarbonFilterFarmActuatorWindowControl
- [x] BioXASCarbonFilterFarmActuatorFilterControl

There are many more possible applications for these controls on BioXAS, and hopefully they will make implementing the ladder control on SGM easier too.


dretrex:
Overall, I think it looks good.  Definitely a good start.

As a pedantic detail, I would say that the name could be a bit more general.  To represent the abstraction, maybe something like `AMPseudoMotorEnumeratedControl` or something that showcases that this subclass is meant for enumerated pseudomotors.

Time Estimate: 10 minutes

helfrij:
Quite right--I've renamed the control.

helfrij:
Hey @dretrex. I've done some more work on the controls and some testing with the carbon filter farms on Side and Main. So far, so good. Are you available to take another look at these classes?

dretrex:
I only have one question about the changes.  What is the use case for the duplicate entries?  Can you give an example?  Otherwise, changes seem good.

helfrij:
For the BioXAS carbon filter farm controls, it didn't make sense to have duplicate entries (eg. two '50' filter options) even though there are actually two valid filters with thickness 50. But I didn't want to restrict the general class unnecessarily.

dretrex:
Okay.  Is this good to be merged in or were you wanting to do more work on it?

helfrij:
It's done. Thanks @dretrex.

helfrij:
Time Estimate: 3 days

Issue #1566, opened 2015-08-27
2a) Extending the Energy pseudo-motor control to handle continuous movements


dretrex:
Presumably, if the control developed in #1565 is well designed, this should hopefully be a rather straightforward extension of the control to handle the addition logic developed in #1562.  However, there is not a large knowledge base on generic continuous movement and this will likely involve a lot of ironing kinks out due to technical realities.

PR: #1704

iainworkman:
Okay, so I've tested undulator motions coordinated with the grating angle over the three grating selections (forwards and backwards) and it is 'working'.

Performing rudimentary observations** on the motions have show:
- They seem to start at the same time. There is no obvious notable delay from either.
- The undulator seems to reach its target just before the grating angle. On all moves I tested this was 1s (+/- 0.2s) whether on a 10s or 20 move.

** cagetting the movement statuses of the undulator and grating angle, visually noting how synchronized they seem on start and then working out the delta time based on the epics outputs. I said rudimentary, didn't I?

iainworkman:
To Do:
- [ ] Fix issue where the exit slit is being moved to the optimized position for the start, then the mean  position for the trajectory, wasting time.

iainworkman:
To Do:
- [ ] Add energy trajectory PV to energy coordinator.
- [ ] Add widgets for energy trajectory to the SGMEnergyView.

iainworkman:
Time Estimate: 36 hours

PR #1704, opened 2015-11-09
Pull Request for Issue 1566: Continuous Energy Motion

iainworkman:
- Added ability to do continuous trajectory movements of energy to the pseudomotor
- Added PVs for trajectory movement to the db file
- Added the trajectory movement PVs to the coordinator, in order to pass along the move to the pseudomotor
- Added the new trajectory movement PVs to Acquaman, so they can be written to.
- Added a new view for controlling the trajectory motion, and added it to the SGMEnergyView.

#1566


source/beamline/SGM/SGMEnergyCoordinator.cpp
```diff
@@ -556,7 +559,8 @@ void SGMEnergyCoordinator::onEnergySetpointPVChanged(double value)
 	if(energySetpointInitialized_)  {
 
 		qDebug() << QString("Forwarding energy setpoint value of %1 to energy control").arg(value);
-		energyControlCoordinator_->move(value);
+		AMControl* controlVersion = energyControlCoordinator_;
```
dretrex: Is there a reason this isn't just `energyControlCoordinator_->move(value);`?

source/beamline/SGM/SGMEnergyCoordinator.cpp
```diff
@@ -576,6 +580,27 @@ void SGMEnergyCoordinator::onEnergyStopPVChanged(double)
 	}
 }
 
+void SGMEnergyCoordinator::onEnergyTrajectoryStartPVChanged(double value)
+{
+	qDebug() << "\tTrajectory start value of" << value << "received";
```
dretrex: Are these the debugs for terminal information or just testing ones?

source/beamline/SGM/energy/SGMEnergyControl.cpp
```diff
@@ -162,65 +152,162 @@ AMControl *SGMEnergyControl::gratingTranslationControl() const
...
+
+	moveAction->start();
+
+	return AMControl::NoFailure;
 }
 
 void SGMEnergyControl::setUndulatorHarmonic(SGMUndulatorSupport::UndulatorHarmonic undulatorHarmonic)
 {
-    if(energyPositionController_) {
+	if(energyPositionController_) {
```
dretrex: We really need to get these tabs sorted out.

source/beamline/SGM/SGMEnergyCoordinator.cpp
iainworkman: It is to stop the initial setting of the energy during start up from being actually prompting the energy control from making a full move.
I did this because there may be a case where someone moves to a given energy with undulator tracking off, then for some reason restarts the coordinator. If this check wasn't here when the energy is calculated from the grating on start up, and this value written to the energy setpoint (to ensure it's at the right value to being with) the undulator (which because tracking was off during the initial move to this position isn't at the right spot) would begin to move. 

In short I didn't want a situation to arise where the undulator would begin to move just because the application restarted.

source/beamline/SGM/SGMEnergyCoordinator.cpp
iainworkman: This one was just for testing. I will remove it.

source/beamline/SGM/energy/SGMEnergyControl.cpp
iainworkman: Yep. Sorry :/

source/beamline/SGM/energy/SGMEnergyTrajectory.cpp
```diff
@@ -1,199 +1,218 @@
...
+		startEnergyPosition_->setAutoDetectUndulatorHarmonic(false);
+		endEnergyPosition_->setAutoDetectUndulatorHarmonic(false);
+
+		startEnergyPosition_->setUndulatorHarmonic(SGMUndulatorSupport::FirstHarmonic);
+		endEnergyPosition_->setUndulatorHarmonic(SGMUndulatorSupport::FirstHarmonic);
+	}
+
+	time_ = time;
+
+	gratingAngleVelocityProfile_ = AMTrapezoidVelocityProfile(startEnergyPosition_->gratingAngle() * gratingAngleStepsPerEncoderCount,
```
dretrex: The conversion here is due to the fact that the acceleration and velocity are in steps but all of the positions are in real units, right?

iainworkman: Yes, that is correct.

source/beamline/SGM/energy/SGMEnergyTrajectory.cpp
```diff
@@ -1,199 +1,218 @@
...
 double SGMEnergyTrajectory::endEnergy() const
 {
-    return endEnergy_;
+	return endEnergy_;
+}
+
+double SGMEnergyTrajectory::energyVelocity() const
+{
+	if(time_ != 0) {
+		return qAbs(endEnergy_ - startEnergy_) / time_;
```
dretrex: It is okay that this can return negative numbers, right?

source/beamline/SGM/energy/SGMGratingAngleControl.cpp
```diff
@@ -0,0 +1,227 @@
...
+	                                                                    "Moving Grating Angle"),
+	                                              AMListAction3::Sequential);
+
+
+	if(isClosedLoop()) {
+
+		moveAction->addSubAction(AMActionSupport::buildControlMoveAction(encoderControl_, setpoint));
+		moveAction->addSubAction(AMActionSupport::buildControlWaitAction(encoderControl_, setpoint, 20, AMControlWaitActionInfo::MatchWithinTolerance));
+	} else {
+		;
```
dretrex: Rogue ;!  Get 'em!

source/beamline/SGM/energy/SGMUndulatorControl.h
```diff
@@ -0,0 +1,87 @@
...
+#define DEFAULT_UNDULATOR_VELOCITY 5000
+#define DEFAULT_UNDULATOR_ACCELERATION  2500
+
+#include "beamline/AMPseudoMotorControl.h"
+
+class SGMUndulatorControl : public AMPseudoMotorControl
+{
+	Q_OBJECT
+public:
+	explicit SGMUndulatorControl(QObject *parent = 0);
```
dretrex: It looks like the `SGMUndulatorControl` and the `SGMGratingControl` are almost identical.  As a future cleanup, I'm pretty sure they could inherit from a single class that does all the work.

source/beamline/SGM/energy/SGMEnergyTrajectory.cpp
iainworkman: Will double check that. I think it should be an absolute value.

source/ui/SGM/SGMEnergyTrajectoryView.h
```diff
@@ -0,0 +1,39 @@
...
+	Q_OBJECT
+public:
+	explicit SGMEnergyTrajectoryView(SGMEnergyControlSet* controlSet,
+	                                 QWidget *parent = 0);
+
+signals:
+
+public slots:
+protected slots:
+	void onStartPushButtonClicked();
```
dretrex: A couple header comments would be good.

source/util/AMTrapezoidVelocityProfile.cpp
```diff
@@ -1,49 +1,62 @@
...
-    return targetVelocity_;
+	return time_;
 }
 
-double AMTrapezoidVelocityProfile::returnAcceleration() const
+double AMTrapezoidVelocityProfile::targetVelocity() const
 {
-    return returnAcceleration_;
+	double halfAccelByTime = (acceleration_*time_) / 2;
+	double toRoot = pow(halfAccelByTime, 2) - acceleration_ * qAbs(endPosition_ - startPosition_);
```
dretrex: Why `qAbs()`? I'm trying to remember a function that would require the qAbs and can't put my finger on it.

source/util/AMTrapezoidVelocityProfile.cpp
```diff
@@ -1,49 +1,62 @@
...
+
+	if(toRoot < 0) {
+		return halfAccelByTime + sqrt(qAbs(toRoot));
```
dretrex: Same here.  Something seems a bit fishy if you need to qAbs this to me.

source/beamline/SGM/energy/SGMEnergyTrajectory.cpp
iainworkman: This is fine, I believe. The validator should pick up cases where the time < 0.

source/beamline/SGM/SGMEnergyCoordinator.cpp
dretrex: That makes sense with with the setpoint initialization, but you are also setting the control here for no reason.

dretrex:
Overall, this looks good.  But I am going to reiterate that we need to get those tabs sorted out.  That was probably >4000 lines of impact in tabs alone and very time consuming.
Time Estimate: 35 minutes

source/util/AMTrapezoidVelocityProfile.cpp
iainworkman: This was arrived on with help from David. Essentially this bottom case is when the acceleration is not sufficient to reach the target velocity within the delta T. In that case this simply needs to ensure its a value above that peak, which this does.

dretrex: Okay. Sounds good.

Issue #1684, opened 2015-11-02
8) Continuous Scan Infrastructure (1D)


iainworkman:
Once the energy coordinator is in place it will be necessary to create an interface through which continuous 1d scans can be performed upon a provided generic control. 

PR: #1709

dretrex:
Technically, this will be required for the Hexapod too.

iainworkman:
To do:
- [ ] Figure out why scanning backwards isn't working.
- [ ] Allow the necessary classes access to the trajectory and grating encoder translation value functions.
- [ ] Refine the algorithm for distinguishing the start and end of the scanning portion of a move.

dretrex:
I saw from @davidChevrier's email that backwards scans isn't working.  Do we have any more information on that @iainworkman?  I can potentially brainstorm something up as well.

iainworkman:
I'm looking at it in the coordinator. Things that were working don't seem
to be any more. The velocity calculation is coming out with a different
value when moving backwards. And not just a negative version, a totally
different value. I'm trying to track it down.

davidChevrier:
Items left to complete:
- [ ] Generalize detector actions in the assembler. This will require the new scalerChannelDetector to properly implement sharedTriggerSource to be fully general.
- [ ] Add the actual encoder start point, maybe as a custom detector emulator
- [ ] Issue #1714 
- [ ] Fix the up/down scanning ... because down scanning will not fill the datastore correctly at the moment. It's just an iterating issue really.

davidChevrier:
@iainworkman is doing the actual work for implementing #1714 in #1724. We leave that for it's own branch and merge it back into sgmupgrade at some point. I'm considering this done otherwise.

davidChevrier:
Time Estimate: 35.5 hours

Issue #1691, opened 2015-11-02
9) Data Recorder/ Scalar Pulse Alignment


iainworkman:
Using the TTL signal acquired from the scalar, and the data recorder information acquired from the hexapod driver server, there needs to be some work done aligning the beginning and end of a motion with the data within the data recorder in such a way that the position at a given time can be known correctly.

davidChevrier:
@iainworkman what was your time estimate on this one?
Because I added the necessary interpolation algorithms into this one, it kind of exploded in terms of time. Also really had to bring the 1D Line Scan along for the ride to make sure it was working.
Time Estimate: 35.5 Hours

iainworkman:
Time Estimate: 16 Hours

PR #1709, opened 2015-11-12
Pull Request for Issue1684: Continuous 1D Scans

iainworkman:
Creating the pull request now in order to keep track of changes made by all contributors, namely @dretrex @davidChevrier and myself

#1684


source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -0,0 +1,513 @@
...
+			break;
+		}
+
+		case AMAgnosticDataAPIDefinitions::AxisValueFinished:
+
+			// Not relevant for continuous scans.
+			break;
+
+		case AMAgnosticDataAPIDefinitions::DataAvailable:{
+			AMAgnosticDataAPIDataAvailableMessage *dataAvailableMessage = static_cast<AMAgnosticDataAPIDataAvailableMessage*>(&message);
```
dretrex: We are probably responsible for the memory of this message aren't we?

source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -0,0 +1,513 @@
...
+//	}
+
+	for(int x = 0; x < totalCount; x++){
+		AMDSLightWeightScalarDataHolder *asScalarDataHolder = qobject_cast<AMDSLightWeightScalarDataHolder*>(scalerClientDataRequst->data().at(x));
+		if(asScalarDataHolder){
+			QVector<qint32> oneVector = asScalarDataHolder->dataArray().constVectorQint32();
+
+			int tempRunningSum;
+			QString channelString;
+			for(int y = 0, ySize = oneVector.count(); y < ySize; y++){
```
dretrex: This seems really inefficient.  Not sure of a better way at the moment off the top of my head.

source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -0,0 +1,513 @@
...
+				channelString = scalerChannelIndexMap.value(y);
+				tempRunningSum = scalerChannelRunningSums.value(channelString);
+				tempRunningSum += oneVector.at(y);
+				scalerChannelRunningSums[channelString] = tempRunningSum;
+
+				if( (((x+1)*baseScalerTimeScale) % largestBaseTimeScale) == 0){
```
dretrex: I'm not 100% sure what this is doing.  I understand that we need to rebase it, but I think I'm missing a key detail though.
In any case, this conditional does not depend on `y` and should therefore not be inside the inner loop.

source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -0,0 +1,513 @@
...
+	for(int x = oneAmptekDataRequest->data().count()-1; (x > 0) && !foundMovementStart; x--){
+		AMDSDwellSpectralDataHolder *dataHolderAsDwellSpectral = qobject_cast<AMDSDwellSpectralDataHolder*>(oneAmptekDataRequest->data().at(x));
+		int encoderPulsesInPeriod = dataHolderAsDwellSpectral->dwellStatusData().generalPurposeCounter();
+		if(!foundMovementEnd && encoderPulsesInPeriod > 20){
+			qDebug() << "Found movement end at index " << x;
+			foundMovementEnd = true;
+		}
+		else if(foundMovementEnd && encoderPulsesInPeriod < 1){
+			foundMovementStart = true;
+			initiateMovementIndex = x;
```
dretrex: This is the initial index of an axis?

source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+
+	int scalerInitiateMovementIndex = 0;
+	bool foundScalerMovementStart = false;
+	bool foundScalerMovementEnd = false;
+
+	QVector<qint32> encoderUpVector = scalerChannelVectors.value("EncoderUp");
+	QVector<qint32> encoderDownVector = scalerChannelVectors.value("EncoderDown");
+
+	if(double(continuousConfiguration_->scanAxes().at(0)->regionAt(0)->regionStart()) < double(continuousConfiguration_->scanAxes().at(0)->regionAt(0)->regionEnd())){
```
dretrex: This code could be organized better.

source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -0,0 +1,513 @@
...
+			qDebug() << "Found scaler movement start index " << scalerInitiateMovementIndex;
+		}
+	}
+	*/
+
+	qDebug() << "Amptek: " << initiateMovementIndex;
+	qDebug() << "Scaler: " << scalerInitiateMovementIndex;
+
+	if(!metaDataMap_.contains("GratingEncoderFeedback")){
+		qDebug() << "FATAL ERROR, MISSING READBACK VALUE FOR INITIAL ENCODER POSITION";
```
dretrex: How often do these fatal errors happen?  If it's ever possible via non-development means then there should be more cleanup code inside these if statements.

source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -0,0 +1,513 @@
...
+	scalerEnergyFeedbacks[0] = SGMGratingSupport::energyFromGrating(SGMGratingSupport::LowGrating, startEncoderValue);
+
+	// Loop from the start of the movement to the end and recreate the axis values (energy in this case) from the encoder pulse changes
+	for(int x = scalerInitiateMovementIndex, size = encoderUpVector.count(); x < size; x++){
+		currentEncoderValue += encoderUpVector.at(x) - encoderDownVector.at(x);
+		scalerEnergyFeedbacks[x-scalerInitiateMovementIndex+1] = SGMGratingSupport::energyFromGrating(SGMGratingSupport::LowGrating, currentEncoderValue);
+	}
+
+	bool upScan = false;
+	if(scalerEnergyFeedbacks.at(0) < scalerEnergyFeedbacks.at(scalerEnergyFeedbacks.count()-1))
```
dretrex: Not important, but this could be `if (scalerEnergyFeedbacks.first() < scalerEnergyFeedbacks.last())`

source/beamline/CLS/CLSAdvancedScalerChannelDetector.cpp
```diff
@@ -115,17 +343,17 @@ void CLSAdvancedScalerChannelDetector::onScalerScanningChanged(bool isScanning){
...
 	if(readMode_ == AMDetectorDefinitions::ContinuousRead){
-		QVector<int> allIntReadings = scaler_->reading();
-		int enabledChannelCount = scaler_->enabledChannelCount();
-		int totalDataCount = allIntReadings.at(0)/enabledChannelCount;
-
-		continuousData_.clear();
-		continuousSize_ = totalDataCount;
-		for(int x = 0; x < totalDataCount; x++)
-			continuousData_.append(allIntReadings.at(x*enabledChannelCount + channelIndex_ + 1));
+//		QVector<int> allIntReadings = scaler_->reading();
```
dretrex: Some commented code we want to hang on to for the time being?

source/dataman/datasource/AMDataSourceSeriesData.cpp
```diff
@@ -145,11 +145,11 @@ void AMDataSourceSeriesData::onAxisValuesChanged()
 	cachedDataRectUpdateRequired_ = true;
 	onDataChanged(AMnDIndex(0), AMnDIndex(axisSize_-1));
 }
-
+#include <QDebug>
```
dretrex: Leftover debug.

source/util/AMScalerCountAnalyser.cpp
```diff
@@ -0,0 +1,145 @@
...
+	int previousCountDifference = 0;
+
+	for (int iScalerCount = 1, totalCounts = scalerCounts_.count(); iScalerCount < totalCounts; ++iScalerCount) {
+
+		int currentCountDifference = scalerCounts_.at(iScalerCount - 1) - scalerCounts_.at(iScalerCount);
+
+		if(iScalerCount == 1 && scalerCounts_.at(0) > countNoiseThreshold_ ) {
+			// The first element is above noise tolerance, mark it as a transition
+
+			transitionIndices_.append(0);
```
dretrex: How many points are we expecting here?  Appending to QVector's is generally not a great idea.

source/util/AMScalerCountAnalyser.h
```diff
@@ -0,0 +1,75 @@
...
+	* two elements will be considered a transition.
+	*/
+	AMScalerCountAnalyser(const QVector<double>& scalerCounts,
+	                      double countNoiseThreshold,
+	                      double countChangeThreshold);
+
+	/*!
+	  * A string representation of the analysed results
+	  */
+	QString toString();
```
dretrex: Technically, this would be a `const` function.

dretrex:
Time Estimate: 45 minutes

source/util/AMScalerCountAnalyser.h
iainworkman: Will fix that up

source/util/AMScalerCountAnalyser.cpp
iainworkman: This was the solution which didn't end up working out. It will be revised in #1691 proper (and this function will be removed). However an underlying issue will remain. We won't know how many points of interest there will be until we've analysed the data which comes in. If this is computationally sensitive it would be possible to allocate a size which is equal to the passed in scaler data, at the cost of higher memory usage.

dretrex: We can worry about the optimization later.  You can ignore this comment for the time being and we can go over it more thoroughly with #1691.

source/acquaman/AMContinuousScanActionController.cpp
davidChevrier: I believe we are in. It's not being dealt with right now and I think we need to set up some performance tests to figure out how best to deal with it.

source/acquaman/AMContinuousScanActionController.cpp
davidChevrier: The values in scalerChannelRunningSums[channelString] and scalerChannelRunningSums.value(channelString) depend on y. The running sums act as the counter that's being increased until we hit the (((x+1)*baseScalerTimeScale) % largestBaseTimeScale) == 0 conditional. Then we store that value and rezero the sum.

source/acquaman/AMContinuousScanActionController.cpp
davidChevrier: The data request could return more data than just the move we're interested in. This should return the index of the start of the move assuming that the "last move" in the buffer is the one we're interested in. I believe that assumption has to be valid in the paradigm we're using.

source/acquaman/AMContinuousScanActionController.cpp
davidChevrier: Yeah, I thought this could be optimized, but I wasn't sure what the best way to do it was.

source/acquaman/AMContinuousScanActionController.cpp
davidChevrier: Hopefully never. They can likely be changed to errorMons. Could you handle that @iainworkman ?

source/acquaman/AMContinuousScanActionController.cpp
davidChevrier: That would be more readable.

davidChevrier:
I think we can leave the memory management as a note for when we profile this (it's not clear to me when to clean those up). If @iainworkman could take care of those couple errorMons and the scalerEnergyFeedbacks.first() portion, I think this thing is good to go on my end.

source/beamline/CLS/CLSAdvancedScalerChannelDetector.cpp
davidChevrier: Yep, as a reminder that this needs to be handled when we have trigger/dwell working.

iainworkman:
@dretrex @davidChevrier 
Is there anything else I need to do on this?
Wasn't sure about the reference to errorMons

dretrex:
The errorMons are to replace the qDebugs of "FATAL ERROR!".

iainworkman:
Done

Issue #1705, opened 2015-11-10
Generalize Use of AMDS clientAppController for AMAppController beamline applications


davidChevrier:
Once #1679 is merged in, there will be an immediate need to generalize the use of the AMDS clientAppController. Right now, it's created directly in the instance of the CLSAmptekSDD123DetectorNew class, which would lead to one created for every detector instance. It really needs to be that they all share one between them (as well as the scaler and PV server eventually).

 - Need one instance of the clientAppController ... probably as a singleton paradigm
 - Hopefully the onNetworkSessionOpening(), onNetworkSessionOpened(), onNewServerConnected(), and onServerError() can be moved to that class and just have a signal for when data requests are ready
 - @XuetaoLiu-CLS can likely look over some of the weaker usages I've made of AMDS classes and fix them up

XuetaoLiu-CLS:
Time Estimate: 4 hours

PR #1710, opened 2015-11-12
PullRequest for Issue1705: Generalize the usage AMDSClientAppController

XuetaoLiu-CLS:
Make AMDSClientAppController singleton so that the individual detectors only need to know about the specific server it wants to talk with.

https://github.com/acquaman/acquaman/issues/1705


dretrex:
Looks good.
Time Estimate: 5 minutes

PR #1713, opened 2015-11-14
Pull Request for Issue1626: Binning analysis block editors may be unstable

dretrex:
I found the issue with crashing when the low binning value is greater than the higher binning value.  Fixed by invalidating the analysis block.  

I was unable to reproduce the database bug, so I will have to double check that to make sure there isn't something I'm missing.

@helfrij, could you do the pull request for me on this one?  I am going to assume it's done until I get confirmation on the database bug.

### DIFF
--- a/source/analysis/AM3DBinningAB.cpp
+++ b/source/analysis/AM3DBinningAB.cpp
@@ -224,7 +224,7 @@ void AM3DBinningAB::reviewState()
 
     int s = inputSource_->size(sumAxis_);
 
-    if(sumRangeMin_ >= s || sumRangeMax_ >= s) {
+    if(sumRangeMin_ >= s || sumRangeMax_ >= s || sumRangeMin_ > sumRangeMax_) {
 	setState(AMDataSource::InvalidFlag);
     }
     else
@@ -472,8 +472,8 @@ void AM3DBinningAB::setSumRangeMin(int sumRangeMin)
     cacheUpdateRequired_ = true;
     dirtyIndices_.clear();
     reviewState();
-    emitValuesChanged();
     setModified(true);
+    emitValuesChanged();
 }
 
 void AM3DBinningAB::setSumRangeMax(int sumRangeMax)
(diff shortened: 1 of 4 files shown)


source/analysis/AM3DBinningAB.cpp
```diff
@@ -472,8 +472,8 @@ void AM3DBinningAB::setSumRangeMin(int sumRangeMin)
     cacheUpdateRequired_ = true;
     dirtyIndices_.clear();
     reviewState();
-    emitValuesChanged();
```
helfrij: Was the problem here that emitValuesChanged() was called before setModified(true)?

helfrij:
Looks good, @dretrex.

Time Estimate: 5 minutes.

source/analysis/AM3DBinningAB.cpp
dretrex: I don't think so. The problem was the analysis block being valid when the min value was larger than the max value. I have found that for analysis blocks that the signals should be emitted at the end.

Sent from my iPhone

On Nov 14, 2015, at 6:15 AM, helfrij <notifications@github.com<mailto:notifications@github.com>> wrote:


In source/analysis/AM3DBinningAB.cpp<https://github.com/acquaman/acquaman/pull/1713#discussion_r44858933>:

> @@ -472,8 +472,8 @@ void AM3DBinningAB::setSumRangeMin(int sumRangeMin)
>      cacheUpdateRequired_ = true;
>      dirtyIndices_.clear();
>      reviewState();
> -    emitValuesChanged();


Was the problem here that emitValuesChanged() was called before setModified(true)?

-

dretrex: Hmm, forgot to finish that thought.  I have found that analysis blocks behave better when all the necessary pieces of the function have been dealt with first, then the notifications are sent out once they are all done.

helfrij: Ah okay.

dretrex:
This should be sufficient as the database bug was not reproducible.

PR #1727, opened 2015-11-18
Pull Request for Issue 1723: Coordinated movement API implementation for Hexapod

iainworkman:
- Created functions for retrieving hexapod recorder data on the AM side.
- Created an action list which performs coordinated motions of the hexapod. Specifically it is:
 - Move to start point
 - Save current velocity locally
 - Set a velocity for the distance and time
 - Set the data recorder to a rate which will result in (approx) 125 points per move
 - Turn on the data recorder
 - Trigger the move to the end
 - Wait for the move to the end to complete
 - Turn off the data recorder
 - Reset the velocity back to the saved value


source/beamline/AM3DCoordinatedSystemControl.h
```diff
@@ -100,6 +100,22 @@ class AM3DCoordinatedSystemControl : public AMPseudoMotorControl
...
 	  */
 	virtual void addChildControl(AMControl *control);
+
+	/*!
+	  * Virtual function which performs the calculation required to transform a
+	  * vector movement from the global system to the arbitrary transformed one.
+	  * \param globalVector ~ A vector in the global coordinate system which will
+	  * be tranformed into the arbitrary transformed system.
+	  */
+	virtual QVector3D globalAxisToPrime(const QVector3D& globalVector) const {return globalVector;}
```
dretrex: Is this a placeholder?  Why not make it pure virtual?

source/beamline/SGM/SGMHexapod.cpp
```diff
@@ -193,3 +273,20 @@ void SGMHexapod::resetSystem()
 	zAxisPrimeTrajectoryControl_->resetRotationsToGlobal();
 }
 
+QVector<QVector3D> SGMHexapod::transformVectors(const QVector<QVector3D> &coordinates)
```
dretrex: Not to be overly pedantic, but this would probably be better as a QList<QVector3D>.  We use QVector a lot for data because it guarantees that the data is aligned in memory, but that isn't necessary for general containers.  In general, QList is the way to go when wanting a container, QVector for a data container, and QStringList for a string container.

source/beamline/SGM/SGMHexapodTransformedAxis.cpp
```diff
@@ -29,12 +32,196 @@ SGMHexapodTransformedAxis::SGMHexapodTransformedAxis(AxisDesignation axis,
...
+}
+
+AMAction3 * SGMHexapodTransformedAxis::createWaitForCompletionActions()
+{
+	AMListAction3* waitActions = new AMListAction3(new AMListActionInfo3("Waiting for coordinated move",
+	                                                                     "Waiting for coordinated move"),
+	                                               AMListAction3::Sequential);
+
+	// We need to add 1.5 seconds to the delta time, as the data recorder is slow
+	AMAction3* positionWaitAction = AMActionSupport::buildControlWaitAction(this, lastEndPoint_, lastDeltaTime_+1.5, AMControlWaitActionInfo::MatchWithinTolerance);
```
dretrex: Is this supposed to be `+1.5` or `*1.5`?

dretrex:
Time Estimate: 10 minutes

source/beamline/AM3DCoordinatedSystemControl.h
iainworkman: I will look at this again, as it seemed to to me as I moved them just now. I know my thinking at the time was that if a coordinate system control like this doesn't specify a means of transforming, then if it is asked to transform a vector it should just return that vector untransformed. 
I will look further whether this should actually be here, or at a level below (after all a coordinated system doesn't necessarily have to have a concept of being transformed).

source/beamline/SGM/SGMHexapodTransformedAxis.cpp
iainworkman: +1.5
With the data recorder taking about a second to do its retrieval whatever the length of time over which the motion was performed, I opted for a flat amount.

iainworkman:
@dretrex 
Those changes have been made.
Time Estimate: 10 minutes

source/beamline/SGM/SGMHexapodTransformedAxis.cpp
dretrex: Okay, sounds good.

PR #1728, opened 2015-11-18
Pull Request for Issue 1624: Coordinated movement API implementation for Energy

iainworkman:
Created the required sets of actions for performing continuous moves of the SGM beamline energy.  Also did work to generalize the scan action assembler so it makes use of the general API now.


dretrex:
This looks good @iainworkman.
Time Estimate: 5 minutes

PR #1729, opened 2015-11-18
Pull request for Issue 1719: Moved AMDSClientApp from SGMBeamline to SGMAppController

iainworkman:
Moved the initialization code of the AMDSClientAppController from SGMBeamline to SGMAppController. This necessitated the creation of a function in SGMBeamline `configAMDSServer(const QString &hostIdentifier)` which performs the AMDS config on all the beamline controls which rely on the data server with the provided hostname. The app controller can then make this call each time a data server is connected.


dretrex:
Looks good.
Time Estimate: 2 minutes

PR #1730, opened 2015-11-19
Pull Request for Issue 1718: Added default region to Continuous Scan

iainworkman:
Added a default region to the continuous scan configuration in SGMAppController from 270 to 320 over 10 seconds, and made the default control for the scan 'energy'. Also fixed a bug in the view which meant the signals for the widgets weren't being wired up in cases where the config already had an axis when passed in.

**NOTE:** This branch also contains a number of changes related to include paths performed by David last (Wed) night.

### DIFF
--- a/source/beamline/SGM/SGMBeamline.cpp
+++ b/source/beamline/SGM/SGMBeamline.cpp
@@ -165,7 +165,8 @@ void SGMBeamline::configAMDSServer(const QString &hostIdentifier)
 		amptekSDD4_->configAMDSServer(hostIdentifier);
 	}
 
-	if(hostIdentifier == "10.52.48.40:28044" && scaler_) {
+	if(hostIdentifier == "10.52.48.4:28044" && scaler_) {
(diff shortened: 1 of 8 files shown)


source/beamline/SGM/SGMBeamline.cpp
```diff
@@ -165,7 +165,8 @@ void SGMBeamline::configAMDSServer(const QString &hostIdentifier)
 		amptekSDD4_->configAMDSServer(hostIdentifier);
 	}
 
-	if(hostIdentifier == "10.52.48.40:28044" && scaler_) {
+	if(hostIdentifier == "10.52.48.4:28044" && scaler_) {
+		qDebug() << "\n\nHost identified for scaler\n\n";
```
dretrex: @davidChevrier, do we want to keep all these debugs for the time being?

dretrex:
Looks good.
Time Estimate: 3 minutes

PR #1733, opened 2015-11-23
Pull Request for Issue1696: AMDS Trigger/Dwell Interface

davidChevrier:
I've implemented the necessary Acquaman side pieces to use the new trigger/dwell scaler interface coming out of the AMDSScaler Server.

This is related to the Issue1696_Acquaman branch in AMDS. Between the two, there's a working implementation for trigger/dwell scaler on the AMDS Server side; a working CLSAMDSScaler class with associated classes; a working view class for the CLSAMDSScaler; and, a new CLSAMDSScalerChannelDetector. With all of these together I've run a bunch of step scans using the Commissioning Tool on energy against the scaler channels and the ampteks.


davidChevrier:
Whoops, not into master :|

PR #1734, opened 2015-11-23
Pull Request for Issue1696: AMDS Trigger/Dwell Interface

davidChevrier:
I've implemented the necessary Acquaman side pieces to use the new trigger/dwell scaler interface coming out of the AMDSScaler Server.

This is related to the Issue1696_Acquaman branch in AMDS. Between the two, there's a working implementation for trigger/dwell scaler on the AMDS Server side; a working CLSAMDSScaler class with associated classes; a working view class for the CLSAMDSScaler; and, a new CLSAMDSScalerChannelDetector. With all of these together I've run a bunch of step scans using the Commissioning Tool on energy against the scaler channels and the ampteks.


SGMAcquaman_tests.pro
```diff
@@ -14,9 +14,9 @@ HEADERS += \
     source/tests/SGM/SGMEnergyPositionTest.h \
     source/tests/SGM/SGMEnergyControlTestView.h \
     source/tests/SGM/SGMGratingVelocityTests.h \
-    source/tests/SGM/SGMUndulatorVelocityTests.h \
-    source/tests/SGM/AMScalerAnalysisTest.h \
-    source/tests/SGM/AMScalerTestData.h
+    source/tests/SGM/SGMUndulatorVelocityTests.h #\
+    #source/tests/SGM/AMScalerAnalysisTest.h \
+    #source/tests/SGM/AMScalerTestData.h
```
davidChevrier: @iainworkman and @dretrex 
I removed these because I noticed that a lot of builds really struggled on these files. I think they're really big (and fully of text?) ... I saw at least one case where the OPI hung up trying to do this compile.

iainworkman: Yes, I have removed these files myself in #1691. It was the test mono
scaler data that is doing it. Now I made the test take input from a text
file.

source/beamline/CLS/CLSAMDSScaler.cpp
```diff
@@ -0,0 +1,589 @@
...
+	emit dwellModeChanged(mode);
+}
+
+void CLSAMDSScaler::onDwellTimeControlValueChanged(double time)
+{
+	if (!isConnected())
+		return;
+
+	if(dwellTimeSourceRequested_){
+		dwellTimeSourceRequested_ = true;
```
dretrex: `dwellTimeSourceRequested_ = false`?

source/beamline/CLS/CLSAMDSScaler.h
```diff
@@ -0,0 +1,318 @@
...
+	/// Returns the dwell time tolerance.  Passes the tolerance from the dwell time control.
+	double dwellTimeTolerance() const;
+	/// Returns the current acquisition state
+	double dwellState() const;
+
+	/// Returns the number of channels that are currently enabled
+	int enabledChannelCount() const;
+
+	/// Returns the individual scaler channel provided by \param index.
+	CLSAMDSScalerChannel* channelAt(int index);
```
dretrex: These four methods can be `const`.

source/beamline/CLS/CLSAMDSScaler.h
```diff
@@ -0,0 +1,318 @@
...
+
+	/// Returns the minimum voltage range for the ion chamber.
+	double minimumVoltage() const { return voltageRange_.minimum(); }
+	/// Returns the maximum voltage range for the ion chamber.
+	double maximumVoltage() const { return voltageRange_.maximum(); }
+	/// Returns the voltage range for the ion chamber.
+	const AMRange &voltageRange() const { return voltageRange_; }
+
+	/// Returns whether the voltage is within the linear range of the ion chamber.
+	virtual bool withinLinearRange() const { return voltage() <= maximumVoltage() && voltage() >= minimumVoltage(); }
```
dretrex: You could have used the AMRange API for this: `return voltageRange_.withinRange(voltage());`

source/beamline/CLS/CLSAdvancedScalerChannelDetector.h
```diff
@@ -24,119 +24,6 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "beamline/CLS/CLSBasicScalerChannelDetector.h"
 
-class CLSScalerChannelDetector : public AMDetector
```
dretrex: Was it only SGM that used this class?

dretrex:
Couple comments, seems alright otherwise.  There is definitely empty functions that should be looked at again soon though.
Time Estimate: 20 minutes

source/beamline/CLS/CLSAMDSScaler.cpp
davidChevrier: Quite right.

source/beamline/CLS/CLSAMDSScaler.h
davidChevrier: If they're const, do they have to return const pointers?

dretrex: No.  The const for the function is so modifications to the class can't be done with that method.  If you returned a const pointer then you wouldn't be able to modify that instance.

source/beamline/CLS/CLSAdvancedScalerChannelDetector.h
davidChevrier: Yeah, it was my stepping stone class. It used the old CLSSIS3820Scaler.
The new class uses the new scaler so no one needed this anymore.

source/beamline/CLS/CLSAMDSScaler.h
davidChevrier: Okay. This has always been a point of confusion for me. Since we're returning a pointer to member owned by the class, if we don't return a const pointer then the class will call the function const but the programmer could call non-const functions on the pointer ... effectively taking non-const action against the class as a whole. I'm more than happy to change the function declarations, I just want to make sure I'm on the right page for when to and when not to.

source/beamline/CLS/CLSAdvancedScalerChannelDetector.h
dretrex: Sounds good.

source/beamline/CLS/CLSAMDSScaler.h
dretrex: `const ClassA *ClassB::method() const;`
The first const says that you can't modify the `ClassA` instance that has been given to you.
The second const says `ClassB` will not be modified while you call `method()`.

dretrex:
@davidChevrier, I added an estimate for you.

source/beamline/CLS/CLSAMDSScaler.h
davidChevrier: I've always found the semantics of const so gross in these cases. I can change these up, I have the branch open. Let me know.

dretrex: May as well do it up.

davidChevrier:
All of my changes should be up. Good to go.

PR #1749, opened 2015-11-26
Pull Request for Issue 1743: Fix Energy Scanning

iainworkman:
Changes made:
- Removed all the 'initialization' steps (moving to start, getting the exit slit into position) for a continuous move from within the energy coordinator, as this is now done in the acquaman side.
- Altered the pseudo motor so it now doesn't emit the movingChanged(false) signal whenever its move actions completed (whether or not its child controls are still moving)
- Changed all the createDefaultsAction() functions in the various pseudo-motors to to non const (was required for setting the tolerance of the grating angle, and I wanted them all to be consistent).

Sorry, no laptop today.
Will set this properly tonight, but for now:
Estimate: 1

### DIFF
--- a/source/beamline/SGM/energy/SGMEnergyCoordinatorControl.cpp
+++ b/source/beamline/SGM/energy/SGMEnergyCoordinatorControl.cpp
@@ -163,33 +163,29 @@ AMControl *SGMEnergyCoordinatorControl::exitSlitPositionControl() const
 	return exitSlitPositionControl_;
 }
 
-AMControl::FailureExplanation SGMEnergyCoordinatorControl::move(double startSetpoint, double finalSetpoint, double time)
+AMControl::FailureExplanation SGMEnergyCoordinatorControl::move(double targetSetpoint, double time)
 {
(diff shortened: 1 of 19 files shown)


source/beamline/SGM/energy/SGMEnergyCoordinatorControl.cpp
```diff
@@ -527,33 +516,6 @@ AMAction3 *SGMEnergyCoordinatorControl::createMoveAction(SGMEnergyTrajectory* en
 				                                         AMListAction3::Sequential);
 
 				continuousMoveAction->addSubAction(gratingAngleControl_->createDefaultsAction());
-				// Moving to start position and set exit slit position:
```
dretrex: Is this code that was moved into SGMAcquaman?

dretrex:
Looks good.
Time Estimate: 5 minutes

source/beamline/SGM/energy/SGMEnergyCoordinatorControl.cpp
iainworkman: Yep

PR #1751, opened 2015-11-26
Pull Request for Issue1750: Energy coordinator getting in bad state when PVs disconnect

iainworkman:
- Added check into the SGM Energy Coordinator so that when it loses connection to its required PVs it will gracefully exit with an error message

Again, sorry. No chrome.
Estimate: 1


dretrex:
Looks good.
Time Estimate: 1 minute

PR #1756, opened 2015-11-27
Pull Request for Issue1687: Time Evolution Strip Tool

dretrex:
Here is the first round of the Time Evolution Strip Tool.  It can connect to the scaler and ampteks, but hasn't done anything for the PVs yet.  The PV data server can be connected to and the little bit of code necessary to get it to work once we have that all figured out.  I'm pretty happy with how it turned out for the most part.


davidChevrier:
Time Estimate: 30 minutes

PR #1758, opened 2015-11-27
Pull Request for Issue1754: No Exporter Registered for SGM XAS Scans

dretrex:
Very straight forward registration of exporters for SGM.  Line and map scans are registered with normal ASCII files but I setup the XAS scans to use the XDI format.


davidChevrier:
Time Estimate: 2 minutes

PR #1759, opened 2015-11-27
Pull Request for Issue1753: Continuous Scans Missing Thumbnails

dretrex:
This was actually an issue of doing the scan cleanup (exporting, thumbnail generation, etc.) before the data was placed in the data store.  Fixed the behaviour by calling `onScanningActionsSucceeded()` inside the continuous scan controller rather than relying on the signals from the `scanActionsTree_`.


davidChevrier:
Time Estimate: 5 minutes

PR #1761, opened 2015-11-27
Pull Request for Issue1744: Allow AMDS Detectors to Monitor their Connected State

iainworkman:
Rather than go the whole hog and implement a full isConnected() mechanism for AMDS instruments, I instead ensured that an attempt to acquire from them when there was no server resulted in an appropriate error mon output and a failing of the scan (rather than a segfault or the scan sitting in acquisition forever). 

I suggest a future look at a standard interface for items which require an AMDS, which will include a proper isConnected mechanism.


source/actions3/actions/AMDetectorTriggerAction.cpp
```diff
@@ -61,16 +61,24 @@ void AMDetectorTriggerAction::startImplementation(){
 
 	if(!detector_) {
 		AMErrorMon::alert(this,
-						  AMDETECTORTRIGGERACTION_NO_VALID_DETECTOR,
-						  QString("There was an error trigger the detector '%1', because the detector was not found. Please report this problem to the Acquaman developers.").arg(detectorTriggerInfo()->name()));
+				  AMDETECTORTRIGGERACTION_NO_VALID_DETECTOR,
+				  QString("There was an error trigger the detector '%1', because the detector was not found. Please report this problem to the Acquaman developers.").arg(detectorTriggerInfo()->detectorInfo()->name()));
+		setFailed();
```
dretrex: Did this error not have a `setFailed()` in it to begin with?  That's crazy.

dretrex:
Looks good.
Time Estimate: 2 minutes

source/actions3/actions/AMDetectorTriggerAction.cpp
iainworkman: It did, it's just the way git has ordered the changes here. The old line is on 66 / 74 below

PR #1763, opened 2015-11-28
Pull Request for Issue1738: SGM Continuous Energy Up/Down Auto Select

dretrex:
This ended up being a bit bigger than this particular issue because it involved making modifications in AMScanConfiguration and propagating them to the necessary places.  Preliminary tests show that it appears to work as intended.


davidChevrier:
Time Estimate: 15 minutes

PR #1764, opened 2015-11-30
Pull Request for Issue1685: Continuous Scan Infrastructure (2D)

dretrex:
This enables the continuous scanning in 2 dimensions.  By that, I mean that it does one axis continuously and then moves a step axis for the outer loop.


dretrex:
Hey @iainworkman, could you take a quick look at this?  You don't need to do the whole pull request because it has changes from @davidChevrier's #1691 branch, but if you could look at the SGMHexapodTransformedAxis and SGMHexapodMapScanControllerAssembler to make sure it's at least sane and rational, I'd appreciate it.

source/beamline/SGM/SGMHexapodTransformedAxis.cpp
```diff
@@ -185,6 +200,111 @@ AMAction3 * SGMHexapodTransformedAxis::createSetParameterActions(double startPoi
 	return action;
 }
 
+AMAction3 *SGMHexapodTransformedAxis::createSetParametersActions(const QVector3D &startPoint, const QVector3D &endPoint, double deltaTime)
```
iainworkman: What are you hoping for with this action to set up moves in the native coordinate system?
I'm not sure when these would be used.

dretrex: It is only used in the SGMHexapodMapScanControllerAssembler to explicitly create all the trajectories.  With the trajectory always being "X is the trajectory, Y moves by a step, Z needs to be preserved" this will create all the actions for that specific trajectory.

source/acquaman/SGM/SGMHexapodMapScanActionControllerAssembler.cpp
```diff
@@ -0,0 +1,207 @@
...
+	AMAction3 *initializeControlPosition = AMActionSupport::buildControlMoveAction(stepControl, stepAxis->axisStart());
+	initializeControlPosition->setGenerateScanActionMessage(true);
+	initializationActions->addSubAction(initializeControlPosition);
+	mapActions->addSubAction(initializationActions);
+
+	// There will be only 1 region in these step scans because it is a map.
+	AMScanAxisRegion *stepRegion = stepAxis->regionAt(0);
+	QList<double> stepRegionPositions = generateStepAxisPositions(stepRegion);
+	// Grabbing current position so we know what the z value is.
+	QVector3D currentPosition = QVector3D(SGMBeamline::sgm()->exposedControlByName("Hexapod Global X Axis")->setpoint(),
```
iainworkman: I think this is also going to run into trouble. From looking at it, it seems that it is going to perform the requested scan in the native coordinate system of the hexapod, not the transformed axes.

source/beamline/SGM/SGMHexapodTransformedAxis.cpp
```diff
@@ -185,6 +200,111 @@ AMAction3 * SGMHexapodTransformedAxis::createSetParameterActions(double startPoi
...
+					   AMListAction3::Sequential);
+
+
+
+		// Create the required move actions in the global system:
+		AMListAction3* startMoveActions = new AMListAction3(new AMListActionInfo3("Moving parameter controls for start",
+										     "Moving parameter controls for end"),
+													   AMListAction3::Parallel);
+
+		startMoveActions->addSubAction(AMActionSupport::buildControlMoveAction(globalXAxis_, startPoint.x()));
```
iainworkman: It looks, from usage, that this function is passed vectors which are in the prime coordinate system, and is then using them to set controls for the global system. I think this needs a primeToGlobal() call in here before they are used.

iainworkman:
Other than those couple of points regarding axes transforms this looks good.
30 minutes

dretrex:
Something more like this @iainworkman?

iainworkman:
Yes, that looks correct to me.

PR #1785, opened 2015-12-04
Pull Request for Issue1783: Updating scaler base PV name.

helfrij:
Exactly as it sounds, just modified the scaler base pv name in BioXASMainBeamline.


dretrex:
Looks good.
Time Estimate: 10 seconds

PR #1793, opened 2015-12-11
Pull Request for Issue1792: Adding stop button to M1, M2, mono views

helfrij:
The M1, mono, and M2 classes all already inherit from BioXASBeamline Component, which is a direct descendent of AMControl. Each should be able to implement the stop() functionality that AMControl provides. I refactored the classes, developed a simple stop button for stopping controls, and added it to the views for each.

### DIFF
--- a/source/beamline/BioXAS/BioXASBeamlineComponent.cpp
+++ b/source/beamline/BioXAS/BioXASBeamlineComponent.cpp
@@ -49,26 +49,22 @@ bool BioXASBeamlineComponent::canStop() const
 
 bool BioXASBeamlineComponent::stop()
 {
-	// Want to return true if all the controls are valid and those
-	// that can be stopped are stopped, false otherwise. Returns
-	// true if there are no child controls, the expectation is that
-	// this control can't be moving otherwise.
+	bool result = false;
 
-	bool childrenValid = true;
-	bool childrenStopped = true;
+	if (canStop()) {
 
-	foreach (AMControl *control, children_) { // We want to iterate through all children and try to stop them.
+		bool childrenStopped = true;
 
-		if (control) {
-			if (control->canStop())
+		// Iterate through all child controls, attempting
+		// to stop them.
+
+		foreach (AMControl *control, children_) {
+			if (control)
 				childrenStopped &= control->stop();
-
-		} else {
-			childrenValid &= false;
 		}
-	}
 
-	bool result = childrenValid && childrenStopped;
+		result = childrenStopped;
+	}
 
 	return result;
 }
--- a/source/beamline/BioXAS/BioXASBeamlineComponent.cpp
+++ b/source/beamline/BioXAS/BioXASBeamlineComponent.cpp
@@ -1,5 +1,4 @@
 #include "BioXASBeamlineComponent.h"
-#include <QDebug>
 
 BioXASBeamlineComponent::BioXASBeamlineComponent(const QString &name, QObject *parent) :
 	AMControl(name, "", parent)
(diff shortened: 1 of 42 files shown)


source/beamline/BioXAS/BioXASBeamlineComponent.cpp
```diff
@@ -1,4 +1,5 @@
 #include "BioXASBeamlineComponent.h"
+#include <QDebug>
```
dretrex: Still need the debugs?

source/beamline/BioXAS/BioXASBeamlineComponent.cpp
```diff
@@ -11,9 +12,42 @@ BioXASBeamlineComponent::~BioXASBeamlineComponent()
...
+{
+	// Want to return true if all the controls are valid and those
+	// that can be stopped are stopped, false otherwise. Returns
+	// true if there are no child controls, the expectation is that
+	// this control can't be moving otherwise.
+
+	bool childrenValid = true;
+	bool childrenStopped = true;
+
+	foreach (AMControl *control, children_) {
```
dretrex: Can `children_` contain a null pointer?  If so, that would be a bit troubling since you could have a situation where `childrenValid` was false, but `childrenStopped` was true.

source/beamline/BioXAS/BioXASBeamlineComponent.h
```diff
@@ -13,6 +13,16 @@ class BioXASBeamlineComponent : public AMControl
 	/// Destructor.
 	virtual ~BioXASBeamlineComponent();
 
+	/// Returns true if this control should be able to be stopped (if connected), false otherwise.
+	virtual bool shouldStop() const { return true; }
```
dretrex: Is this true for all BioXASBeamlineComponent classes?  Are components all motors?

source/ui/beamline/AMControlStopButton.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+void AMControlStopButton::setStyleSheet(const QString &styleSheet)
+{
+	stopButton_->setStyleSheet(styleSheet);
+}
+
+void AMControlStopButton::onStopButtonClicked()
+{
+	qDebug() << "\n\nStop button clicked.";
```
dretrex: Debugs still required?

dretrex:
Overall, this looks good.  Although I didn't see an obvious way, it feels like your `setXYZMotor()` methods could be turned into some kind of factory function.  Just an observation.
Time Estimate: 15 minutes

source/beamline/BioXAS/BioXASBeamlineComponent.cpp
helfrij: Good catch. Shouldn't be needed any longer, I've removed it.

source/beamline/BioXAS/BioXASBeamlineComponent.cpp
helfrij: BioXASBeamlineComponent::addChildControl() will not add a null child, but there is no restriction in AMControl::addChildControl(). It could happen, technically. I've moved the checks for whether the children are valid and can be stopped to canStop(), maybe that helps to clarify?

source/ui/beamline/AMControlStopButton.cpp
helfrij: Nope, I've removed it.

source/beamline/BioXAS/BioXASBeamlineComponent.h
helfrij: Hmmm. I'm a little hazy on the difference is between shouldStop() and canStop(). My impression is that shouldStop() is a more 'relaxed' requirement: it returns true if a control could potentially provide stopping functionality, if it is connected. I guess this implies there is something to connect to, like a motor. Is that true?

PR #1795, opened 2015-12-14
Pull Request for Issue1794: Exporter bug for incomplete scans

dretrex:
Turns out the problem was a random occurrence where if the x-axis values were -1.XYZ that the cast would be the same.  I removed the cast, which is safe for these cases because the values are set as -1 and are not manipulated in any way - therefore the double comparison isn't super shady. 

@iainworkman, can you look at this one for me?


dretrex:
Hey @iainworkman, would you be able to take a look at this?  It shouldn't take long.

iainworkman:
Sure

iainworkman:
Looks good to me.
Time Estimate: 5 Minutes

PR #1656, opened 2015-10-13
Pull Request for Issue1630c: Implementing Side-specific beam on/off controls

helfrij:
This pull request is an instalment for #1630, where I'm developing beam status controls for BioXAS. The changes here are mainly for the creation of three Side-specific controls and some views for each: 
- BioXASSidePOEBeamStatusControl
- BioXASSideSOEBeamStatusControl
- BioXASSideBeamStatusControl

The first depends on the front-end beam status control (#1652), the pre-mirror (m1) mask, the pre-mono mask. It should be read-only for now, as there is not yet enough information about where the desired setpoints should be for the mirror mask (may change depending on beamline configuration).
- is "On" when the front-end beam status is "On", the pre-mirror mask is not closed, and the pre-mono mask is not closed.
- if "Off" when any one of the above conditions aren't met.
- turns "On" by turning on the front-end control, opening the pre-mirror mask to setpoint, and opening the pre-mono mask to setpoint (not implemented yet).
- turns "Off" by closing the pre-mirror mask (not implemented yet).
- would be turned "Off" if user wants to stop beam before it reaches any optical equipment.

The second control reports the status of the beam in the SOE, at the sample. It depends on the POE beam status and the endstation shutter.
- is "On" when the POE beam status is "On" and the endstation shutter is open.
- is "Off" when either of the above conditions aren't met.
- turns "On" by turning on the POE control and opening the endstation safety shutter.
- turns "Off" by closing the endstation safety shutter.
- would be turned "Off" if the user wants to stop beam from getting to the sample, but would like to keep the optical equipment warm.

A final main control, BioXASSideBeamStatusControl, has the front-end status, the POE status, and the SOE status as member variables, and reports the value of the SOE status. It enforces the relationships between the beam status controls, so that when one is set (eg the front-end beam status control), it's relationships with the other three are also updated (eg the POE beam status control is updated with this control).

Develop controls:
- [x] BioXASBeamStatusControl - ancestor for all beam status controls, subclass of AMPseudoMotorControl.
- [x] BioXASSSRLMonochromatorMaskControl - reports whether or not the premono mask is open.
- [x] BioXASSidePOEBeamStatusControl - reports whether beam is available in the POE.
- [x] BioXASSideSOEBeamStatusControl - reports whether beam is available in the SOE.
- [x] BioXASSideBeamStatusControl - reports the total beam status, a function of the three beam status controls.

[...]

PR #1803, opened 2015-12-17
Pull Request for Issue1630: creating beam on/off controls

helfrij:
This pull request initially tackled this task by developing a series of subcontrols handled in separate requests: #1651, #1652, #1656. The class structure developed was pretty unwieldy, so it was refactored in a separate branch (https://github.com/acquaman/acquaman/tree/valvesRefactor). The new structure is as follows:
- BioXASM1MirrorMask ~ inherits from BioXASBeamlineComponent, contains information about the M1 mirror mask: the upper blade control and the current state.

- BioXASM1MirrorMaskState ~ inherits from AMSingleEnumeratedControl, reports the current Open/Closed state of the mirror mask. This control presently supports closing only.

- BioXASSSRLMonochromatorMask ~ inherits from BioXASBeamlineComponent, contains information about the mono mask: the upper and lower blade motors and the current state.

- BioXASSSRLMonochromatorMaskState ~ inherits from AMSingleEnumeratedControl, reports the current Open/Closed state of the mask. This control presently supports closing only.

- BioXASBiStateGroup ~ inherits from AMEnumeratedControl. A general kind of bi-state control that can determine the current state or move to a new state by mapping controls to appropriate bistate values.

  - BioXASValves ~ inherits from BioXASBiStateGroup. Provides general group functionality specific to valves. Can recognize Open/Closed states, but only moves to Open state.

    - BioXAS{FrontEnd, Side, Main, Imaging, Master}Valves ~ inherits from BioXASValves, contains instances of specific valves in that particular area. Master valves has instances of all valves.

  - BioXASShutters ~ inherits from BioXASBiStateGroup. Provides general group functionality specific to shutters. Can recognize Open/Closed states, but only moves to Open state.

    - BioXASFrontEndShutters ~ inherits from BioXASShutters, contains instances of the front-end shutters.

  - BioXASBeamStatus ~ inherits from BioXASBiStateGroup. Provides general group functionality specific to beam status. Can recognize On/Off states, but is read-only by default (for now).

    - BioXASFrontEndBeamStatus ~ inherits from BioXASBeamStatus, takes information about the front end valves and shutters to determine the beam status for in the front-end.

      - BioXASSidePOEBeamStatus ~ inherits from BioXASFrontEndBeamStatus, considers front-end information as well as the mirror mask and mono mask to determine the beam status for the Side POE.

        - BioXASSideSOEBeamStatus ~ inherits from BioXASSidePOEBeamStatus, considers front-end and POE information as well as the endstation shutter to determine the beam status for the Side SOE.

          - BioXASSideBeamStatus ~ inherits from BioXASSideSOEBeamStatus, considers same information as the Side SOE status but also contains instances of the front-end, POE, and SOE controls. To identify the beam status, only the SideBeamStatus needs to be created.

The result is a collection of largely read-only controls for the front-end and BioXAS Side. Main has access to the front-end controls, but no Main-specific work has been done yet. I expect it would follow the same structure as Side pretty closely.


source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -136,24 +142,27 @@ QString AMPseudoMotorControl::toString() const
...
-
 	int childCount = childControlCount();
 
-	for (int childIndex = 0; childIndex < childCount; childIndex++) {
-		AMControl *child = childControlAt(childIndex);
+	if (childCount > 0) {
 
-		if (child) {
-			controlConnected += "\t" + child->objectName() + " connected: ";
+		controlConnected += "\n";
```
dretrex: The creation of this string could probably be done in a single QString call using args.

source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -174,7 +183,7 @@ QString AMPseudoMotorControl::toString() const
 
 	// Create and return complete info string.
 
-	QString result = controlName + "\n" + controlDescription + "\n" + controlConnected + "\n" + controlMoving + "\n" + controlCalibrating;
+	QString result = controlName + "\n" + controlDescription + "\n" + controlValue + "\n" + controlConnected + "\n" + controlMoving + "\n" + controlCalibrating;
```
dretrex: Could use string builder for this instead of `+`.

source/beamline/BioXAS/BioXASBiStateGroup.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+bool BioXASBiStateGroup::clearBiStateControls()
+{
+	bool result = false;
+
+	QList<AMControl*> children = childControls();
+
+	if (children.count() > 0) {
+		foreach (AMControl *child, childControls())
```
dretrex: Since you made `removeBiStateControl()` a boolean function, you could use that boolean here inside this loop.

source/beamline/BioXAS/BioXASBiStateGroup.cpp
```diff
@@ -0,0 +1,363 @@
...
+	QList<AMControl*> children = childControls();
+
+	if (children.count() > 0) {
+
+		bool childrenConnected = true;
+
+		for (int i = 0, count = children.count(); i < count && childrenConnected; i++) { // we want to stop if any one child isn't connected.
+			AMControl *child = children.at(i);
+
+			if (child && child->isConnected())
```
dretrex: This is a little unnecessary, the value of `childrenConnected` is always true until it isn't.  This should probably just be,

```c++
if (!(child && child->connected())
     childrenConnected = false;
```

source/beamline/BioXAS/BioXASBiStateGroup.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+AMAction3* BioXASBiStateGroup::createMoveChildrenToState1Action()
+{
+	AMListAction3 *moveChildren = new AMListAction3(new AMListActionInfo3("Moving bistate group.", "Moving bistate group."), AMListAction3::Parallel);
+
+	foreach (AMControl *child, children_) {
+		AMAction3 *action = 0;
+
+		if (child) {
```
dretrex: Since this is a BioXAS class it is probably fine, but is it valid to assume that if you can't move that you just wait for it to go to a certain value?  There is probably an assumption made somewhere that makes sense for this situation, but I haven't noticed it yet.

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
```diff
@@ -0,0 +1,97 @@
...
+{
+	AMListAction3 *actionList = new AMListAction3(new AMListActionInfo3("Opening front-end shutters", "Opening front-end shutters"), AMListAction3::Sequential);
+	actionList->addSubAction(createCheckChildIsOpen(upstreamPhotonShutter_, 10));
+	actionList->addSubAction(createMoveChildToOpen(safetyShutter_));
+	actionList->addSubAction(createMoveChildToOpen(downstreamPhotonShutter_));
+
+	return actionList;
+}
+
+//AMAction3* BioXASFrontEndShutters::createMoveToClosedAction()
```
dretrex: What is the reason for not being able to close the shutters?

source/beamline/BioXAS/BioXASMasterValves.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+BioXASMasterValves::~BioXASMasterValves()
+{
+
+}
+
+bool BioXASMasterValves::isOpen() const
+{
+	bool open = ( isConnected() && BioXASValves::isOpen() );
```
dretrex: Implemented as is, wouldn't this code already be covered by the polymorphic attributes of `BioXASValves::isOpen()` since it already asks about the connectivity?

source/beamline/BioXAS/BioXASSSRLMonochromatorMaskState.cpp
```diff
@@ -0,0 +1,100 @@
+#include "BioXASSSRLMonochromatorMaskState.h"
+#include "actions3/AMActionSupport.h"
+#include "actions3/AMListAction3.h"
+
+BioXASSSRLMonochromatorMaskState::BioXASSSRLMonochromatorMaskState(const QString &name, QObject *parent) :
+	AMSingleEnumeratedControl(name, "", parent)
```
dretrex: Since this has two controls, shouldn't this be a descendent of AMEnumeratedControl rather than AMSingleEnumeratedControl?

source/beamline/BioXAS/BioXASValves.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+AMAction3* BioXASValves::createMoveAction(double setpoint)
+{
+	AMAction3 *result = 0;
+
+	if (setpoint == Open)
+		result = createMoveToOpenAction();
+
+	// Closing valves is not supported at this time.
+	// This is because a user accidentally closing valves
```
dretrex: Really?  Could you explain the situation as I don't recall the reasons for the machine protection.

source/beamline/BioXAS/BioXASValves.cpp
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	return action;
+}
+
+int BioXASValves::currentIndex() const
+{
+	int result = indicesNamed("Unknown").first();
+
+	if (isOpen())
```
dretrex: Technically, this won't ever give you the result of "Unknown" based on the definitions of `isOpen()` and `isClosed()`.  Is that the desired behaviour?

source/ui/beamline/AMControlConnectedLEDView.cpp
```diff
@@ -0,0 +1,25 @@
...
+void AMControlConnectedLEDView::update()
+{
+	if (control_) {
+		if (control_->isConnected())
+			setAsGreenOn();
+		else
+			setAsRedOn();
+
+	} else {
+		setAsGreenOff();
```
dretrex: Only set as green off even if a red light?  There is a red and yellow off too.

dretrex:
Time Estimate: 40 minutes

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
helfrij: Right now, there's the potential for someone to accidentally close shutters at one beamline and affect what someone else might be doing on another beamline. This functionality would be easy to add, but may cause unintended problems, especially for users later on.

source/beamline/BioXAS/BioXASBiStateGroup.cpp
helfrij: Ah yes. I think I understand what you mean. This was added specifically to accommodate the front-end shutters. We don't have control over the first photon shutter, can't actually open or close it. But its status is important to the overall shutters open/closed status. This particular control is a child that can't move.

source/beamline/BioXAS/BioXASMasterValves.cpp
helfrij: Yes, you're right. Or, er, it should. Looking at BioXASValves, I don't actually see a connectivity check. I'll fix this, and check the other descendants of BioXASBiStateGroup for the same error.

source/beamline/BioXAS/BioXASBiStateGroup.cpp
helfrij: Ok. I'll make the change.

source/beamline/BioXAS/BioXASBiStateGroup.cpp
helfrij: Yup, that makes sense. Will change.

source/beamline/BioXAS/BioXASSSRLMonochromatorMaskState.cpp
helfrij: It could be, for sure. It is an AMSingleEnumeratedControl because the mono mask state began as a read-only control that mapped the blades status pv to either Open or Closed. The choice definitely made sense in this case. This control is no longer read-only, has the ability to move between Open/Closed states with the addition of the blade controls. Because this control is still essentially mapping the value of one pv, I've kept it as a single enumerated control.

source/beamline/BioXAS/BioXASValves.cpp
helfrij: If any valve is closed, on any BioXAS beamline, all beamlines cannot run.

source/beamline/BioXAS/BioXASValves.cpp
helfrij: Hmmm, I don't think so. I'll make the change. It seems this is the only class that does this (not BioXASShutters, BioXASBeamStatus).

source/ui/beamline/AMControlConnectedLEDView.cpp
helfrij: What do you suggest?

dretrex: I'd imagine if you were a red light the red off would be the one you wanted? If it looks better with the green off then by all means keep it as it is now.

Sent from my iPhone

On Dec 22, 2015, at 11:32 AM, helfrij <notifications@github.com<mailto:notifications@github.com>> wrote:


In source/ui/beamline/AMControlConnectedLEDView.cpp<https://github.com/acquaman/acquaman/pull/1803#discussion_r48276920>:

> +
> +AMControlConnectedLEDView::~AMControlConnectedLEDView()
> +{
> +
> +}
> +
> +void AMControlConnectedLEDView::update()
> +{
> +     if (control_) {
> +             if (control_->isConnected())
> +                     setAsGreenOn();
> +             else
> +                     setAsRedOn();
> +
> +     } else {
> +             setAsGreenOff();


What do you suggest?

-

helfrij:
Hey @dretrex, was there anything more you were concerned about for this pull request? Would it be okay to merge?

dretrex:
Looks good, @helfrij.  I'll merge it in.

helfrij:
Thanks, @dretrex!

PR #1813, opened 2015-12-24
Pull Request for Issue1811: Create status buttons for the shutters and valves

helfrij:
Created new view for BioXAS that mirrors the functionality of a widget in the VESPERS persistent view: a row of buttons for beamline components. Each button is colored according to the corresponding component's status. Clicking on the button causes more information about that component to be displayed.

Created:
- AMToolButton: general class that provides some convenience functions for changing button colors.
- AMControlToolButton: subclass of AMToolButton, handles setting controls and general settings.
- BioXASShuttersButton: subclass of AMControlToolButton, has shutters-specific settings (icons, etc).
- BioXASValvesButton: subclass of AMControlToolButton, has valves-specific settings (icons, etc).


source/icons/icons.qrc
```diff
@@ -215,5 +215,7 @@
         <file>22x22/go-downleft-dark.png</file>
         <file>22x22/go-upleft-dark.png</file>
         <file>22x22/go-upright-dark.png</file>
+        <file>shutterIcon2.png</file>
```
dretrex: I like these icons!

source/ui/BioXAS/BioXASBeamlineStatusBar.h
```diff
@@ -0,0 +1,71 @@
...
+
+	/// The buttons group.
+	QButtonGroup *buttonsGroup_;
+	/// The list of buttons.
+	QList<QAbstractButton*> buttonsList_;
+	/// The buttons layout.
+	QHBoxLayout *buttonsLayout_;
+
+	/// The mapping between buttons and corresponding views.
+	QList<QWidget*> buttonViewsList_;
```
dretrex: This would probably be better as a QMap.  I can tell that it works by the nature of how you add to the list, but it would probably be a bit more concise.

dretrex:
Overall, looks good.
Time Estimate: 5 minutes

source/ui/BioXAS/BioXASBeamlineStatusBar.h
helfrij: Looking back over my commits, I see that I actually started with a QMap but moved away from it at some point. Did I include why in the commit? I did not. I'll have another look over the status bar class and make the change.

helfrij:
I have a question related to styling widgets. In AMToolButton, I used QPalette to change the button colors instead of stylesheets. This choice was mostly because it seems simpler: style changes related to hovering, clicking, selecting are handled automatically. I think for stylesheets, these properties have to be accounted for manually. Is there a preferred way of styling widgets?

dretrex:
If you are mostly just wanting to change the color or background color of a widgets then using QPalette inside of the QStyle system is preferred because QStyle handles all the multiplatform aspects of painting while maintaining the same look.

Stylesheets will usually look exactly how you want, but require a fair bit more effort because not only do you need to specify the look during transitions like hovering or clicking, but also with having buttons look 3D and other aspects.

dretrex:
Looks good, @helfrij!  Is this ready for merging?

source/ui/BioXAS/BioXASBeamlineStatusBar.h
helfrij: I've replaced the button/views lists with a single QMap between the two, the status bar seems to still work as expected.

helfrij:
Yes. Thanks @dretrex.

PR #1818, opened 2016-01-05
Pull Request for Issue1817: Adding userRate.ESLO control to CLSSIS3820ScalerChannel

helfrij:
Added a new control to CLSSIS3820ScalerChannel, for the counts-to-volts conversion slope value. Added the ability to set a slope value preference to the scaler channel class--a preference because at the time this value would be specified (ie in AMBeamline::setupComponents()) the control may not be connected. The control is updated with the preference when it reports as connected. Added calls to set the preference to the scaler channels on BioXAS Side and Main.


dretrex:
Looks good.
Time Estimate: 2 minutes

PR #1819, opened 2016-01-06
Pull Request for Issue1660 alt: Adding stop capabilities to control editors on BioXAS

helfrij:
Created a new class BioXASControlEditor, a subclass of AMExtendedControlEditor. This editor provides a context menu with the available action options that can be performed with a control: move, stop, calibrate. The action options are enabled if that option is available for that particular control, disabled otherwise. Replaced instances of AMExtendedControlEditor with the new editor for all controls in the mono, m1 mirror, and m2 mirror views.

### DIFF
--- a/source/ui/BioXAS/BioXASControlEditor.cpp
+++ b/source/ui/BioXAS/BioXASControlEditor.cpp
@@ -1,0 +1,119 @@
+#include "BioXASControlEditor.h"
+#include "beamline/AMControl.h"
+
+BioXASControlEditor::BioXASControlEditor(AMControl *control, QWidget *parent) :
+	AMExtendedControlEditor(control, 0, false, false, parent)
+{
+	// Initialize class variables.
+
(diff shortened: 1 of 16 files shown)


source/ui/BioXAS/BioXASControlEditor.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void BioXASControlEditor::onContextMenuRequested(const QPoint &clickPosition)
+{
+	// Update the actions to reflect current control settings.
+
+	updateActions();
+
+	// Create context menu.
+
+	QMenu contextMenu;
```
dretrex: That's really slick.  I should probably look into making my context menus this way.  For some reason, I've always avoided using QActions.

dretrex:
Looks good.
Time Estimate: 2 minutes

helfrij:
Thanks @dretrex.

PR #1826, opened 2016-01-11
Pull Request for Issue 1824: Fix from SGM Scaler View

iainworkman:
Changed format of the SGM Scaler view to display int values instead of doubles.


iainworkman:
Looks good to me.
Time Estimate: 1 minute

PR #1835, opened 2016-01-12
Pull Request for Issue1832: Adding the QE65000 to the SGM

iainworkman:
- Added the QE65000 detector to the SGM Beamline.
- Added the detector view to the SGM App Controller.
- Fixed a reversed logic bug in CLSQE65000Detector

### DIFF
--- a/source/beamline/CLS/CLSQE65000Detector.cpp
+++ b/source/beamline/CLS/CLSQE65000Detector.cpp
@@ -158,12 +158,20 @@ bool CLSQE65000Detector::initializeImplementation(){
 }
 
 bool CLSQE65000Detector::acquireImplementation(AMDetectorDefinitions::ReadMode readMode){
-	if(!isConnected() || readMode != AMDetectorDefinitions::SingleRead)
+	if(!isConnected() || readMode != AMDetectorDefinitions::SingleRead) {
+		if(!isConnected()) {
+		}
+
+		if(readMode != AMDetectorDefinitions::SingleRead) {
+		}
 		return false;
+	}
 
 	AMControl::FailureExplanation failureExplanation = startAcquisitionControl_->move(1);
-	if(failureExplanation != AMControl::NoFailure)
+	if(failureExplanation == AMControl::NoFailure) {
 		return true;
+	}
+
 	return false;
 }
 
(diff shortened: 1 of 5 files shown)


source/beamline/CLS/CLSQE65000Detector.cpp
```diff
@@ -158,12 +158,18 @@ bool CLSQE65000Detector::initializeImplementation(){
 }
 
 bool CLSQE65000Detector::acquireImplementation(AMDetectorDefinitions::ReadMode readMode){
-	if(!isConnected() || readMode != AMDetectorDefinitions::SingleRead)
+	if(!isConnected() || readMode != AMDetectorDefinitions::SingleRead) {
+
+		if(readMode != AMDetectorDefinitions::SingleRead) {
```
dretrex: Is that return value supposed to be inside those curly braces?

source/beamline/CLS/CLSQE65000Detector.cpp
```diff
@@ -158,12 +158,18 @@ bool CLSQE65000Detector::initializeImplementation(){
...
+		}
 		return false;
+	}
 
 	AMControl::FailureExplanation failureExplanation = startAcquisitionControl_->move(1);
-	if(failureExplanation != AMControl::NoFailure)
+	if(failureExplanation == AMControl::NoFailure) {
```
dretrex: So was this broken the whole time?  This seems like a pretty big deal to switch from `!=` to `==`.

dretrex:
Couple of comments.
Time Estimate: 2 minutes

source/beamline/CLS/CLSQE65000Detector.cpp
iainworkman: Nope, I added that conditional in order to put a debug in there to show the reason why this call was returning false. When I cleaned up the code ready for submission I got rid of all the debugs, but forgot to remove this conditional block.

source/beamline/CLS/CLSQE65000Detector.cpp
iainworkman: Until 3e31c42c2fee83d181a425dc770faf9f9a9a7db3 the AMDetectorTriggerAction didn't pay attention to the return result of the call to the detector's acquire() function. The action succeeded either way. So while this function was returning the wrong value nothing was paying attention to it.

dretrex: Ah.  Well, that's a bit troubling.  Good catch then.

PR #1837, opened 2016-01-12
Pull Request for Issue 1828: Added current undulator harmonic to trajectory calculator

iainworkman:
Changed the Energy Trajectory class so that it gets passed the Undulator Harmonic directly.

### DIFF
--- a/source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
+++ b/source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
@@ -310,7 +310,8 @@ void SGMEnergyTrajectoryTestView::setEnergyPlotData(SGMGratingSupport::GratingTr
 		                               -1.53,
 		                               1000,
 		                               -5598,
-		                               25.07);
+		                               25.07,
+		                               SGMUndulatorSupport::FirstHarmonic);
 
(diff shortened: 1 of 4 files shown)


source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
```diff
@@ -366,7 +367,8 @@ void SGMEnergyTrajectoryTestView::setTrajectoryPlotData(SGMGratingSupport::Grati
 		                               -1.53,
 		                               1000,
 		                               -5598,
-		                               25.07);
+		                               25.07,
+		                               SGMUndulatorSupport::FirstHarmonic);
```
dretrex: I know this is just in your tests code.  But is this second one also supposed to be the first harmonic or should it be the third?

dretrex:
Time Estimate: 1 minute

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
iainworkman: I created this view initially to give me a visual feedback as to how the various optimization calculations vary as a function of energy. Really this view isn't super useful any more, but I kept it in, in order to act as simple check. In reality they should be expanded to show plots for both harmonics, but this didn't seem like a huge priority.

PR #1838, opened 2016-01-12
Pull Request for Issue1836: Add BioXAS-specific actions to change between 'Continuous' and 'Single shot' modes

helfrij:
- Added two new virtual methods to CLSSIS3820Scaler that create and return new actions to move the scaler to 'Continuous' and 'Single shot' modes.
  - createMoveToContinuousAction()
  - createMoveToSingleShotAction()
- Modified the CLSSIS3820Scaler so that these methods are used in place of any previous means of moving between modes.
- Added reimplementation of these methods to BIoXASSIS3820Scaler that take the BioXAS-specific requirements into account.


dretrex:
Looks good.
Time Estimate: 1 minute

PR #1846, opened 2016-01-14
Pull Request for Issue1845: Make SGM Default to Using Local Storage

iainworkman:
Added setDefaultUseLocalStorage(true) to the SGMAppController constructor.

Had to play around with adding a AcquamanLocalData directory to my machine, as I couldn't figure out why the option still wasn't appearing. :/


dretrex:
Looks good.
Time Estimate: 1 minute

PR #1849, opened 2016-01-14
Pull Request for Issue1844: Ge detector regions of interest not set with XAS scans.

helfrij:
The problem ended up being with a misplaced connect statement: user configurations were connected to after they would have emitted their loadedFromDb() signals. This issue exposed another problem in BioXASXASScanActionController where the Ge detector's normalization ABs were being passed invalid data sources. This problem was fixed in this issue as well.


dretrex:
Looks good.
Time Estimate: 1 minute

PR #1855, opened 2016-01-15
Pull Request for Issue1851: Add missing Q_PROPERTY calls to AMActionInfo3 Classes

iainworkman:
Added Q_PROPERTY calls to all AMActionInfo3 classes for which they were missing.
In some cases this required adding the necessary setters for the WRITE argument.
In cases where the existing dbRead...() or dbLoad...() functions were public, I made them protected.

### DIFF
--- a/source/actions3/actions/AMAxisFinishedActionInfo.cpp
+++ b/source/actions3/actions/AMAxisFinishedActionInfo.cpp
@@ -47,4 +47,5 @@ AMActionInfo3 *AMAxisFinishedActionInfo::createCopy() const
 void AMAxisFinishedActionInfo::setAxisName(const QString &axisName)
 {
 	axisName_ = axisName;
+	setModified(true);
 }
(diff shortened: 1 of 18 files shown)


source/actions3/actions/AMAxisFinishedActionInfo.cpp
```diff
@@ -43,3 +43,8 @@ AMActionInfo3 *AMAxisFinishedActionInfo::createCopy() const
 	info->dissociateFromDb(true);
 	return info;
 }
+
+void AMAxisFinishedActionInfo::setAxisName(const QString &axisName)
+{
+	axisName_ = axisName;
```
dretrex: This probably needs `setModified(true);` added wouldn't it?

dretrex:
I believe you need to add `setModified(true);` to all the setters that aren't `dbLoadXYZ()`.  Otherwise it looks good.

dretrex:
Time Estimate: 3 minutes

source/actions3/actions/AMAxisFinishedActionInfo.cpp
iainworkman: Ah yes. There will be a few more like this for added setters. I will go
through and add them all.

PR #1862, opened 2016-01-18
Pull Request for Issue1861: Fast shutter doesn't turn on in scan initialization actions for XAS

dretrex:
Pretty straight forward.  It needed to check the shutter state and open it if it is closed.


iainworkman:
Looks good to me.
Time Estimate: 5 minutes

Issue #1457, opened 2015-06-22
Integrate continuous mode inside of AM rather than using the continuous PV


dretrex:
Although known for a long time, it was made painfully clear on VESPERS that switching in and out of continuous seems to have a lot of issues with the scaler.  It should be implemented inside of AM and be an actionable request.  Necessary pieces for single shot and continuous mode are the following:

#### Continuous
- scan count = 0
- number of scans = 1
- start scan = 1

#### Single Shot
- scan count = what it was before switching to continuous (this is typically 1, but not guaranteed)
- number of scans = what it was before switching to continuous (also typically 1, but not guaranteed)
- start scan = 0

The "scanning" state and whether to turn it on or not may be part of the problem. This part may need a little investigation.

helfrij:
Time estimate: 7 hours

PR #1863, opened 2016-01-18
Pull Request for Issue1860: Refactor scaler continuous mode control

helfrij:
A couple of small changes to the scaler and scaler continuous mode control classes. The biggest change was to switch the scaler continuous mode control ancestor from AMPseudoMotorControl to AMEnumeratedControl.

### DIFF
--- a/source/beamline/CLS/CLSSIS3820ScalerContinuousMode.cpp
+++ b/source/beamline/CLS/CLSSIS3820ScalerContinuousMode.cpp
@@ -1,9 +1,9 @@
-#include "CLSSIS3820ScalerModeControl.h"
+#include "CLSSIS3820ScalerContinuousMode.h"
 #include "beamline/CLS/CLSSIS3820Scaler.h"
 #include "actions3/AMActionSupport.h"
 #include "actions3/AMListAction3.h"
 
-CLSSIS3820ScalerModeControl::CLSSIS3820ScalerModeControl(const QString &name, QObject *parent) :
+CLSSIS3820ScalerContinuousMode::CLSSIS3820ScalerContinuousMode(const QString &name, QObject *parent) :
 	AMEnumeratedControl(name, "", parent)
 {
 	// Initialize inherited variables.
@@ -26,12 +26,12 @@ CLSSIS3820ScalerModeControl::CLSSIS3820ScalerModeControl(const QString &name, QO
 	addOption(CLSSIS3820Scaler::Continuous, "Continuous");
 }
 
-CLSSIS3820ScalerModeControl::~CLSSIS3820ScalerModeControl()
+CLSSIS3820ScalerContinuousMode::~CLSSIS3820ScalerContinuousMode()
 {
 
 }
 
-bool CLSSIS3820ScalerModeControl::canMeasure() const
+bool CLSSIS3820ScalerContinuousMode::canMeasure() const
 {
 	bool result = false;
 
@@ -41,7 +41,7 @@ bool CLSSIS3820ScalerModeControl::canMeasure() const
 	return result;
 }
 
-bool CLSSIS3820ScalerModeControl::canMove() const
+bool CLSSIS3820ScalerContinuousMode::canMove() const
 {
 	bool result = false;
 
@@ -51,7 +51,7 @@ bool CLSSIS3820ScalerModeControl::canMove() const
 	return result;
 }
 
-bool CLSSIS3820ScalerModeControl::isContinuous() const
+bool CLSSIS3820ScalerContinuousMode::isContinuous() const
 {
 	bool result = false;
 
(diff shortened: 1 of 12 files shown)


helfrij:
Hey @iainworkman, would you look over this pull request?

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.cpp
```diff
@@ -0,0 +1,192 @@
...
+	bool result = false;
+
+	if (isConnected()) {
+		if (scanCountControl_->value() > 0)
+			result = true;
+	}
+
+	return result;
+}
+
```
iainworkman: What happens in the above cases if the scanCountControl_->value == 0 and the scartScanControl_->value() != Scanning? I take it then we're just considered to be in an Unknown state? Is this okay?

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.h
```diff
@@ -1,31 +1,25 @@
...
+#ifndef CLSSIS3820SCALERCONTINUOUSMODE_H
+#define CLSSIS3820SCALERCONTINUOUSMODE_H
 
-#include "beamline/AMPseudoMotorControl.h"
+#include "beamline/AMEnumeratedControl.h"
 
-class CLSSIS3820ScalerModeControl : public AMPseudoMotorControl
+class CLSSIS3820Scaler;
+
+class CLSSIS3820ScalerContinuousMode : public AMEnumeratedControl
```
iainworkman: This name seems a little less obviously descriptive as to the function of the class than the old one.

iainworkman:
Looks good to me. 
Had a couple of questions, relating to the underlying way the scaler expresses that it is in continuous mode and naming of a class.
Time Estimate: 30 minutes

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.cpp
helfrij: I think you're right, though I'll check. The checks for whether the scaler is in continuous or in single shot mode could be made stronger by removing the 'scanning' condition from isContinuous(). If 'Unknown' states are a concern this could be an option. What do you think @dretrex?

The original specifications for each of the modes can be found here (#1457), for reference.

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.h
helfrij: There are at least two different 'mode groups' for the scaler: Continuous/SingleShot, InputMode. The purpose of renaming this class was to differentiate between the two. If you think the old one was fine, I can change it back.

iainworkman: No, that's fine. That seems like a good reason to make the change of name.

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.cpp
dretrex: That is definitely a good point.  Since we have very rigid definitions of what it means to be in single shot or continuous, we may as well have an Unknown state for when that is not the case.

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.h
dretrex: I think it would probably be a better idea to go with "ScanningMode" to distinguish between mode and "Input Mode" since Continuous is one of the possible ScanningMode states.

dretrex: And now that I've read through the pull request myself, there is also a "Scan Mode".  My suggestion is "Acquisition Mode" since it distinguishes it from Scan and Input mode sufficiently.

dretrex:
Time Estimate: 5 minutes

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.h
helfrij: AcquisitionMode works for me! I'll make the change, for both the class name (CLSSIS3820ScalerContinuousMode -> AcquisitionMode) and the enum name.

source/beamline/CLS/CLSSIS3820ScalerContinuousMode.cpp
helfrij: 'Unknown' is an option that's added by default in AMEnumeratedControl. I think for the purposes of this pull request, there is no change necessary here.

dretrex:
Nice.  I like how this ended up. I'll merge this in now.

helfrij:
I think I've made the necessary changes. Thanks @iainworkman @dretrex!

Time Estimate: 10 minutes.

PR #1865, opened 2016-01-18
Pull Request for Issue1834: Add differentiation between read-only and move enum values in AMEnumeratedControl

helfrij:
AMEnumeratedControl and subclasses use QMaps to map indices to particular index properties. I tackled this issue by adding a new QMap to AMEnumeratedControl that maps a given enum index to its read-only state. I modified the methods that add enum options to AMEnumeratedControls (and descendants) to include an optional argument that specifies whether a given index is read-only. I then made changes to the BioXASControlEditors so that they display only move indices as move destination options.


helfrij:

source/beamline/AMSingleEnumeratedControl.cpp
```diff
@@ -108,9 +111,14 @@ bool AMSingleEnumeratedControl::addValueOption(int index, const QString &optionS
...
-bool AMSingleEnumeratedControl::addValueOption(int index, const QString &optionString, double optionSetpoint)
+bool AMSingleEnumeratedControl::addValueOption(int index, const QString &optionString, double optionMin, double optionMax)
+{
+	return addValueOption(index, optionString, -1, optionMin, optionMax, true);
+}
+
+bool AMSingleEnumeratedControl::addValueOption(int index, const QString &optionString, double optionSetpoint, bool readOnly)
 {
-	return addValueOption(index, optionString, optionSetpoint, optionSetpoint, optionSetpoint);
+	return addValueOption(index, optionString, optionSetpoint, optionSetpoint, optionSetpoint, readOnly);
```
XuetaoLiu-CLS: Just want to make sure that for this overloaded `addValueOption()`, do we really want the `optionMin` == `optionMax` == `optionSetpoint`?

source/beamline/BioXAS/BioXASM1MirrorMaskState.cpp
```diff
@@ -11,8 +11,8 @@ BioXASM1MirrorMaskState::BioXASM1MirrorMaskState(const QString &name, QObject *p
 
 	// Setup basic value options.
 
-	addValueOption(Open, "Open", 0, 0.01, 10); // these are 'pretend' values, for now. should be safe to use them, as this control does not support moving.
-	addValueOption(Closed, "Closed", 0, 0, 0.01);
+	addValueOption(Closed, "Closed", 0.0, 0.0, 0.01);
+	addValueOption(Open, "Open", 0.01, 10.0);
```
XuetaoLiu-CLS: With this call, the default optionSetpoint for the `Open` option is "-1". But the original codes the value of optionSetpoint is "0". Is this "-1" correct?

source/beamline/BioXAS/BioXASSSRLMonochromatorMaskState.cpp
```diff
@@ -12,8 +12,8 @@ BioXASSSRLMonochromatorMaskState::BioXASSSRLMonochromatorMaskState(const QString
 
 	// Setup value options.
 
-	addValueOption(Open, "Open", Open);
-	addValueOption(Closed, "Closed", Closed);
+	addValueOption(Closed, "Closed", 1);
+	addValueOption(Open, "Open", 0, true);
```
XuetaoLiu-CLS: In the `BioXASSSRLMonochromatorMaskState.h`, the value of `Open` is `1`. Should this setpoint be 1?

helfrij: No, because the blades state pv control uses 1 to indicate Closed and 0 to indicate Open. Closed for the mono state must be 0 right now, because the move enums must be listed before read-only enums in AMControl.

XuetaoLiu-CLS: Thanks. :+1:

XuetaoLiu-CLS:
Looks great! 
Time estimate: 1 hour

XuetaoLiu-CLS:
@dretrex u can have a proof review now. Thx :100:

source/beamline/BioXAS/BioXASM1MirrorMaskState.cpp
helfrij: Hmmm. I'm not sure I see what you mean. Where is the -1 value set?

source/beamline/AMSingleEnumeratedControl.cpp
helfrij: Yup! It's for the case where a single value is mapped, not a value range.

source/ui/BioXAS/BioXASControlEditor.h
```diff
@@ -63,7 +67,7 @@ public slots:
 	/// Sets whether the control's name is used for the title text.
 	void setUseControlNameAsTitle(bool useName);
 	/// Sets the value.
-	virtual void setValue(double newValue);
+	virtual void setValue(AMNumber newValue);
```
dretrex: Is there a particular reason for using AMNumber instead of double?  If you are going to use AMNumber it should be `const AMNumber &`.

source/ui/BioXAS/BioXASValueEditor.h
```diff
@@ -25,7 +25,7 @@ class BioXASValueEditor : public QGroupBox
 	/// Returns the title text.
 	QString title() const { return title_; }
 	/// Returns the value.
-	double value() const { return double(value_); }
+	AMNumber value() const { return double(value_); }
```
dretrex: Probably don't need to cast the value to a double?

source/ui/BioXAS/BioXASValueEditor.h
```diff
@@ -45,7 +47,7 @@ class BioXASValueEditor : public QGroupBox
 	/// Notifier that the title text has changed.
 	void titleChanged(const QString &newText);
 	/// Notifier that the value has changed.
-	void valueChanged(double newValue);
+	void valueChanged(AMNumber newValue);
```
dretrex: Same AMNumber comment as before.

source/ui/BioXAS/BioXASValueEditor.h
```diff
@@ -68,7 +72,7 @@ public slots:
 	/// Sets the editor title text.
 	virtual void setTitle(const QString &newText);
 	/// Sets the value.
-	virtual void setValue(double newValue);
+	virtual void setValue(AMNumber newValue);
```
dretrex: Here too.

dretrex:
Only a couple nit picky comments.
Time Estimate: 7 minutes

source/beamline/AMEnumeratedControl.h
```diff
@@ -11,6 +11,9 @@ class AMEnumeratedControl : public AMPseudoMotorControl
     Q_OBJECT
 
 public:
+	/// Enumeration of some default option indices.
+	enum DefaultOptions { Unknown = -1, Invalid = -2 };
```
dretrex: Maybe they are used elsewhere, but I didn't see any use for these enums.  Are they used anywhere?

source/ui/BioXAS/BioXASControlEditor.h
helfrij: I use AMNumbers here because in theory it's possible for the editor value to be invalid. I'm mostly thinking of the cases where no control is set, or when a control has been set but isn't measurable.

source/ui/BioXAS/BioXASValueEditor.h
helfrij: Ha! You're right. I'll change this.

source/beamline/AMEnumeratedControl.h
helfrij: Nope, they aren't used anywhere yet. I tried to implement them early in this pull request, but I'm not ready for them yet. The plan is to use them eventually. Should I take them out?

dretrex: If there is a plan for them then I don't see a reason to remove them.  Only if they aren't needed or desired.

helfrij:
Thanks @XuetaoLiu-CLS @dretrex!

PR #1866, opened 2016-01-19
Pull Request for Issue1857: Dynamic time units in Zebra pulse controls

helfrij:
Added two new member variables to BioXASZebraPulseControl: delayTime_ and pulseTime_. These variables contain the delay value and pulse width in seconds. Added getters and setters.

The delayTime_ and pulseTime_ values are updated when the 'delay before' control and the 'pulse width' control values change.

The control values are updated when the units control reports a value change. Part of this update is to determine whether the control values are valid with respect to the new units, and to modify the value (and potentially units) to correspond to an equivalent valid value/units combination.


helfrij:
Hey @iainworkman, would you mind looking over this pull request?

source/beamline/BioXAS/BioXASZebraPulseControl.cpp
```diff
@@ -181,38 +206,41 @@ void BioXASZebraPulseControl::setEdgeTriggerValue(int value)
...
-void BioXASZebraPulseControl::setTimeUnitsValue(int value)
+void BioXASZebraPulseControl::setPulseTime(double newTime)
 {
-	if (!timeUnitsControl_->withinTolerance(double(value)))
-		timeUnitsControl_->move(double(value));
+	if (pulseTime_ != newTime) {
+		pulseTime_ = newTime;
+		updatePulseWidthValue();
+
+		emit pulseWidthValueChanged(pulseTime_);
```
helfrij: Perhaps there should be different signals for when these time values change? 'pulseWidthValueChanged' seems to imply that the pulse width control's value has changed, when that's not the event this signal is describing.

dretrex: Good point!  Probably have the current signals for all the controls and separate ones where have our own abstraction.

source/beamline/BioXAS/BioXASZebraPulseControl.cpp
```diff
@@ -167,6 +152,46 @@ bool BioXASZebraPulseControl::outputValue() const
...
 }
 
+bool BioXASZebraPulseControl::validTimeValue(double timeValue) const
+{
+	bool result = false;
+
+	if (timeValue >= BIOXASZEBRAPULSECONTROL_PULSE_WIDTH_MIN && timeValue <= BIOXASZEBRAPULSECONTROL_PULSE_WIDTH_MAX)
+		result = true;
+
+	return result;
```
iainworkman: This function could be compressed to a single line with
```c++
return timeValue >= BIOXASZEBRAPULSECONTROL_PULSE_WIDTH_MIN && timeValue <= BIOXASZEBRAPULSECONTROL_PULSE_WIDTH_MAX;
```

source/beamline/BioXAS/BioXASZebraPulseControl.cpp
```diff
@@ -167,6 +152,46 @@ bool BioXASZebraPulseControl::outputValue() const
...
+}
+
+double BioXASZebraPulseControl::convertTimeValue(double timeValue, double timeUnits, double desiredTimeUnits) const
```
iainworkman: Is there a reason why the time units are passed around this class as doubles, rather than the enum or integers?

iainworkman:
A couple of comments, but everything else looks good.
Time Estimate: 15 minutes.

source/beamline/BioXAS/BioXASZebraPulseControl.cpp
helfrij: This may be a convenience thing. The timeUnitsControl_ will always report its value as a double. Arguments that are doubles mean the function will accept arguments that are in double, int, or enum form--it's the most flexible.

source/beamline/BioXAS/BioXASZebraPulseControl.cpp
helfrij: This version saves space, but I think is a little less readable.

PR #1868, opened 2016-01-19
Pull Request for Issue1867: Advanced check box in AMChooseFolderDialog doesn't allow you to check it if the line edit isn't empty

dretrex:
I have updated the choose folder dialog and removed the advanced check because it is no longer necessary.  I believe that the messages and the disabled state of the okay button should be sufficient for proper usage.


iainworkman:
Can the advancedCheckbox_ be removed from the .h file, now that it is not being used?

dretrex:
Yup.

iainworkman:
Other than the one comment everything looks good.
Time Estimate: 5 minutes.

PR #1871, opened 2016-01-19
Pull Request for Issue717: Implement Generic File Repair for CDF Files

dretrex:
I removed the code rather than implementing a generic repair like the original issue stated because it isn't actually necessary.


iainworkman:
Looks good to me.
Time Estimate: 5 minutes

PR #1875, opened 2016-01-20
Pull Request for Issue1644: Big beam stages are not propagated into XAS scan configurations

dretrex:
Added the Big Beam motors to the XAS _Go To Position_ which only required updating the app controller, the configuration view, and the scan controller.

### DIFF
--- a/source/acquaman/VESPERS/VESPERSXASScanActionController.cpp
+++ b/source/acquaman/VESPERS/VESPERSXASScanActionController.cpp
@@ -218,6 +218,12 @@ AMAction3* VESPERSXASScanActionController::createInitializationActions()
 			initializationAction->addSubAction(VESPERSBeamline::vespers()->realSampleStageMotorGroupObject()->horizontalAxis()->createTranslateMoveAction(configuration_->x()));
 			initializationAction->addSubAction(VESPERSBeamline::vespers()->realSampleStageMotorGroupObject()->verticalAxis()->createTranslateMoveAction(configuration_->y()));
 		}
+
+		else if (motor.testFlag(VESPERS::BigBeamX) && motor.testFlag(VESPERS::BigBeamZ)){
+
+			initializationAction->addSubAction(VESPERSBeamline::vespers()->bigBeamMotorGroupObject()->horizontalAxis()->createTranslateMoveAction(configuration_->x()));
+			initializationAction->addSubAction(VESPERSBeamline::vespers()->bigBeamMotorGroupObject()->verticalAxis()->createTranslateMoveAction(configuration_->y()));
+		}
 	}
 
 	return initializationAction;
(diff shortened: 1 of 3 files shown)


dretrex:
Hey @helfrij, I have tested this and is now ready for review.  Could you look at this for me?

source/acquaman/VESPERS/VESPERSXASScanActionController.cpp
```diff
@@ -218,6 +218,12 @@ AMAction3* VESPERSXASScanActionController::createInitializationActions()
 			initializationAction->addSubAction(VESPERSBeamline::vespers()->realSampleStageMotorGroupObject()->horizontalAxis()->createTranslateMoveAction(configuration_->x()));
 			initializationAction->addSubAction(VESPERSBeamline::vespers()->realSampleStageMotorGroupObject()->verticalAxis()->createTranslateMoveAction(configuration_->y()));
 		}
+
+		else if (motor.testFlag(VESPERS::BigBeamX) && motor.testFlag(VESPERS::BigBeamZ)){
+
+			initializationAction->addSubAction(VESPERSBeamline::vespers()->bigBeamMotorGroupObject()->horizontalAxis()->createTranslateMoveAction(configuration_->x()));
+			initializationAction->addSubAction(VESPERSBeamline::vespers()->bigBeamMotorGroupObject()->verticalAxis()->createTranslateMoveAction(configuration_->y()));
```
helfrij: The 'else if' checks for an x and a z coordinate/state, but the action initializes an x and a y axis. Just checking--is this right?

source/ui/VESPERS/VESPERSEXAFSScanConfigurationView.cpp
```diff
@@ -430,6 +430,14 @@ void VESPERSEXAFSScanConfigurationView::setScanPosition()
 		savedYPosition_->setText(QString("Z: %1 mm").arg(y, 0, 'g', 3));
 	}
 
+	else if (motor == (VESPERS::BigBeamX | VESPERS::BigBeamZ)){
+
+		x = VESPERSBeamline::vespers()->bigBeamMotorGroupObject()->horizontalAxis()->translationMotor()->value();
+		y = VESPERSBeamline::vespers()->bigBeamMotorGroupObject()->verticalAxis()->translationMotor()->value();
+		savedXPosition_->setText(QString("X: %1 mm").arg(x, 0, 'g', 3));
+		savedYPosition_->setText(QString("Z: %1 mm").arg(y, 0, 'g', 3));
```
helfrij: Here again the y/z mismatch. Just a quirk of using different coordinate systems? Or doesn't mean what I think it means?

helfrij:
Had a quick question about coordinate systems. Other than that, looks good.

Time Estimate: 10 minutes.

source/acquaman/VESPERS/VESPERSXASScanActionController.cpp
dretrex: Yeah, it's right.  The "Go To Position" aspect to the XAS scan is VERY old code and uses X and Y.  VESPERS uses the general convention of X and Z for horizontal and vertical motion because Y is reserved for the beam direction.

source/ui/VESPERS/VESPERSEXAFSScanConfigurationView.cpp
dretrex: Bad, old code, quirk.

source/acquaman/VESPERS/VESPERSXASScanActionController.cpp
helfrij: Ah okay. Makes sense.

PR #1876, opened 2016-01-20
Pull Request for Issue1864: Add storage space monitor to dataman applications

iainworkman:
- Created a class, AMStorageInfo, which takes a lot of code from Qt5.4's QStorageInfo and provides space information on a given volume.
- Added a storage info class to AMDatamanAppController if local storage is being used.
- Created a timer mechanism in AMDatamanAppController which checks storage usage on a minute frequency when local storage is being used
- Added a progress bar to AMDatamanAppController to visualize storage usage when local storage is being used.
- Added an error mon call when the storage usage gets above 85% (suppressed after being shown once, so the user isn't spammed every minute with error mons).

### DIFF
--- a/acquamanCommon.pri
+++ b/acquamanCommon.pri
@@ -441,7 +441,8 @@ HEADERS += \
 	source/beamline/AMExclusiveStatesEnumeratedControl.h \
 	source/ui/beamline/AMControlStopButton.h \
     source/ui/beamline/AMControlToolButton.h \
-    $$PWD/source/ui/AMToolButton.h
+    $$PWD/source/ui/AMToolButton.h \
+    source/util/AMStorageInfo.h
 
 FORMS += \
 
@@ -844,7 +845,8 @@ SOURCES += \
 	source/beamline/AMExclusiveStatesEnumeratedControl.cpp \
     source/ui/beamline/AMControlStopButton.cpp \
     source/ui/beamline/AMControlToolButton.cpp \
-    $$PWD/source/ui/AMToolButton.cpp
+    $$PWD/source/ui/AMToolButton.cpp \
+    source/util/AMStorageInfo.cpp
 
 RESOURCES *= source/icons/icons.qrc \
 		source/configurationFiles/configurationFiles.qrc \
@@ -861,3 +863,5 @@ contains(DEFINES, AM_BUILD_REPORTER_ENABLED){
 
 	SOURCES *= source/util/AMRunTimeBuildInfo.cpp
 }
+
+
--- a/source/application/AMDatamanAppController.cpp
+++ b/source/application/AMDatamanAppController.cpp
@@ -1973,7 +1973,8 @@ void AMDatamanAppController::updateStorageProgressBar()
 				storageWarningLabel_->setVisible(true);
 			}
 
-			if(storageWarningCount_ % 30 == 0) {
+			// Only show the warning every 60 minutes
+			if(storageWarningCount_ % 60 == 0) {
 				double percentageRemaining = 100 - percentageUsed;
 				AMErrorMon::alert(this,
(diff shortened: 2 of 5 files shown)


iainworkman:
@XuetaoLiu-CLS Would you mind taking a look at this one?
Thanks.

acquamanCommon.pri
```diff
@@ -844,7 +845,8 @@ SOURCES += \
 	source/beamline/AMExclusiveStatesEnumeratedControl.cpp \
     source/ui/beamline/AMControlStopButton.cpp \
     source/ui/beamline/AMControlToolButton.cpp \
-    $$PWD/source/ui/AMToolButton.cpp
+    $$PWD/source/ui/AMToolButton.cpp \
```
XuetaoLiu-CLS: just out of curious, why do we need `$$PWD/` for some files?

dretrex: That is something from newer Qt Creators - maybe even only on Macs.

iainworkman: I believe this is a factor of different Qt versions. 
When I add a class it appears in the project file in the format
```
source/blah/thing
```
whereas it seems as if when either Jessie or Darren add it it prepends the $$PWD onto the front. I have seen no observable difference between both. If the $$PWD prefix is removed everything still works fine.

source/application/AMDatamanAppController.cpp
```diff
@@ -1875,3 +1932,33 @@ void AMDatamanAppController::onOpenOtherDatabaseClicked()
...
+
+		storageInfo_.refresh();
+		double storageUsed = double(storageInfo_.bytesTotal() - storageInfo_.bytesAvailable());
+		double percentageUsed = (storageUsed / storageInfo_.bytesTotal()) * 100;
+
+		if(percentageUsed > 85) {
+
+			if(!storageWarningProvided_) {
+				double percentageRemaining = 100 - percentageUsed;
+				AMErrorMon::alert(this, AMDATAMANAPPCONTROLLER_LOCAL_STORAGE_RUNNING_LOW, QString("Warning: Local storage space is at %1%. Please inform the beamline staff.").arg(percentageRemaining));
```
XuetaoLiu-CLS: Do we want a visual timeout message box for a visual alert?
```
	static void alert(const QObject *src = 0, int code = 0, const QString &desc = "", bool showTimeoutMsgBox=false, QString msgBoxTitle="Alert")
```

source/application/AMDatamanAppController.cpp
```diff
@@ -1875,3 +1932,33 @@ void AMDatamanAppController::onOpenOtherDatabaseClicked()
...
+				storageWarningProvided_ = true;
```
XuetaoLiu-CLS: How will we clear the warning provided flag? By far, it seems that the only way to clear `storageWarningProvided_` is via:
 -- close acquaman
 -- remove/delete files on local storage to gain free space ? 10%
 -- restart acquaman

source/application/AMDatamanAppController.cpp
iainworkman: That might be a good idea.

XuetaoLiu-CLS:
Look great! Nice done :8ball: 

Time Estimate: 20 minutes

dretrex:
Looks good to me too.  I'll build it on my mac first and if that works I'll wait for the comments from @XuetaoLiu-CLS to be addressed before merging.
Time Estimate: 5 minutes

iainworkman:
Made those updates:
- The suppression check for the error mon is now based on an incremented integer. Instead of only displaying once the error mon is configured to only display every 30 checks where the space is found to be consecutively running low (30 minutes).
- Added a warning label which appears when space is running low.

iainworkman:
@XuetaoLiu-CLS @dretrex You can have a look at this again now, if required.

dretrex:
Also, I built it on my mac and it didn't build, but you've already addressed it with the ifdef with the includes.

XuetaoLiu-CLS:
It looks good to me with the changes.

dretrex:
I'm good too. I'll merge it in.  Good work everyone.

PR #1880, opened 2016-01-20
Pull Request for Issue1878: EXAFS regions don't produce the right time when the time and maximum time are equal

dretrex:
This was mostly a simple logic comparison bug, but I also made the VESPERS XAS code easier to read.


helfrij:
Looks good, @dretrex!

Time Estimate: 5 minutes.

PR #1881, opened 2016-01-21
Pull Request for Issue1872: Implementing AND and OR blocks for Zebra

helfrij:


helfrij:
Hey @XuetaoLiu-CLS, would you look at this pull request?

source/beamline/BioXAS/BioXASZebraInput.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+	return int(valueControl_->value());
+}
+
+QString BioXASZebraInput::inputValueString() const
+{
+	return BioXASZebraCommand::nameFromCommand(inputValue());
+}
+
+bool BioXASZebraInput::inputStatus() const
```
XuetaoLiu-CLS: This looks weird to me. When I see "xxx status", I'm expecting something like state instead of boolean value. Recommend to have a function mapping with the "1" status.

source/beamline/BioXAS/BioXASZebraLogicBlock.cpp
```diff
@@ -0,0 +1,69 @@
...
+	outputStatusControl_ = new AMReadOnlyPVControl(
+				QString("LogicBlockOutput"),
+				QString("%1_OUT").arg(baseName),
+				this);
+
+	allControls_->addControl(outputStatusControl_);
+
+	// Make connections.
+
+	connect( allControls_, SIGNAL(connected(bool)), this, SLOT(updateConnected()) );
```
XuetaoLiu-CLS: do we need to handle the `connected()` signal of `outputStatusControl_` as well?

source/beamline/BioXAS/BioXASZebraLogicBlock.cpp
```diff
@@ -0,0 +1,69 @@
...
+	connect( outputStatusControl_, SIGNAL(valueChanged(double)), this, SLOT(onOutputStatusValueChanged()) );
+}
+
+BioXASZebraLogicBlock::~BioXASZebraLogicBlock()
+{
+
+}
+
+bool BioXASZebraLogicBlock::outputStatus() const
```
XuetaoLiu-CLS: same question as `BioXASZebraInput`. Maybe an enum return result with (ON, OFF) value or rename function as `isOutputOn()` ?

source/beamline/BioXAS/BioXASZebraLogicBlock.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+	if (connected_ != isConnected) {
+		connected_ = isConnected;
+		emit connected(connected_);
+	}
+}
+
+void BioXASZebraLogicBlock::updateConnected()
+{
+	setConnected(allControls_->isConnected());
```
XuetaoLiu-CLS: as mentioned above, we might need to chek `outputStatusControl_`  connectivity as well

source/beamline/BioXAS/BioXASZebraLogicBlock.cpp
helfrij: Because outputStatusControl_ is added to the allControls_ control set, I think it's connect status is taken care of.

source/beamline/BioXAS/BioXASZebraInput.cpp
helfrij: That's a good idea! What the result of a function with this name really means is a little ambiguous.

source/beamline/BioXAS/BioXASZebraLogicBlockInput.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+
+}
+
+bool BioXASZebraLogicBlockInput::enabledStatus() const
+{
+	return (int(enabledStatusControl_->value()) == Enabled);
+}
+
+bool BioXASZebraLogicBlockInput::invertedStatus() const
```
XuetaoLiu-CLS: same here..

source/ui/BioXAS/BioXASZebraLogicBlockInputInvertedView.cpp
```diff
@@ -0,0 +1,77 @@
+#include "BioXASZebraLogicBlockInputInvertedView.h"
+
+BioXASZebraLogicBlockInputInvertedView::BioXASZebraLogicBlockInputInvertedView(BioXASZebraLogicBlockInput *control, QWidget *parent) :
```
XuetaoLiu-CLS: This class is so alike with `BioXASZebraLogicBlockInputEnabledView`. I think we should generalize it.

source/beamline/BioXAS/BioXASZebraLogicBlock.cpp
XuetaoLiu-CLS: yeah... missed that part :+1:

XuetaoLiu-CLS:
Look good.
Time Estimate: 20 minutes

source/beamline/BioXAS/BioXASZebraLogicBlock.cpp
helfrij: Same here, I think outputStatusControl_ connectivity is handled by the allControls_ set.

source/ui/BioXAS/BioXASZebraLogicBlockInputInvertedView.cpp
helfrij: The two classes are very similar! But I'm not sure there is an easy way to generalize them. I think the alternative might be to set up mappings and button groups in BioXASZebraLogicBlockView. I thought this way might be clearer, though it's at the expense of some code duplication. @dretrex, what do you suggest?

source/beamline/BioXAS/BioXASZebraLogicBlock.h
```diff
@@ -0,0 +1,66 @@
...
+
+	/// Returns the connected state.
+	bool isConnected() const { return connected_; }
+
+	/// Returns the output state value.
+	double outputStateValue() const;
+	/// Returns true if the output state is low.
+	bool isStateLow() const { return !isStateHigh(); }
+	/// Returns true if the output state is high.
+	bool isStateHigh() const;
```
dretrex: It would probably be better to be explicit with these names since there are many states.  (`bool isOutputStateHigh() const;`)

source/ui/BioXAS/BioXASZebraLogicBlockInputInvertedView.cpp
dretrex: I think @XuetaoLiu-CLS is right that we could probably generalize these.  Since you've made the ZebraInput class answer both `isEnabled()` and `isInverted()` questions, you should be able to make a view that updates its appearance  based on those questions being answered (and correspondingly, their signals being emitted).

dretrex:
Time Estimate: 15 minutes

source/ui/BioXAS/BioXASZebraLogicBlockInputInvertedView.cpp
helfrij: Okay! I have some questions about how best to go about this, when you have a second @dretrex.

source/beamline/BioXAS/BioXASZebraLogicBlock.h
helfrij: Okay! I'll make the changes.

helfrij:
I think I've made the necessary corrections. Thanks @XuetaoLiu-CLS @dretrex.

helfrij:
Hey @dretrex, let me know if there are more changes you'd like to see made.

dretrex:
Nope!  Looks good!

PR #1885, opened 2016-01-21
Pull Request for Issue1604: Add pile up peaks to AMSingleSpectrumView

dretrex:
I essentially ported over the existing code from AMXRFDetailedDetectorView to do this, therefore there isn't a lot of new code.  Just duplicated code.

### DIFF
--- a/source/ui/dataman/AM2DScanView.cpp
+++ b/source/ui/dataman/AM2DScanView.cpp
@@ -492,7 +492,22 @@ void AM2DScanView::mousePressEvent(QMouseEvent *e)
 
 void AM2DScanView::setPlotRange(double low, double high)
 {
-	spectrumView_->setPlotRange(low, high);
+	spectrumView_->setEnergyRange(low, high);
+}
+
+void AM2DScanView::addSingleSpectrumEmissionLineNameFilter(const QRegExp &newNameFilter)
+{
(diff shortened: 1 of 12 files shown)


source/ui/dataman/AM2DScanView.cpp
```diff
@@ -492,7 +492,52 @@ void AM2DScanView::mousePressEvent(QMouseEvent *e)
 
 void AM2DScanView::setPlotRange(double low, double high)
 {
-	spectrumView_->setPlotRange(low, high);
```
helfrij: Just curious, what's the difference between setPlotRange(...) and setEnergyRange(...) for the spectrum view? Or has this method just been renamed?

dretrex: It has been renamed in the spectrum view.

helfrij: Since the scan view methods just call this renamed function, should the scan view methods be renamed as well?

helfrij:
Looks good!

Time Estimate: 15 minutes.

source/ui/dataman/AM2DScanView.cpp
dretrex: Yeah, that's probably a good idea.  I'll make that change.

Issue #920, opened 2014-10-21
Removing the synchronized dwell time from AMDetector


dretrex:
Since the main reason why @davidChevrier implemented the synchronized dwell time stuff in AMDetector was for requirements that are no longer relevant (and our own tools are now much better and refined), I think we should torch the synchronized dwell time.  I am currently removing it from VESPERS in #915 which would leave only SGM as a beamline that uses it.

What are people's thoughts on the matter?  If we are in favour of removing it, then issue #194 will no longer be required either.

dretrex:
Since PGM requires and uses the synchronized dwell time to get their pico ammeters to work, this needs to stay.

PR #1890, opened 2016-01-22
Pull Request for Issue1889: Some scans hang when running with ImmediateRead detectors

iainworkman:
- Added a check in [AMEXAFS/AMGeneric/AMTimed]ScanActionControllerAssembler that a given detector supports synchronized dwell acquisition before attempting to create an action to set its dwell time.

### DIFF
--- a/source/acquaman/AMEXAFSScanActionControllerAssembler.cpp
+++ b/source/acquaman/AMEXAFSScanActionControllerAssembler.cpp
@@ -213,10 +213,12 @@ AMAction3 *AMEXAFSScanActionControllerAssembler::generateActionTreeForEXAFSStepA
 
 			for(int x = 0; x < detectors_->count(); x++){
 
-				AMAction3 *detectorSetDwellAction = detectors_->at(x)->createSetAcquisitionTimeAction(variableIntegrationTimes.at(i));
+				if(detectors_->at(x)->supportsSynchronizedDwell()) {
+					AMAction3 *detectorSetDwellAction = detectors_->at(x)->createSetAcquisitionTimeAction(variableIntegrationTimes.at(i));
 
-				if(detectorSetDwellAction)
-					detectorSetDwellList->addSubAction(detectorSetDwellAction);
+					if(detectorSetDwellAction)
+						detectorSetDwellList->addSubAction(detectorSetDwellAction);
+				}
 			}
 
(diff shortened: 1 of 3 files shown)


iainworkman:
@XuetaoLiu-CLS Hey Sheldon, would you mind having a look over this for me?

Thanks

source/acquaman/AMEXAFSScanActionControllerAssembler.cpp
```diff
@@ -213,10 +213,12 @@ AMAction3 *AMEXAFSScanActionControllerAssembler::generateActionTreeForEXAFSStepA
 
 			for(int x = 0; x < detectors_->count(); x++){
 
-				AMAction3 *detectorSetDwellAction = detectors_->at(x)->createSetAcquisitionTimeAction(variableIntegrationTimes.at(i));
+				if(detectors_->at(x)->supportsSynchronizedDwell()) {
```
XuetaoLiu-CLS: Just want to confirm that this won't affect the usage of some detectors. I noticed that the `VESPERSCCDDetector/VESPERSFourElementVotexDetector/VESPERSSingleElementVotexDetector/SXRMBFourElementVortexDetector/SXRMBBrukerDetector/REIXSESMCPDetector/IDeasKETEKDetector/IDEAS13ElementGeDetector/AMXspress3XRFDetector(super class of n-element detector of BioXAS)` doesn't support sychronized dwell.

XuetaoLiu-CLS:
Looks good, except one question.

Time Estimate: 10 minutes

source/acquaman/AMEXAFSScanActionControllerAssembler.cpp
dretrex: I agree with this comment.  The supportsSynchronizedDwell() may not be the call you want because I believe that is specifically referencing the Synchronized dwell time that we have replaced and I've even made an issue to remove from the AMDetector API (re: #920).

dretrex:
Time Estimate: 5 minutes

source/acquaman/AMEXAFSScanActionControllerAssembler.cpp
iainworkman: Yeah, it seems as if the check will have to be whether the read method is request read.

iainworkman:
Okay, updates those to check the read mode of the detector instead. 
Also found a couple more places in the XAFS one which needed the check.

dretrex:
Alright, great.  I'll merge this in.

Issue #1857, opened 2016-01-15
Pulse modules should dynamically change their Time Units if the current time is not possible


dretrex:
Since the Zebra has limitations on the time, the AMXspress3XRFDetector should dynamically alter (or maybe the pulse module itself?) the time units before setting the pulse width.

As a corollary, if we are ever doing an XAS scan with an EXAFS region we should immediately set the time units to be 10 seconds because EXAFS regions have maximum times of >6.5 seconds.  (In general, the pulse modules should probably default to 10s since that will offer the most time options for the time scales we care about.)

helfrij:
Done in #1866.

Time Estimate: 1 day.

PR #1898, opened 2016-01-25
Pull Request for Issue1873: Creating time control for BioXASZebra

helfrij:
The time control handles resolving a pulse width and time units to a time value in seconds. The time logic introduced previously in #1857 has been moved to this new class, and this control is now used as the 32 Ge detector's acquisition time control.


helfrij:

iainworkman:
Looks good to me.
Time Estimate: 10 minutes

helfrij:
I have a question, @iainworkman: in 1f823c0c6135df87ed1953025211abd82e439816, I change AMPseudoMotorControl::setValue(...) to use withinTolerance(...) to check whether a value can be updated. This potentially means that the value_ is no longer updated every time a value is set. Do you think this could potentially cause any problems?

dretrex:
That can definitely cause potential problems.  It's like controls that don't update if you write the same value to it will have a special if-statement that calls `setStarted()` and `setSucceeded()` right after the other.  You need to be really careful when dealing with the values of controls.  In general, `withinTolerance()` should probably only be used in conjunction with move() and associated methods.

helfrij:
Made that change to AMPseudoMotorControl. Thanks @iainworkman @dretrex!

dretrex:
Aside from the one remaining comment, could you also fix the merge conflict @helfrij?

source/beamline/BioXAS/BioXASZebraTimeSeconds.cpp
```diff
@@ -0,0 +1,222 @@
+#include "BioXASZebraTimeSeconds.h"
+#include "actions3/AMActionSupport.h"
+#include "actions3/AMListAction3.h"
+
+BioXASZebraTimeSeconds::BioXASZebraTimeSeconds(const QString &name, QObject *parent) :
```
dretrex: I know this is really pedantic, but I feel like the name should include Control in some capacity since it is mapping a concept to PVs.  Even if it was just BioXASZebraTimeSecondsControl... I don't know.  Thoughts @acquaman/acquaman-working-group?

source/beamline/BioXAS/BioXASZebraTimeSeconds.cpp
```diff
@@ -0,0 +1,222 @@
...
+	} else if (validTimeValue(setpointDS)) {
+		timeValueSetpoint = setpointDS;
+		timeUnitsSetpoint = DSeconds;
+	}
+
+	// If the time value and units setpoints are valid, create and return valid action.
+
+	if (validTimeValue(timeValueSetpoint) && validTimeUnits(timeUnitsSetpoint)) {
+		AMListAction3 *moveAction = new AMListAction3(new AMListActionInfo3(QString("Moving %1").arg(name()), QString("Moving %1").arg(name())), AMListAction3::Parallel);
+		moveAction->addSubAction(AMActionSupport::buildControlMoveAction(timeUnits_, timeUnitsSetpoint));
```
dretrex: Do the PVs (particularly the time units) monitor even if writing the same value?  If not, this action will fail because it's trying to move something and will be waiting for a while and then fail.

dretrex:
Time Estimate: 10 minutes

helfrij:
Note to self: check that move action doesn't fail when subcontrol values are identical to the current subcontrol values.

helfrij:
@dretrex

I changed the parallel move action list for the zebra time control to a sequential list, as discussed. Playing with the zebra EDM screen, the time value and time units PV controls do not signal a value change when the input value is the same as the current value. I did some more test scans this morning and they seemed to work as expected. Let me know if there are other things that I should look into.

dretrex:
If the control interacts properly then it sounds good to me.  I'll merge this in.  Thanks for looking into those extra couple points @helfrij!

PR #1901, opened 2016-01-25
Pull Request for Issue1899: Adding Zebra checks to BioXASXASScanActionController.

helfrij:


helfrij:

XuetaoLiu-CLS:
Looks good. The function `buildScanControllerImplementation()` is really big though.

Time Estimates: 5 minutes

dretrex:
Also looks good from my end.  @XuetaoLiu-CLS, in general, the `buildScanControllerImplementation()` method is going to be really large because most of the meat of the controller is built there.  Now if you're referring to having smaller helper functions, then that is a different story and probably quite valid since some of the scan controller code can get a little unwieldy at times.

Time Estimate: 5 minutes

helfrij:
Breaking this single method up into smaller function calls is something worth looking into. It may be tricky for BioXAS, as many of the component pieces are interconnected (eg. only do this if I0Detector is non-zero, etc).

dretrex:
You may be able to get away with making some helper methods like the following:

```c++
if (hasI0())
     addI0Helper();
```

rather than having something like:

```c++
addI0();
addI1();
addI2();
```

Some food for thought.

Issue #1786, opened 2015-12-07
Acquaman crashed when add "2D Summing" analysis tool for EXAFS scan


XuetaoLiu-CLS:
Acquaman crashed when trying to add a "2D summing" analysis tool for EXAS scan. Crash stack tracking:
```
ls/home/sxrmb/beamline/programming/acquaman/source/application/AMCrashMonitorSupport.cpp:132
sigaction.c:0
/home/sxrmb/beamline/programming/acquaman/source/analysis/AM2DSummingAB.cpp:76
/home/sxrmb/beamline/programming/acquaman/source/dataman/AMAnalysisBlock.cpp:62
/home/sxrmb/beamline/programming/acquaman/source/ui/dataman/AMDataSourcesEditor.cpp:435
/home/sxrmb/beamline/programming/acquaman/build_files/moc_AMDataSourcesEditor.cpp:94
??:0
??:0
??:0
??:0
??:0
??:0
/home/sxrmb/beamline/programming/acquaman/source/application/SXRMB/SXRMBMain.cpp:54
```

crash happened in function ```setInputDataSourcesImplementation()``` of AM2DSummingAB.cpp, when trying to disconnect `valuesChanged(AMnDIndex, AMnDIndex)` signal.
```disconnect(inputSource_->signalSource(), SIGNAL(valuesChanged(AMnDIndex,AMnDIndex)), this, SLOT(onInputSourceValuesChanged(AMnDIndex,AMnDIndex)));```

@dretrex Does this look familiar with you? Thx

dretrex:
It's because `inputSource_` isn't being initialized in the constructor.

dretrex:
Was this issue not fixed earlier this week @XuetaoLiu-CLS?

XuetaoLiu-CLS:
It should be fixed. But the users didn't test it yet.

helfrij:
Seen today on BioXAS.

XuetaoLiu-CLS:
@helfrij Thx. I just tested on SXRMB. It still exists. It will crash when adding "2D-summing" during scan and won't crash after the scan is finished.

XuetaoLiu-CLS:
Time Estimates: 3 days

PR #1902, opened 2016-01-25
PullRequest for Issue1786: disable sumSelector to change axisSum_ for 2DSumming, 3D/4DBinning analysis block

XuetaoLiu-CLS:
#1786
- some code refactor for AM2DSummingAB
- disable the combobox of sumSelector to change the axisSum_ of 2DSummingAB, 3DBinningAB and 4DBiningAB

### DIFF
--- a/source/analysis/AM2DSummingAB.cpp
+++ b/source/analysis/AM2DSummingAB.cpp
@@ -34,12 +34,14 @@ AM2DSummingAB::AM2DSummingAB(const QString& outputName, QObject* parent)
 	sumRangeMax_ = 0;
 	analyzedName_ = "";
 	canAnalyze_ = false;
+
+	// cached information
 	cacheUpdateRequired_ = false;
+	dirtyIndices_.clear();
 	cachedDataRange_ = AMRange();
 
 	axes_ << AMAxisInfo("invalid", 0, "No input data");
 	setState(AMDataSource::InvalidFlag);
-
 }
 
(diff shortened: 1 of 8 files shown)


XuetaoLiu-CLS:
@dretrex I didn't use QLabel for the sum selector as you suggested. I still keep the combox, but I disabled it. Maybe in the future we might still want to use the combox?

dretrex:
That's fair.

dretrex:
Couple comments:

- With the refactors, does the 2D summing block still work appropriately after your refactor of `setInputSource()`?
- Did AM3DBinningAB and AM4DBinningAB not need to be initialized to use the last rank of the input source?
- Was the second version of `canAnalyze()` not used?  I just want to make sure that the refactors you did do not cause further issues.

Time Estimate: 10 minutes

XuetaoLiu-CLS:
@dretrex 
1) yes, `setInputDataSourcesImplementation` still works after the refactor. Basically there are a lot of redundant codes in `setInputDataSourcesImplementation()` and `updateInputSource()`. and the `updateInputSource()` was called by the `setInputDataSourcesImplementation`.
2) AM2DBinningAB already sets sumAxis_ to 2 and AM3DBinningAB already set sumAxis to 3
3) the second version of `canAnalyze()` is not used since what it did is to update the `canAnalyze_` flag, which is updated again in the function `updateInputSource()`

dretrex:
Okay.

PR #1904, opened 2016-01-25
Pull Request for Issue1897: Customized region of interest values are not propogated

dretrex:
The regions of interest were being added and removed from the detectors and scan configurations correctly, but there was no update if the regions were customized.  This has now been remedied.  I updated for all beamlines that require it, which is all but REIXS and SGM.

@iainworkman, will this change need to happen to SGM?  It's something that should be considered an addition when you merge master into SGMUpgrade.


source/acquaman/AMGenericStepScanConfiguration.cpp
```diff
@@ -269,6 +269,16 @@ void AMGenericStepScanConfiguration::removeRegionOfInterest(AMRegionOfInterest *
 		}
 }
 
+void AMGenericStepScanConfiguration::setRegionOfInterestBoundingRange(AMRegionOfInterest *region)
+{
```
helfrij: Should there be a check here to make sure that the region argument is valid?

source/acquaman/IDEAS/IDEASScanConfigurationDbObject.cpp
```diff
@@ -81,3 +81,13 @@ void IDEASScanConfigurationDbObject::removeRegionOfInterest(AMRegionOfInterest *
 			setModified(true);
 		}
 }
+
+void IDEASScanConfigurationDbObject::setRegionOfInterestBoundingRange(AMRegionOfInterest *region)
+{
```
helfrij: Could there be a region check here, as well?

source/acquaman/SXRMB/SXRMBScanConfigurationDbObject.cpp
```diff
@@ -81,6 +81,16 @@ void SXRMBScanConfigurationDbObject::removeRegionOfInterest(AMRegionOfInterest *
 		}
 }
 
+void SXRMBScanConfigurationDbObject::setRegionOfInterestBoundingRange(AMRegionOfInterest *region)
+{
```
helfrij: Region check here?

source/acquaman/VESPERS/VESPERSScanConfigurationDbObject.cpp
```diff
@@ -161,3 +161,13 @@ void VESPERSScanConfigurationDbObject::removeRegionOfInterest(AMRegionOfInterest
 		}
 }
 
+void VESPERSScanConfigurationDbObject::setRegionOfInterestBoundingRange(AMRegionOfInterest *region)
+{
```
helfrij: And here...

source/dataman/BioXAS/BioXASUserConfiguration.cpp
```diff
@@ -49,3 +49,13 @@ void BioXASUserConfiguration::removeRegionOfInterest(AMRegionOfInterest *region)
 			setModified(true);
 		}
 }
+
+void BioXASUserConfiguration::setRegionOfInterestBoundingRange(AMRegionOfInterest *region)
+{
```
helfrij: And here?

helfrij:
Quick question about checking whether regions are valid.

Time Estimate: 10 minutes.

source/acquaman/AMGenericStepScanConfiguration.cpp
dretrex: Pedantically speaking, you are correct.  That has been typically omitted in this (and the addRegionOfInterest and removeRegionOfInterest) because the region of interest itself is emitting the signal through `emit signal(this);` and if the region was not valid then that would cause a seg fault.

PR #1905, opened 2016-01-25
Pull Request for Issue1841: Zebra initialization on application startup

helfrij:
- Created new control for the soft inputs, a subclass of AMSinglePVControl. This control additionally has the .HIGH field control as a member variable. Added the ability to specify a .HIGH field preference before the control is actually connected, and have that preference applied once the control connects.
- Added initialization of the pulse controls and the soft input controls to BioXASSideBeamline.


helfrij:

source/beamline/BioXAS/BioXASZebraPulseControl.cpp
```diff
@@ -200,8 +203,10 @@ void BioXASZebraPulseControl::setInputValue(int value)
 
 void BioXASZebraPulseControl::setEdgeTriggerValue(int value)
 {
-	if (!edgeTriggerControl_->withinTolerance(double(value)))
-		edgeTriggerControl_->move(double(value));
+	if (edgeTriggerValue_ != value) {
+		edgeTriggerValue_ = value;
```
XuetaoLiu-CLS: Don't we need to update `edgeTriggerValue_` when the value of `edgeTriggerControl_` is changed (for example by the caput)?

source/beamline/BioXAS/BioXASZebraSoftInputControl.cpp
```diff
@@ -0,0 +1,46 @@
...
+	if (timeBeforeResetControl_->canMeasure())
+		result = timeBeforeResetControl_->value();
+
+	return result;
+}
+
+void BioXASZebraSoftInputControl::setTimeBeforeReset(double newValue)
+{
+	if (timeBeforeResetValue_ != newValue) {
+		timeBeforeResetValue_ = newValue;
```
XuetaoLiu-CLS: same question here.

XuetaoLiu-CLS:
Looks good, other than two questions.

Time Estimates: 10 minutes.

source/beamline/BioXAS/BioXASZebraPulseControl.cpp
helfrij: I guess it depends! I see what you mean, if someone caput's a new value to the edge trigger control or makes a value change using the EDM screen, the edgeTriggerValue_ is no longer identical to the value of the edgeTriggerControl_ as written. 

The point of storing the value is so that it can be specified before the control connects and applied to the control once it is connected, intentionally overriding any previous values on application startup. If someone changes the value outside of AM, I think it's an indication of something going wrong or that the control is being misused. I'm not sure whether that is best represented as having the edgeTriggerValue_ synchronized to the control's value or not.

helfrij: Perhaps instead of calling this value edgeTriggerValue_, a better name would be edgeTriggerInitialValue_ or edgeTriggerValuePreference_?

dretrex: You are right @helfrij.  And I do like the idea of changing the name to preference.  I think it makes it clearer.

XuetaoLiu-CLS: Thanks @helfrij. The name change makes it clearer.

helfrij:
I believe I've made the appropriate changes for the issues that were brought up. Thanks @XuetaoLiu-CLS @dretrex!

dretrex:
Looks good.
Time Estimate: 3 minutes

PR #1908, opened 2016-01-26
Pull Request for Issue 1906: Remove Detector Mode Check from Scan Action Assemblers

iainworkman:
Removed the check to ensure that the read mode was request read when adding detectors to the scan action assemblers, as it was interfering with XRF detectors.

### DIFF
--- a/source/acquaman/AMEXAFSScanActionControllerAssembler.cpp
+++ b/source/acquaman/AMEXAFSScanActionControllerAssembler.cpp
@@ -156,12 +156,11 @@ AMAction3 *AMEXAFSScanActionControllerAssembler::generateActionTreeForEXAFSStepA
 
 		for(int x = 0; x < detectors_->count(); x++){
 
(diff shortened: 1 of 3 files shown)


iainworkman:
Hey @XuetaoLiu-CLS  Would you mind having a look at this pull request. It basically undos the code added by #1890 , so it should be straight forward.

XuetaoLiu-CLS:
Looks good. But why do we want to redo the changes?

Time Estimates: 5 minutes

dretrex:
Looks good.

Time Estimate: 1 minute

dretrex:
@XuetaoLiu-CLS, it's because the read mode is not actually used by the scaler or AMXRFDetector.  It was causing issues where the scans could not dynamically change the time at the end of regions and during EXAFS scans because the Scaler and XRF detectors aren't set that way.  In fact, you may want to update the SXRMB deployment because they will probably be effected by this.

PR #1916, opened 2016-01-26
Pull Request for Issue1839: SGM Undulator Calculation Wrong For High Energies on the First Harmonic

iainworkman:
- Updated the energy trajectory test view in order to show the whole range of energy values
- Added a check to the undulator position calculation so that it is capped if the result would tend to infinity
- Set the undulator max position to 270.00mm

### DIFF
--- a/source/beamline/SGM/energy/SGMUndulatorSupport.h
+++ b/source/beamline/SGM/energy/SGMUndulatorSupport.h
@@ -4,6 +4,13 @@
 #define UNDULATOR_STEP_TO_POSITION_SLOPE 8.467394369e-5
 #define UNDULATOR_MIN_POSITION 12.51
 #define UNDULATOR_MAX_POSITION 270.0
+
+// The optimized undulator position function uses the following 3 parametric fit
+// values:
+#define UNDULATOR_PARAMETRIC_FIT_1 0.14295709668
+#define UNDULATOR_PARAMETRIC_FIT_2 36.00511212946
+#define UNDULATOR_PARAMETRIC_FIT_3 1737.41045746644
+
 #include "SGMGratingSupport.h"
 /*!
   * Namespace containing enumerators and functions related to the undulator on the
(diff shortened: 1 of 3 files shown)


iainworkman:
@helfrij Would you mind looking over this one?
Thanks!

source/beamline/SGM/energy/SGMUndulatorSupport.h
```diff
@@ -94,10 +94,14 @@ inline static bool validEnergy(UndulatorHarmonic undulatorHarmonic, double energ
...
   */
 inline static double optimizedUndulatorPosition(double energy,
-				  SGMUndulatorSupport::UndulatorHarmonic undulatorHarmonic,
-				  double undulatorOffset)
+                                                SGMUndulatorSupport::UndulatorHarmonic undulatorHarmonic,
+                                                double undulatorOffset)
 {
 	double rawUndulatorPosition = (-1/0.14295709668) * log( (1/36.00511212946)*((1737.41045746644/(energy/int(undulatorHarmonic))) -1)) + undulatorOffset;
+
+	if(energy > 1737.41045746644 && undulatorHarmonic == SGMUndulatorSupport::FirstHarmonic) {
```
helfrij: Quick question: what does this energy constant 1737.4... represent? Is it a calculated value?

helfrij:
I had a question about a large number used in a calculation in SGMUndulatorSupport. Maybe using a #define statement might make its purpose clearer?

Time Estimate: 10 minutes.

source/beamline/SGM/energy/SGMUndulatorSupport.h
iainworkman: According to the SGM wiki the 0.14295709668, 36.00511212946 and 1737.41045746644 values are all "parametric fit parameters", so I didn't know what to call them other than parametricFit1, parametricFit2 and parametricFit3 which I didn't think was much more helpful.

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
```diff
@@ -242,34 +263,20 @@ void SGMEnergyTrajectoryTestView::setupUi()
...
-		default:
-			startValue = 0;
-			endValue = 0;
-		}
+		double startValue = 200;
+		double endValue = 3000;
 
 		double increment = 1;
 
 		int numberOfItems = int((endValue - startValue) / increment);
```
dretrex: I know that this line isn't part of the pull request, but I think I must be missing something here.  There is the following code:

```c++
double increment = 1;
int numberOfItems = int(endValue-startValue)/increment);
```

Am I missing something and increment is supposed to be updated somewhere?  Also, generally I think you usually want to add a +1 to that calculation when computing the count of a number of items.

dretrex:
Only one comment.
Time Estimate: 5 minutes

source/tests/SGM/SGMEnergyTrajectoryTestView.cpp
iainworkman: increment is the value to increment the energy by for each X point. This way if I need to test something with greater fidelity than 1eV I can alter this one number (which I did during testing of this issue to ensure the I had the value which caused the asymptote down to values of 0.001eV). A better name for this would probably be deltaEnergy.

I've cleaned up this area of code a little bit and will push up shortly.

PR #1924, opened 2016-01-28
Pull Request for Issue1920: Resolving 'Top' window duplicate filter

helfrij:
- Removed the duplicate entries code in AMEnumeratedControl. It isn't used anywhere and is an unnecessary complication.
- Modified setWindowPreference for the actuator filter control, marks all entries that are not the preference to be read-only.

### DIFF
--- a/source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorFilterControl.cpp
+++ b/source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorFilterControl.cpp
@@ -76,7 +76,24 @@ void BioXASCarbonFilterFarmActuatorFilterControl::clearFilters()
 
 void BioXASCarbonFilterFarmActuatorFilterControl::setWindowPreference(double filter, int windowIndex)
 {
-	filterWindowPreferenceMap_.insert(filter, windowIndex);
+	if (indices_.contains(windowIndex)) {
+
+		// Identify any other window indices with the given filter value.
+		// Mark these windows as read-only.
+
+		QList<int> filterMatches = indexFilterMap_.keys(filter);
+
+		foreach (int index, filterMatches)
+			setIndexReadOnlyStatus(index, true);
+
+		// Insert preferred window to the map of filter preferences.
+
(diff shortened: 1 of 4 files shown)


helfrij:
Hey @iainworkman, would you take a look at this pull request? I made some modifications to AMEnumeratedControl, but I don't think they should impact SGM, from what I remember.

iainworkman:
Looks good to me.
Time Estimate: 5 minutes

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorFilterControl.cpp
```diff
@@ -76,7 +76,24 @@ void BioXASCarbonFilterFarmActuatorFilterControl::clearFilters()
 
 void BioXASCarbonFilterFarmActuatorFilterControl::setWindowPreference(double filter, int windowIndex)
 {
-	filterWindowPreferenceMap_.insert(filter, windowIndex);
+	if (indices_.contains(windowIndex)) {
+
+		// Identify any other window indices with the given filter value.
+		// Mark these windows as read-only.
+
+		QList<int> filterMatches = indexFilterMap_.keys(filter);
```
dretrex: Maybe I don't understand what's happening, but why is it the provided preference filter becomes the read only state?

dretrex:
Time Estimate: 5 minutes

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorFilterControl.cpp
helfrij: This is what I meant to do. If there are multiple windows that have the same filter (eg. the upstream actuator on Side has a 50um filter in Top and Bottom windows), you can choose to set a window preference for that particular filter (so the Bottom window is always used when you want a 50um filter). If there is a preference, it should be the only one that is a move destination, the other windows with that filter should be read-only.

dretrex: Oh okay.  I see what I was missing now.

Issue #1597, opened 2015-09-09
Exporting Ge 32-el spectrum data to file, outside of scans


helfrij:
The BioXAS team have been using the germanium detector on Side to help identify foreign material deposited on the M1 viewport. They have been using the detector interface in AM, manually acquiring data at a particular energy and examining the plotted results. They would like to be able to export the spectrum data to file.

XuetaoLiu-CLS:
@dretrex I think #1191 can cover this request, right?

dretrex:
Don't worry about it @XuetaoLiu-CLS.  I'm working on (and almost done) #1191 which will handle this issue at the same time.  Thanks for looking into it though!

dretrex:
Er.  Yes.  I completely mis-read your comment.

dretrex:
Handled in #1191.

PR #1928, opened 2016-01-29
Pull Request for Issue1191: Generalize XRF scans

dretrex:
This refactors all the specific XRF scan controller code into one place.  There is also a common "Export" button that is now just a part of the detailed detector view rather than requiring all the beamlines to implement it themselves.


dretrex:
Hey @iainworkman @helfrij @XuetaoLiu-CLS, I have mostly finished this issue.  I would like to do a bit of testing on Monday for some final pieces but overall I'm pretty happy with how this turned out.  As you can see, I refactored the XRF scans to use a single AM class (kind of along the same lines as the generic scan action controller).

Since this is a fairly important upgrade that effects many pieces of the framework, I would like all three of you to please review it for me.

A small piece that came for the ride is I added a `virtual QString details() const;` method to AMDetector and AMBeamline.  These are meant to be used to give a very human readable description of the state of the beamline/detector at the point of calling it.  I have implemented some basic versions to reproduce the same behaviour from the old XRF scan controllers, but if you'd like to see how I was imagining it being used, take a look at the VESPERS implementation.

@helfrij this also handles #1597.

@epengr this changes how these scans run.  I have provided some means of encapsulating the information you need (and I did some of it for you, I think).

source/beamline/AMBeamline.h
```diff
@@ -67,22 +67,21 @@ class AMBeamline : public AMControl, public AMBeamlineControlAPI, public AMBeaml
 	static AMBeamline* bl();
 	/// Call this to delete the beamline object instance
 	static void releaseBl();
-
+	/// Destructor.
 	virtual ~AMBeamline();
 
+	/// Returns a string with a human readable text of what is important about this detector.
```
helfrij: This looks like a function related to detectors, but is in AMBeamline? Should it be in AMDetector?

helfrij: Ah, I see. The comment confused me, but there is one function in AMDetector and another one in AMBeamline.

source/dataman/AMDbUpgrade1Pt6.cpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+QStringList AMDbUpgrade1Pt6::upgradeFromTags() const
+{
+	return QStringList();
+}
+
+bool AMDbUpgrade1Pt6::upgradeNecessary() const
+{
+	return databaseToUpgrade_->tableExists("VESPERSXRFScanConfiguration_table")
```
helfrij: Are the old tables removed at the end of the upgrade? If they aren't, it looks like this database upgrade will happen every time on startup for existing applications.

helfrij:
Looks good, @dretrex. Just a couple of questions.

Time Estimate: 20 minutes.

source/dataman/AMDbUpgrade1Pt6.cpp
dretrex: Good eye.  There is actually no code to drop tables in any of the database upgrade system.  In this case, the upgrade won't happen every time because the AMDbUpgrades table keeps track of which upgrades have happened by comparing the tags.  So we're safe with a call like this.

dretrex:
@iainworkman, @XuetaoLiu-CLS can you review this please?

dretrex:
I have tested this on BioXAS and it appears to work as expected.

iainworkman:
I've made it through parts of this about 4 times, but keep getting interrupted. I'm not ignoring it. Will try and have a look tonight.

IDEASAcquaman_internal.pro
```diff
@@ -13,8 +13,6 @@ HEADERS +=	source/application/IDEAS/IDEASAppController.h \
 	source/acquaman/IDEAS/IDEASXASScanConfiguration.h \
 	source/ui/IDEAS/IDEASXASScanConfigurationView.h \
 	source/beamline/IDEAS/IDEASKETEKDetector.h \
-	source/acquaman/IDEAS/IDEASXRFScanConfiguration.h \
-	source/acquaman/IDEAS/IDEASXRFScanController.h \
 	source/ui/IDEAS/IDEASXRFDetailedDetectorView.h \
```
iainworkman: It looks as if a lot of the other beamline specific XRF detailed detector views have been removed, but this one remains. Is there something specific to this one that means it has to stay?

source/acquaman/AMXRFScanController.cpp
```diff
@@ -0,0 +1,159 @@
...
+{
+	detector_ = qobject_cast<AMXRFDetector *>(AMBeamline::bl()->exposedDetectorByInfo(configuration->detectorConfigurations().at(0)));
+
+	scan_ = new AMScan;
+	scan_->setScanConfiguration(configuration);
+	scan_->setName(QString("XRF Scan - %1").arg(detector_->name()));
+	scan_->setFileFormat("amCDFv1");
+	scan_->setFilePath(AMUserSettings::defaultRelativePathForScan(QDateTime::currentDateTime()) % ".cdf");
+	scan_->setRunId(AMUser::user()->currentRunId());
+	scan_->setNotes(QString("%1\n\n%2\n\n").arg(detector_->details()).arg(AMBeamline::bl()->details()));
```
iainworkman: Now that the details() mechanism exists, is there a potential to do a similar thing for the other generic scan types?

source/application/AMDatamanAppController.cpp
```diff
@@ -750,11 +759,11 @@ bool AMDatamanAppController::startupRegisterDatabases()
 	success &= AMDbObjectSupport::s()->registerClass<AMXESScan>();
 	success &= AMDbObjectSupport::s()->registerClass<AM2DScan>();
 	success &= AMDbObjectSupport::s()->registerClass<AM3DScan>();
-	success &= AMDbObjectSupport::s()->registerClass<AMXRFScan>();
```
iainworkman: You can get rid of this because you're just using AMScan instead of a specific AMXRFScan type, right?

source/beamline/VESPERS/VESPERSBeamline.cpp
```diff
@@ -1176,6 +1176,60 @@ VESPERSBeamline::~VESPERSBeamline()
 
 }
 
+QString VESPERSBeamline::details() const
+{
+	// Build the notes for the scan.
```
iainworkman: Comment could be updated to be specific to the fact that this code now lives in the beamline.

IDEASAcquaman_internal.pro
iainworkman: I answered this question for myself just by reading the rest of the PR.

iainworkman:
Looks really good. 
Couple of question, mostly just to satisfy my own curiosity.
Time Estimate: 40 minutes

source/acquaman/AMXRFScanController.cpp
dretrex: Yes.  Something similar to the the `setNotes()` line could be added to both generic scan controllers.

source/application/AMDatamanAppController.cpp
dretrex: I did.  That's why there is a red - beside it :)

source/beamline/VESPERS/VESPERSBeamline.cpp
dretrex: Haha, very true.  I'll do that.

source/application/AMDatamanAppController.cpp
iainworkman: Haha yeah, I was just making sure I understood how you'd been able to remove this class. I got the answer to this too when I'd read the full pull request.

XuetaoLiu-CLS:
Looks great! Lots of code removals :fireworks: .

Time Estimate: 15 minutes

Issue #1412, opened 2015-05-25
Add veto PV toggle to REXIS XES Scans when pausing 


epengr:
The REIXS XES detector can't actually pause, but a veto line can be use to have it ignore all incoming counts - which amounts to the same thing.

I'd like the XES scan to toggle the OprVeto to '1' when the scan is paused and to '0' when it unpauses.

                    PDTR1610-4-I21-01:OprVeto       - turns veto on and off
                    PDTR1610-4-I21-01:Veto:state   - shows the current state of the veto

XuetaoLiu-CLS:
Time Estimate: 6 hours

PR #1929, opened 2016-01-29
PullRequest for Issue1412: add pausable AMTimer and enable the pause feature for the MCP detector of REIXS

XuetaoLiu-CLS:
1) add the pausable AMTimer, which can be paused/resumed
2) for the MCP detector of REIXS, when the XES scan is paused/resumed, the dwell timer will be paused/resumed as well.
30 when MCP detector dwelling is paused/resumed, the Veto controller will be disabled/enabled as well

#1412


XuetaoLiu-CLS:
@iainworkman do you mind review this PR? thx

source/acquaman/REIXS/REIXSXESScanController.cpp
```diff
@@ -182,6 +182,16 @@ bool REIXSXESScanController::startImplementation() {
...
 
+void REIXSXESScanController::pauseImplementation()
+{
+	REIXSBeamline::bl()->mcpDetector()->pauseDwelling();
+}
+
+void REIXSXESScanController::resumeImplementation()
+{
+	REIXSBeamline::bl()->mcpDetector()->resumeDwelling();
+}
```
iainworkman: What happens to the detector in cases where the Scan Controller is paused, and then cancelled or skipped? Will the veto remain active for the detector?

iainworkman:
Looks good. Just one question about ensuring the state of the detector always returns to the same state through each scan.

source/beamline/REIXS/REIXSXESMCPDetector.cpp
```diff
@@ -205,6 +211,24 @@ void REIXSXESMCPDetector::setTotalCountTarget(int totalCountTarget){
 		totalCountTarget_ = totalCountTarget;
 }
 
+void REIXSXESMCPDetector::pauseDwelling()
+{
+	if (dwellTimeTimer_) {
```
dretrex: If we don't have a dwell timer, do we still want to do this toggleVeto?  Seems like they would work together.

source/acquaman/REIXS/REIXSXESScanController.cpp
dretrex: More importantly, there are no `setPaused()` or `setResumed()` calls.  @XuetaoLiu-CLS, did you manage to get it into these states in testing?

source/util/AMTimer.h
```diff
@@ -0,0 +1,160 @@
...
+#define AMTIMER_H
+
+#include <QTimer>
+#include <QTime>
+/*!
+  * A class which provides a wrapper of a QTimer, adding the ability to pause/
+  * unpause, with setPaused(bool) and to find out how long time timer has been
+  * running with elapsedTime().
+  */
+class AMTimer : public QObject
```
dretrex: Design choice question.  What is the reason for choosing to encapsulate a QTimer rather than inherit from it?

dretrex:
Couple questions.
Time Estimate: 5 minutes

source/util/AMTimer.h
iainworkman: I worked on the original outline of the class back at the end of May before starting on the SGM Upgrade (Sheldon picked it up off me). To be honest I can't remember the reason I chose one over the other.

source/acquaman/REIXS/REIXSXESScanController.cpp
XuetaoLiu-CLS: @iainworkman Good call. When scan is cancelled/skipped, we will stop the timer ans set it to null. But Veto should be off when scan is cancelled/skipped/finished and be on when scan start. I will update that.

@dretrex I didn't test it yet. I want Iain to have a check whether I'm on the right track.

source/util/AMTimer.h
XuetaoLiu-CLS: When I was working on the functions `isActive()`, `isRunning()`, `isPaused()`, I found that it is easier/clearer to encapsulate a QTimer than inherit from it.

source/beamline/REIXS/REIXSXESMCPDetector.cpp
XuetaoLiu-CLS: It is changed now, ToggleVeto will work with the timer

XuetaoLiu-CLS:
As the test result, toggling veto won't control the MCP detector from pausing/resuming updating the total count

XuetaoLiu-CLS:
Test is done.

@dretrex @iainworkman I made some small changes due to the test. you can review those changes when you have time. Thx

source/beamline/REIXS/REIXSXESMCPDetector.cpp
```diff
@@ -343,3 +371,13 @@ void REIXSXESMCPDetector::acquisitionCancelledHelper(){
...
+
+void REIXSXESMCPDetector::toggleVeto(bool on)
+{
+	if (vetoControl_->isConnected() && vetoStateControl_->isConnected()) {
+		if (on)
+			vetoControl_->move(0); // veto off is to enable updating
+		else
+			vetoControl_->move(1); // veto on is to disable updating
+	}
+}
```
iainworkman: I feel like the semantics of this are backwards.
veto suggests stopping updates, which we can see in the if statement of this function is currently semantically reversed to the pv values. 
```
void REIXSXESMCPDetector::toggleVeto(bool on) {
    if(on)
        move veto pv to off
    else
        move veto pv to on
}
```

the result of this is that all the calls in the rest of the class also seem backwards eg.

```C++
void REIXSXESMCPDetector::resumeDwelling()
{
    if (dwellTimeTimer_) {
        dwellTimeTimer_->resume();
        toggleVeto(true);
     }
}
```
If we're resuming dwelling, it makes sense to turn the veto off.
The logic of the if statement needs to be reversed in ```toggleVeto```, and the boolean values need to be reversed in all the calls to ```toggleVeto```

XuetaoLiu-CLS: @iainworkman yes, the function name is not clear. It might be clear to rename the `toggleVetor()` function to `toggleMCPCountUpdate()`.

iainworkman:
With those changes this looks good to me.

XuetaoLiu-CLS:
@dretrex just noticed that this PR is still open. Do you mind taking a look at it? Thx.

@iainworkman did his review and the test is done.

dretrex:
Certainly.  Sorry for the delay.

source/beamline/REIXS/REIXSXESMCPDetector.cpp
dretrex: I think for clarity of this particular method we should not use "toggle".  Since we are specifically stopping the detector from counting photons, I would suggest a name such as `disableMCPCountUpdate(bool)`.  It also makes the boolean value map to the control value a little better as well because, if I'm reading correctly, the function would look like this,

```c++
if (disableUpdate)
    vetoControl_->move(1);
else
    vetoControl_->move(0);
```

dretrex:
Only the one comment.  Looks good otherwise!
Time Estimate: 5 minutes

PR #1931, opened 2016-01-30
Pull Request for Issue1671: Enumerated bi-state control

helfrij:
- Modified AMExclusiveStatesEnumeratedControl to inherit from AMSingleEnumeratedControl, instead of AMEnumeratedControl.
- Added default implementation of updateConnected to AMPseudoMotorControl. This method checks to see if all children are connected.
- Replaced BioXASEndstationShutter class with instance of CLSExclusiveStatesControl
- Replaced instances of CLSBiStateControl with CLSExclusiveStatesControl throughout BioXAS Side, Main, Imaging applications.


helfrij:

source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -593,6 +598,28 @@ AMAction3* AMPseudoMotorControl::createCalibrateAction(double oldValue, double n
...
+			AMControl *child = childControlAt(i);
+
+			if ( !(child && child->isConnected()) )
+				connected = false;
+		}
+
+		result = connected;
+	}
+
+	return result;
```
XuetaoLiu-CLS: if `childCount == 0`, the result is `false`. I think it should be true, right?

source/beamline/BioXAS/BioXASFrontEndValves.cpp
```diff
@@ -1,16 +1,16 @@
...
-	addValve(new CLSBiStateControl("VVR1607-5-I22-01", "VVR1607-5-I22-01", "VVR1607-5-I22-01:state", "VVR1607-5-I22-01:opr:open", "VVR1607-5-I22-01:opr:close", new AMControlStatusCheckerDefault(2), this), CLSBiStateControl::Open, CLSBiStateControl::Closed);
-	addValve(new CLSBiStateControl("VVR1607-5-I21-01", "VVR1607-5-I21-01", "VVR1607-5-I21-01:state", "VVR1607-5-I21-01:opr:open", "VVR1607-5-I21-01:opr:close", new AMControlStatusCheckerDefault(2), this), CLSBiStateControl::Open, CLSBiStateControl::Closed);
-	addValve(new CLSBiStateControl("VVR1607-5-I10-01", "VVR1607-5-I10-01", "VVR1607-5-I10-01:state", "VVR1607-5-I10-01:opr:open", "VVR1607-5-I10-01:opr:close", new AMControlStatusCheckerDefault(2), this), CLSBiStateControl::Open, CLSBiStateControl::Closed);
+	addValve(new AMReadOnlyPVControl("VVR1407-I00-01", "VVR1407-I00-01:state", this), 1, 4);
+	addValve(new AMReadOnlyPVControl("VVF1407-I00-01", "VVF1407-I00-01:state", this), 1, 4);
```
XuetaoLiu-CLS: So we want to use the magic number "1" and "4" here instead of `CLSBiStateControl::Open, CLSBiStateControl::Closed`?

source/beamline/BioXAS/BioXASSideBeamline.cpp
```diff
@@ -208,7 +208,7 @@ void BioXASSideBeamline::setupComponents()
 
 	// Endstation safety shutter.
 
-	endstationShutter_ = new  BioXASEndstationShutter("BioXASSideEndstationShutter", "SSH1607-5-I22-01", this);
+	endstationShutter_ = new  CLSExclusiveStatesControl("SSH1607-5-I22-01", "SSH1607-5-I22-01:state", "SSH1607-5-I22-01:opr:open", "SSH1607-5-I22-01:opr:close", this);
```
XuetaoLiu-CLS: is it possible to encapsulate  `endstationShutter_` and `frontEndShutters_` into class `BioXASShutters` (as a member of BioXASShutters instead of BioXASBeamline)? 

about the `fastShutter_` below (Ln395), is it a common shutter for main and image? if yes, it might be good to put it in the BioXASShutter too.

source/beamline/CLS/CLSExclusiveStatesControl.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	return (int(value()) == Closed);
+}
+
+bool CLSExclusiveStatesControl::isBetween() const
+{
+	return (int(value()) == Between);
+}
+
+bool CLSExclusiveStatesControl::setBaseControl(AMControl *newControl)
```
XuetaoLiu-CLS: This function is alike the implementation of super class `AMSingleEnumeratedControl`. The difference is this implementation has signal connections. 

My question is whether we need the signal connections for `AMSingleEnumeratedControl` as well? If yes, this implementation can be deleted.

XuetaoLiu-CLS:
Looks good other than some questions.

Time Estimates: 20 minutes

source/beamline/AMPseudoMotorControl.cpp
helfrij: I think I'd prefer to return false in that case. This method is meant to return true if any children are connected. If there are no children, then they can't be connected.

source/beamline/BioXAS/BioXASFrontEndValves.cpp
helfrij: I can add a #define statement for those values. That would probably make their meaning clearer.

source/beamline/BioXAS/BioXASSideBeamline.cpp
helfrij: I add the endstation shutter to BioXASShutters a couple of lines down (220).

It's certainly possible for the fast shutter to be added to the beamline shutters, but it would change some shutters semantics. Right now if the shutters are closed, there is a problem with the beamline and any data collected wouldn't be meaningful. If the fast shutter is added to the shutters group, the shutters would report as closed even if there isn't a problem. They would generally only report as open while acquiring, which isn't terribly helpful.

source/beamline/CLS/CLSExclusiveStatesControl.cpp
helfrij: There is definitely some overlap here, and you're right that the only real difference is the signal connections.
- The updateStates() needs to be called every time the control value changes, because the control's value also dictates the moving status (Between == moving). This is unique to this particular kind of control, I think.
- Same with the alarmChanged. I think this behavior is unique to this particular kind of control.

helfrij: I've modified the CLSExclusiveStatesControl::setBaseControl(...). It now calls AMSingleEnumeratedControl::setBaseControl(...), which cleans up the exclusive states control method considerably.

source/beamline/BioXAS/BioXASSideBeamline.cpp
dretrex: Good explanation, @helfrij.

source/beamline/AMPseudoMotorControl.cpp
dretrex: Either is fine.  If we think about this semantically, what does the connectivity of a pseudomotor even mean with no children?

helfrij: There are other ways to track and report connectivity, other than using child controls--using AMControlSets for example.

source/beamline/CLS/CLSExclusiveStatesControl.h
```diff
@@ -0,0 +1,62 @@
...
+
+protected slots:
+	/// Sets the status control. Reimplemented to consider additional control signals.
+	bool setStatusControl(AMControl *newControl);
+
+	/// Updates the moving status. Reimplemented to use the Between state to indicate movement.
+	virtual void updateMoving();
+
+	/// Adds the open state.
+	void addOpenState(AMControl *openTriggerControl, double openStatusValue, double triggerValue);
```
dretrex: Since these are all protected, what can cause the addition or removal of these states?

dretrex:
As a general comment, @helfrij, would you say that this could replace `CLSBiStateControl`?

source/beamline/BioXAS/BioXASSideBeamline.cpp
XuetaoLiu-CLS: @helfrij I saw you add them to the BioXASShutters. 

What I mean is whether ` endstationShutter_` and ` frontEndShutter_` should be instanced in `BioXASShutters` and be deleted from the `BioXASSideBeamline`?

dretrex:
Time Estimate: 10 minutes

source/beamline/CLS/CLSExclusiveStatesControl.h
helfrij: They exist as helper methods, right now. They are only called from the constructor.

helfrij:
Yes, I think this class is functionally equivalent to CLSBiStateControl.

source/beamline/BioXAS/BioXASSideBeamline.cpp
dretrex: Not quite, @XuetaoLiu-CLS.  Unlike some of the other beamlines, the shutters in BioXASShutters have a component safety aspect compared to the scientific meaning of the fast shutter.  Therefore, these are kept separate to keep their meaning distinct rather grouping them based purely on their name.

dretrex:
Cool, maybe we should make an issue for that?

source/beamline/BioXAS/BioXASSideBeamline.cpp
helfrij: 'BioXASShutters' is also a general class that is meant to be used equally well for Side, Main, and Imaging. All three have the same front-end shutters, but the endstation shutter is different for each. We could just as easily have 'BioXASSideShutters', 'BioXASMainShutters', and 'BioXASImagingShutters' where each shutter set is defined (instead of in each beamline class). I just opted for the 'single class' option.

PR #1932, opened 2016-02-01
Pull Request for Issue1927: Scaler scan initialization and cleanup

helfrij:
Modified the scan initialization and cleanup actions in the BioXAS XAS scan action controller. The scaler should now take an extra acquisition in initialization, and return to Continuous mode in cleanup.

### DIFF
--- a/source/acquaman/BioXAS/BioXASXASScanActionController.cpp
+++ b/source/acquaman/BioXAS/BioXASXASScanActionController.cpp
@@ -73,13 +73,21 @@ AMAction3* BioXASXASScanActionController::createInitializationActions()
 	CLSSIS3820Scaler *scaler = CLSBeamline::clsBeamline()->scaler();
 
 	if (scaler) {
-//		double regionTime = double(bioXASConfiguration_->scanAxisAt(0)->regionAt(0)->regionTime());
 
-//		scalerInitialization = new AMSequentialListAction3(new AMSequentialListActionInfo3("BioXAS Scaler Initialization Actions", "BioXAS Scaler Initialization Actions"));
-//		scalerInitialization->addSubAction(scaler->createContinuousEnableAction3(false));
-//		scalerInitialization->addSubAction(scaler->createDwellTimeAction3(regionTime));
-//		scalerInitialization->addSubAction(scaler->createStartAction3(true));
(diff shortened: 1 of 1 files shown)


helfrij:

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
```diff
@@ -73,13 +73,21 @@ AMAction3* BioXASXASScanActionController::createInitializationActions()
...
-//		scalerInitialization->addSubAction(scaler->createContinuousEnableAction3(false));
-//		scalerInitialization->addSubAction(scaler->createDwellTimeAction3(regionTime));
-//		scalerInitialization->addSubAction(scaler->createStartAction3(true));
-//		scalerInitialization->addSubAction(scaler->createWaitForDwellFinishedAction(regionTime + 5.0));
+		scalerInitialization = new AMSequentialListAction3(new AMSequentialListActionInfo3("BioXAS Scaler Initialization Actions", "BioXAS Scaler Initialization Actions"));
+
+		// Check that the scaler is in single shot mode and is not acquiring.
+
+		scalerInitialization->addSubAction(scaler->createContinuousEnableAction3(false));
+		scalerInitialization->addSubAction(scaler->createStartAction3(false));
```
XuetaoLiu-CLS: A quick question here, should we stop acquiring first or should we change mode first?

XuetaoLiu-CLS:
Looks good.
Time Estimate: 5 minutes

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
helfrij: I think it's safest to change mode first, here, because the scaler may get into a funny state otherwise. If it isn't acquiring, the scaler isn't in Continuous mode; if the total number of scans and the scans per buffer haven't been set, it isn't technically in SingleShot either. The risk now is that the scan will fail completely.

XuetaoLiu-CLS: thx. I'm just curious which one should happen first

dretrex:
Looks good.
Time Estimate: 2 minutes

Issue #1921, opened 2016-01-27
SXRMB: energy PV fails at the beginning of the scan


XuetaoLiu-CLS:
The SXRMB user starts to have this issue from this morning with the lastest acquaman version. It has been working well whole day yesterday.

I rolled back to the last stable version now, which doesn't have local storage 

-----
More information from the beamline staffs:
- This happens when we have the initialization actions. That is, we need to move the energy from the current value to a preset value. After that, when we are trying to do the scan, it will fail. But it will succeed if we rerun the scan at this point (since it is already at the energy).

XuetaoLiu-CLS:
Time Estimates: 5 hours

PR #1937, opened 2016-02-02
PullReques for Issue1921: change the tolerance of the energy PV

XuetaoLiu-CLS:
- For Issue1921, sometime the user might observe failure to move energy. The Energy coordinator had difference tolerance with the energy pv used in acquaman. Change the tolerance of the energy PV to 0.001 so that it will match with the coordinator. Therefore, the error we observed might be solved since we will move the energy even it is a small one.
- Follow Quanfeng's request, change the label of bruker selection to "Select Detector" and the "None" to "Transmission"

#1921


XuetaoLiu-CLS:
@helfrij Hi Jessie, do you mind reviewing this? Thx

source/beamline/SXRMB/SXRMBBrukerDetector.cpp
```diff
@@ -28,6 +28,7 @@ SXRMBBrukerDetector::SXRMBBrukerDetector(const QString &name, const QString &des
 
 	// Stuff specialized by SXRMBBrukerDetector
 	deadTimeControl_ = new AMReadOnlyPVControl("Dead Time", "mca1606-B10-03:mca1.DTIM", this);
+	allControls_->addControl(deadTimeControl_);
 	connect(deadTimeControl_, SIGNAL(connected(bool)), this, SLOT(onDeadTimeControlConnected(bool)));
```
helfrij: Is this connect statement for the deadTimeControl_ necessary, if the control is added to allControls_?

helfrij:
One quick question. Otherwise, looks good!

Time Estimate: 5 minutes.

source/ui/SXRMB/SXRMBScanConfigurationView.cpp
```diff
@@ -30,7 +30,7 @@ void SXRMBScanConfigurationView::updateFluorescenceDetectorComboBox(SXRMB::Fluor
 QComboBox *SXRMBScanConfigurationView::createFluorescenceComboBox()
 {
 	QComboBox *newComboBox = new QComboBox;
-	newComboBox->insertItem(0, "None");
+	newComboBox->insertItem(0, "Transmission");
```
dretrex: Hey @XuetaoLiu-CLS, I think it would be better to go with "Choose XRF Detector" and leave it as "None" as the option.  Transmission is a style of XAS where you only use ion chambers, not a detector choice.  In general, the scan is always doing transmission but you optionally have the choice to add an XRF detector.

source/beamline/SXRMB/SXRMBBrukerDetector.cpp
XuetaoLiu-CLS: It is kind of important. We didn't add the deadTimeControl_ to the set. And if deadTimeControl_ is not connected, we didn't set the connectivity of the bruker detector properly.

source/application/SXRMB/SXRMB.h
```diff
@@ -41,7 +41,7 @@ namespace SXRMB {
 	/// Enum for making the decision on what fluorescence detector the user wants to use.
 	enum FluorescenceDetector
 	{
-		NoXRFDetector = 0,
+		NoXRFDetector = 0, //Transmission detector
```
dretrex: Comment not necessary with my comment below.

source/ui/SXRMB/SXRMBScanConfigurationView.cpp
XuetaoLiu-CLS: @dretrex I will change in your suggestion. Change "None" to "Transmission" is what Quanfeng ased.

source/beamline/SXRMB/SXRMBBrukerDetector.cpp
dretrex: By adding `deadTimeControl_` to the `allControls_` set, you no longer need the specific connect from the dead time control.  Probably a change of:

```c++
connect(this, SIGNAL(connected(bool)), this, SLOT(onDeadTimeControlConnected(bool)));
```

XuetaoLiu-CLS: @dretrex in the slot, we emit the deadtimeChanged() signal, which will trigger the UI to update the value.
Adding to the control set will only update the connectivity status of the bruker detector

dretrex: I agree, that's why I suggested a different connect statement.

source/ui/SXRMB/SXRMBScanConfigurationView.cpp
dretrex: Yes, I remember you saying that.  In this particular case I think the changes you implemented are the better way to go.  If need be, explain to Quanfeng why we did it this way.  Emphasize that they are always doing transmission and this combo box is meant to add an XRF detector as well.

Issue #1930, opened 2016-01-29
Warnings generated from onCustomContextMenuActionImplementation() in AMCurrentAmplifierView


dretrex:
This is a small code cleanup that should be done because it generates a bunch of warnings.  I'm mostly making this an issue because it's the kind of thing that we should try to avoid coding ourselves and also to try and be diligent in our code review to spot these as well.

@XuetaoLiu-CLS, I'm going to assign this one to you since it is in the code you wrote.  Mostly because something I didn't notice before was now there are two context menu slots:

```c++
void onCustomContextMenuRequested(QPoint position);
virtual void onCustomContextMenuActionImplementation(QMenu *contextMenu);
```

Yours is the `QMenu *` and we should stick with that one, but could you make sure that only the `onCustomContextMenuActionImplementation()` version is being used and remove the other one?  Thanks!

XuetaoLiu-CLS:
Time estimate: 10 minutes

XuetaoLiu-CLS:
Both functions are required. One is the slot itself, the other is for customized implementation of the subclasses.

PR #1938, opened 2016-02-02
Pull Request for Issue1930: code refactor - move the slot implementation function to the non-slot functions

XuetaoLiu-CLS:
#1930 

- The customized functions for AMCurrentAmplifierView should be "protected" instead of "protected slot".
- Add Q_UNUSED() for the default implementation of the context menu function


XuetaoLiu-CLS:
@dretrex #1930 is not accurate. Both functions are required. One is the slot itself and the other is for customized implementation. it is similar to the other functions of AMCurrentAmplifiler.

dretrex:
Okay, great.  Thanks a lot for doing this @XuetaoLiu-CLS.  Looks good.
Time Estimate: 5 minutes

PR #1943, opened 2016-02-02
Pull Request for Issue1941: First point in scans using the scaler triggered using the Zebra not correct - Part 2

dretrex:
Turns out the problem with the first point was a little bit bigger than we originally estimated.  Required updating the BioXASSIS3820Scaler class to have the same code as the CLSSIS3820Scaler that waits on the channels before setting the trigger to be done.


helfrij:
Looks good, @dretrex.

Time Estimate: 10 minutes.

PR #1944, opened 2016-02-03
Pull Request for Issue1939: Modify BioXAS scaler to accommodate zebra when scanning. 

helfrij:
Modified the createStartAction3 method in CLSSIS3820Scaler, it is now virtual. BioXASSIS3820Scaler reimplements this method to use softInput_ instead of the startToggle_ for triggering a scaler acquisition. Modified CLSSIS3820Scaler::setScanning(...) to create and run the start action.


helfrij:
Hey @dretrex, I think these changes are ready for review.

dretrex:
Code-wise it looks good.  However, I have a general question about having code about "arming" the scaler in the CLS implementation.  Although it saves on lines of code, the CLS scaler does not arm (it even has requiresArming() return false) and then having code that is related to arming as part of the super class is a bit misleading.  Is it possible to make it so that only the BioXAS scaler needs to know about arming since it is the only one who cares about that information?

dretrex:
Time Estimate: 5 minutes

helfrij:
Hey @dretrex, I think I fixed the 'scanning' state problem we were seeing in 76f8241.

dretrex:
Time Estimate: 10 minutes

PR #1951, opened 2016-02-04
Pull Request for Issue1946: CLSSIS3820Scaler view doesn't show scaler as scanning while performing a scan (BioXAS)

dretrex:
Added a setScanningState call to the onTriggerSourceTriggered method inside the BioXASSIS3820Scaler class because there are two distinct triggering mechanisms now that the scaler is controlled by the Zebra.

### DIFF
--- a/source/beamline/BioXAS/BioXASSIS3820Scaler.cpp
+++ b/source/beamline/BioXAS/BioXASSIS3820Scaler.cpp
@@ -120,6 +120,7 @@ void BioXASSIS3820Scaler::onTriggerSourceTriggered(AMDetectorDefinitions::ReadMo
 {
 	Q_UNUSED(readMode)
 	initializeTriggerSource();
+	setScanningState(true);
 }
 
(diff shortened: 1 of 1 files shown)


helfrij:
Looks good, @dretrex. Thanks for taking care of this.

Time Estimate: 2 minutes.

helfrij:
Hey @dretrex, a question. I just ran a Side Acquaman application on my machine while there was a scan running on the Side OPI, with your changes. The scaler view status LED reports that the scaler is scanning when in Continuous mode, but never when in SingleShot. I guess my question is whether that's to be expected now, given that the scanning state for the BioXAS scaler is handled as an internal flag instead of as a PV value.

dretrex:
This is due to the fact that the scanning state is being set when the trigger source is triggered.  The trigger source will only live in the application that's running.  If we want  that light to be representative more generally, we'll need to listen to the arming and possibly the OR2 gate of the Zebra.

PR #1955, opened 2016-02-04
Pull Request for Issue1945: Refactoring the scaler controls view

helfrij:
Changes:
- Added connect statement, so the view listens for and can respond to changes in the scaler's connectivity.
- Moved some of the initialization code for the individual UI elements from the constructor to the methods that update them. The UI elements will be cleared and not enabled by default; they will only be editable if there is a valid, connected scaler in the appropriate acquisition mode (a change from previous behavior).
- Renamed some variables and functions, hopefully to make their function clearer.

Testing on Side & Main:
- Pressing start causes scaler to start scanning for the specified dwell time, only enabled in SingleShot and when not scanning.
- Pressing stop causes scaler to stop scanning/disarm, only enabled in SingleShot when scanning.
- Changing the dwell time in AM causes scaler EDM screen to update with new time.
- Changing scaler EDM dwell time causes AM screen to update.
- Changing scans per buffer in AM causes scaler EDM screen to update with new value.
- Changing scaler EDM scans per buffer causes AM screen to update.
- Changing total scans in AM causes scaler EDM screen to update with new value.
- Changing scaler EDM total scans causes AM screen to update.

### DIFF
--- a/source/ui/CLS/CLSSIS3820ScalerControlsView.cpp
+++ b/source/ui/CLS/CLSSIS3820ScalerControlsView.cpp
@@ -9,9 +9,7 @@ CLSSIS3820ScalerControlsView::CLSSIS3820ScalerControlsView(CLSSIS3820Scaler *sca
 
 	// Create UI elements.
 
-	modeChoice_ = new QComboBox;
-	modeChoice_->addItems(QStringList() << "Single Shot" << "Continuous");
-	modeChoice_->setCurrentIndex(0);
+	acquisitionModeBox_ = new QComboBox;
 
 	startButton_ = new QPushButton(QIcon(":/22x22/media-playback-start.png"), "Start");
 	startButton_->setMaximumHeight(25);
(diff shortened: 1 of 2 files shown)


helfrij:
Hey @iainworkman, would you take a look at this?

iainworkman:
Looks good to me.
Time Estimate: 15 minutes

source/ui/CLS/CLSSIS3820ScalerControlsView.cpp
```diff
@@ -134,102 +127,153 @@ void CLSSIS3820ScalerControlsView::stopScanning()
...
 
 void CLSSIS3820ScalerControlsView::setScansPerBuffer()
 {
-	if (scaler_) {
-		int newValue = scansPerBuffer_->value();
-
-		if (newValue != scaler_->scansPerBuffer())
-			scaler_->setScansPerBuffer(newValue);
+	if (scaler_ && scaler_->isConnected()) {
+		int newValue = scansPerBufferBox_->value();
```
dretrex: I understand the desire to have the dwell time be on separate lines because of the operation you do to it, but this one and the next ones really should be a single call.

dretrex:
Otherwise, looks good. 
Time Estimate: 5 minutes

source/ui/CLS/CLSSIS3820ScalerControlsView.cpp
helfrij: Np, I'll make the change.

PR #1960, opened 2016-02-04
Pull Request for Issue1948: Add Time Bounds Calculation to SGM XAS Configuration View

iainworkman:
- Added a function to the trapezoid velocity calculator which will give a time when provided with a start position, end position, target velocity and acceleration.
- Added a function to SGMGratingAngleControl which returns a time bounds for a move between two energies (taking into account current acceleration and grating translation)
- Added a calculation to the configuration view which sets the bounds of the time spin box based on these bounds, and displays a label to indicate the possible times to users.


iainworkman:
Hey @helfrij. Would you mind having a look over this pull request?
Thanks.

source/ui/SGM/SGMXASScanConfigurationView.cpp
```diff
@@ -11,6 +13,15 @@ SGMXASScanConfigurationView::SGMXASScanConfigurationView(SGMXASScanConfiguration
 	configuration_ = configuration;
 	detectors_ = 0;
 
+	gratingAngleControl_ = new SGMGratingAngleControl(this);
```
helfrij: Could the gratingAngleControl_ or the gratingTranslationStepControl_ be used anywhere else? If so, perhaps they could be passed into the view as parameters?

source/ui/SGM/SGMXASScanConfigurationView.cpp
```diff
@@ -246,6 +261,53 @@ void SGMXASScanConfigurationView::setDirection(AMScanConfiguration::Direction ne
...
+
+					dwellTime_->setValue(timeBounds.first);
+				}
+
+				dwellTime_->setRange(timeBounds.first, timeBounds.second);
+
+				timeBoundsLabel_->setText(QString("Valid times for energy range are: Min:%3s  Max:%4s")
+										  .arg(timeBounds.first, 0, 'f', 2)
+										  .arg(timeBounds.second, 0, 'f', 2));
+			}
```
helfrij: It looks like the timeBoundsLabel_ is not updated if the current grating is UnknownGrating or if the time bounds from the grating angle control are invalid. Should the view indicate a problem in this case?

helfrij:
A couple of quick questions. Otherwise, looks good.

Time Estimate: 15 minutes.

source/ui/SGM/SGMXASScanConfigurationView.cpp
iainworkman: I'm not super happy about this myself.
The are used elsewhere, but only in the energy coordinator application. Generally I kept all the calculations in there and hoped that the acquaman side could just dumbly throw values at the exposed energy PVs. Unfortunately, in order to do this calculation in the view the grating translation, grating angle encoder counts per step, and the grating acceleration are all required. 

I would like to refactor this in the future, but within the scope of this issue, this was the best arrangement.

source/ui/SGM/SGMXASScanConfigurationView.cpp
iainworkman: That would be a good idea. Thanks.

source/beamline/SGM/energy/SGMGratingAngleControl.cpp
```diff
@@ -166,6 +167,36 @@ AMAction3 * SGMGratingAngleControl::createDefaultsAction()
...
+                                                                      double endEnergy,
+                                                                      SGMGratingSupport::GratingTranslation currentTranslation)
+{
+
+	if(currentTranslation == SGMGratingSupport::UnknownGrating
+	        || startEnergy < 0
+	        || endEnergy < 0
+	        || qAbs(endEnergy - startEnergy) < 1) {
+
+		return QPair<double, double>(-1.0, -1.0);
```
dretrex: If this is a range, I'd suggest using AMRange.

dretrex:
Time Estimate: 5 minutes

dretrex:
Don't forget the estimate for the pull request @iainworkman.

source/ui/SGM/SGMXASScanConfigurationView.cpp
helfrij: Ah, okay!

iainworkman:
Time Estimate: 20 minutes

dretrex:
I also meant the ZenHub estimate ;)  Don't worry about it this time.  Just a friendly reminder to remember to set it when it's ready for review.

iainworkman:
Seems in the rush to get everything done on Friday the SGMUpgrade_mergedMaster branch eneded up in here. I'm going to create a new PR which merges this one into SGMUpgrade_mergedMaster.

PR #1962, opened 2016-02-04
Pull Request for Issue1958: Main scaler errors when moving between 'Continuous' and 'SingleShot'

helfrij:
- Removed the class variables for the scans per buffer and the total scans from the AcquisitionControl, they are not used and could be the potential source of some problems.
- Reimplemented updateValue() so that the control's value is only updated when a move is not in progress.

### DIFF
--- a/source/beamline/CLS/CLSSIS3820ScalerAcquisitionMode.cpp
+++ b/source/beamline/CLS/CLSSIS3820ScalerAcquisitionMode.cpp
@@ -17,9 +17,6 @@ CLSSIS3820ScalerAcquisitionMode::CLSSIS3820ScalerAcquisitionMode(const QString &
 	numberOfScansPerBufferControl_ = 0;
 	startScanControl_ = 0;
 
-	singleShotScanCountValue_ = 1;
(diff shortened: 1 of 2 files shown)


helfrij:

source/beamline/CLS/CLSSIS3820ScalerAcquisitionMode.cpp
```diff
@@ -17,9 +17,6 @@ CLSSIS3820ScalerAcquisitionMode::CLSSIS3820ScalerAcquisitionMode(const QString &
 	numberOfScansPerBufferControl_ = 0;
 	startScanControl_ = 0;
 
-	singleShotScanCountValue_ = 1;
-	singleShotNumberOfScansPerBufferValue_ = 1;
```
XuetaoLiu-CLS: Great catch! This should be removed for sure :100:

XuetaoLiu-CLS:
Looks good. Great code optimization.

Time Estimate: 10 minutes.

dretrex:
Looks good.
Time Estimate: 3 minutes

Issue #483, opened 2013-07-22
AMRun and AMFacility should be changed from hardcoded entries to a dynamic lookup


dretrex:
Like the title states, AMRun and AMFacility have antiquated design ideas in them and should be updated to not be hardcoded and knowing exact database ID's.

XuetaoLiu-CLS:
Time Estimate: 3 days

Issue #1697, opened 2015-11-03
Acquaman database upgrade: to update the beamline ID


XuetaoLiu-CLS:
Need to do SXRMB database upgrade to update AMFacility DB

XuetaoLiu-CLS:
Time Estimate: 2 days

davidChevrier:
We need to revisit this issue after the transition work.

PR #1968, opened 2016-02-05
PullRequest for Issue483: fix the misconnection betweem AMFacility and AMRun

XuetaoLiu-CLS:
- Add facilityId field to AMFacility
- User the facilityId to match that of AMRun
- Code generalization for CLSAppController

#483


XuetaoLiu-CLS:
@helfrij do you mind review this PR? Thx

XuetaoLiu-CLS:
This should wait for the db upgrade issue

XuetaoLiu-CLS:
This should wait for the db upgrade issue

XuetaoLiu-CLS:
This should wait for the db upgrade issue

helfrij:
Hey @XuetaoLiu-CLS, what's the issue number for the DB upgrade you're talking about?

XuetaoLiu-CLS:
@helfrij  #1697 Is the database upgrade issue. you can review this one first. #1697 will base one this branch and merge back. At that time, we can merge this to the master branch.

dretrex:
@XuetaoLiu-CLS, can you do the database upgrade in this issue?  These changes don't make sense without the database upgrade and can possibly corrupt databases.

dretrex:
I have taken a look at this and I some implementation issues.  Although this makes things a little more clean, it doesn't fix the underlying problem.  @XuetaoLiu-CLS we should chat on Monday where I can explain what the issues are and the path forward.

XuetaoLiu-CLS:
Hi @helfrij can you re-review this. 

I made new changes for this: we will retire the facilityId idea for AMFacility and AMRun. That is, in each beamline, there will be only one db record for AMFacility, which has the name and description. Then, in AMRun, there is no needs to track the FaciltyId for each run.

XuetaoLiu-CLS:
I will work on #1697 for the database upgrade, which should still be based on this branch and will merge to master together.

helfrij:
Ok @XuetaoLiu-CLS! I'll take another look.

BioXASAcquaman.pro
```diff
@@ -5,8 +5,9 @@
 TEMPLATE = subdirs
 SUBDIRS += \
 	Initialize.pro \
-	BioXASSideAcquaman.pro \
 	BioXASMainAcquaman.pro \
-	BioXASImagingAcquaman.pro \
+	BioXASMainTableCoordinator.pro \
```
helfrij: Was there a problem with the Main table coordinator that was addressed here?

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -239,7 +233,7 @@ void BioXASAppController::registerClasses()
 	AMDbObjectSupport::s()->registerClass<BioXASXASScanConfiguration>();
 }
 
-void BioXASAppController::setupExporterOptions()
+void BioXASAppController::setupBeamlineExporterOptions()
```
helfrij: Small detail. The beamline itself doesn't really export anything, what do you think of `setupApplicationExporterOptions()` or `setupDataExporterOptions()`?

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -247,22 +241,7 @@ void BioXASAppController::setupExporterOptions()
...
-	BioXASBeamline::bioXAS();
-}
-
-void BioXASAppController::initializePeriodicTable()
-{
-	AMPeriodicTable::table();
-}
-
-void BioXASAppController::setupUserInterface()
+void BioXASAppController::setupAcquamanUserInterface()
```
helfrij: Is it important to make a distinction between `setupUserInterface()` and `setupAcquamanUserInterface()`? I can see it being important if there is more than one user interface.

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -324,6 +303,11 @@ void BioXASAppController::setupUserInterface()
 	addPersistentView(new BioXASPersistentView());
 }
 
+void BioXASAppController::setupBeamlineSignalConnections()
```
helfrij: Do all CLSBeamline subclasses have to implement `setupBeamlineSignalConnections()`? I can see it being useful for some/most, but it doesn't appear to be used here.

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -50,22 +49,12 @@ bool BioXASAppController::startup()
...
-		initializePeriodicTable();
-
-		registerClasses();
+	if (dataFolderOK && CLSAppController::startup()) {
 
 		// Ensuring we automatically switch scan editors for new scans.
 		setAutomaticBringScanEditorToFront(true);
 
-		setupExporterOptions();
 		setupScanConfigurations();
```
helfrij: Because `setupScanConfigurations()` is left behind here, and `setupAcquamanUserInterface()` has been promoted to the CLSBeamline, I think views will be created for scan configurations before they exist on BioXAS.

helfrij:
A couple of minor questions. My biggest concern is that because the order of setting up scan configurations and creating scan configuration views is switched, there might be crashes on BioXAS. Will test to confirm.

Time Estimate: 30 minutes.

BioXASAcquaman.pro
XuetaoLiu-CLS: `BioXASMainTableCoordinator.pro` along with `BioXASSideAcquaman.pro` and `BioXASImagingAcquaman.pro` was original added in the `acquamanSuite.pro`. I think it will be better if we organize all the bioxas related pro files into `BioXASAcquaman.pro`.

or maybe we should move `BioXASMainTableCoordinator.pro` into `BioXASMainAcquaman.pro` and move `BioXASSideTableCoordinator.pro` into `BioXASSideAcquaman.pro`?

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: @helfrij it is not. When doing code refactor, I want to make thing clear that we have a "generic" function to setup the UI. I can change it back for sure.

BioXASAcquaman.pro
helfrij: I agree, it seems like a good idea to have the BioXAS stuff in BioXASAcquaman.pro instead of acquaman_suite.pro. It might be advantageous to have the table coordinators with each specific project if it makes sure they are built along with everything else when bioxasside_beamline.pro and bioxasmain_beamline.pro are built. Do you know if that's the case here?

source/application/BioXAS/BioXASAppController.cpp
helfrij: Was just curious, that's all.

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: I defined this function as a pure virtual function in CLSBeamline. Therefore, all the subclass must implement it. The older beamlines are implementing in this way --- organize all the signals as a code block. The BioXAS beamlines and new SGM beamline are NOT doing in this way. 

I don't think it hurts to leave it here. Maybe we will change minds in the future

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: Good catch.

dretrex:
Hey @XuetaoLiu-CLS, I took a quick look.  I'm really sorry, I think I've been really bad at explaining how I want this to look.  This is much closer, but still not quite what I had in mind.  I will make the change in a different branch to show what I'm thinking.  Again, sorry about that.

XuetaoLiu-CLS:
This branch is closed since we are going to use the new Issue483alt branch

PR #1969, opened 2016-02-05
Pull Reques for Issue1940: Replace CLSBiStateControl with CLSExclusiveStatesControl

helfrij:


helfrij:

source/beamline/VESPERS/VESPERSBeamline.cpp
```diff
@@ -1043,16 +1049,16 @@ void VESPERSBeamline::valveError()
```
XuetaoLiu-CLS: I know this is not related to this issue. But here is much easier to import a bug in the future, since we need to make sure that the first control of valveSet_ MUST be the `vvrFE1_`.

source/beamline/VESPERS/VESPERSBeamline.cpp
```diff
@@ -1233,7 +1239,7 @@ QString VESPERSBeamline::details() const
 bool VESPERSBeamline::allValvesOpen() const
 {
 	for (int i = 0; i < valveSet_->count(); i++)
-		if (valveSet_->at(i)->value() == 0 || valveSet_->at(i)->value() == 2)
+		if (valveSet_->at(i)->value() == CLSExclusiveStatesControl::Open)
```
helfrij: Could simplify this if-check to valveSet_->at(i)->isOpen().

XuetaoLiu-CLS: Although this code works, but as commented above, the first element of `valveSet_` is `vvrFE1_`, which is not an instance of `CLSExclusiveStatesControl`. Therefore, it might be risky to use `CLSExclusiveStatesControl` definitions for it.

helfrij: Ah yes. Here, that would definitely be the source of a bug.

source/beamline/VESPERS/VESPERSBeamline.cpp
```diff
@@ -1243,7 +1249,7 @@ void VESPERSBeamline::openValve(int index)
 {
 	if (index >= 0 && index < valveSet_->count()){
```
XuetaoLiu-CLS: Again, this function will work for sure.
But the index should start from 1, since index 0 is `vvrFE1`,which is an instance of `AMReadOnlyPVwStatusControl`.

XuetaoLiu-CLS:
Looks great! Except for one special case of VESPERSBeamline, where the `valveSet_` has a special Valve there. Maybe it should not be in the set?

Time Estimate: 10 minutes

source/beamline/VESPERS/VESPERSBeamline.cpp
helfrij: You're right. I'll check the areas you point out over again.

dretrex:
No further comments from me.  Looks good!
Time Estimate: 10 minutes

PR #1994, opened 2016-02-09
Pull Request for Issue1856: Synchronizing fast shutter to scaler when Ge detector not in use

helfrij:


helfrij:
It appears that the fast shutter is now active in a scan whether or not the Ge detector is involved, as discussed in #2066. This solution then is no longer needed.

Time Estimate: 1 day

PR #1995, opened 2016-02-09
Pull Request for Issue1989: Remove old code from VESPERS

dretrex:
Removed some really old applications that are no longer necessary.


helfrij:
Looks good, @dretrex.

Time Estimate: 2 minutes.

PR #2000, opened 2016-02-09
Pull Request for Issue1998: Remove old AM code

dretrex:
Removed some super old code that isn't required anymore.


helfrij:
Looks good, @dretrex. Out of curiosity, what were the analysis block plugins?

Time Estimate: 5 minutes.

dretrex:
We dabbled with the idea of making the analysis blocks plugins so that other people could potentially write their own and then just put them in a location and then AM would know about them.

PR #2010, opened 2016-02-10
Pull Request for Issue2005: JJ slits refactor

helfrij:
- Created new class AMSlitControl. This class can have two blade controls set, and provides the protected methods necessary to calculate the gap and center from the blade positions.
- Created new class AMSlitGap. This class is a subclass of AMSlitControl and it's value is the gap between the provided blades.
- Created new class AMSlitCenter. This class is a subclass of AMSlitControl and it's value is the center of the provided blades.
- Created new class AMSlit. This class has gap and center controls as members, and can have two blade controls set. It can also open or close the slit if the appropriate setpoints are provided.
- Created new class AMSlitView for displaying slit information.
- Updated CLSJJSlitsView to use AMSlitView and optionally include the blades controls and open/close buttons.
- Modified CLSJJSlits to use AMSlit instances for the vertical slit and horizontal slit.
- Removed classes that are no longer used:
  - CLSJJSlitBladesControl
  - CLSJJSlitGapControl
  - CLSJJSlitCenterControl


helfrij:

source/application/SXRMB/SXRMBAppController.cpp
```diff
@@ -374,7 +374,6 @@ void SXRMBAppController::setupUserInterface()
 	CLSCrossHairGeneratorControlView *crossHairView = new CLSCrossHairGeneratorControlView(SXRMBBeamline::sxrmb()->crossHairGenerator());
 	SXRMBCrystalChangeView *crystalChangeView = new SXRMBCrystalChangeView(SXRMBBeamline::sxrmb()->crystalSelection());
 	CLSJJSlitsView *jjSlitsView = new CLSJJSlitsView(SXRMBBeamline::sxrmb()->jjSlits());
-	jjSlitsView->setDataRange(18, 0);
```
XuetaoLiu-CLS: The dataRange is added for SXRMB because of a bug. I'm not sure removing this line will cause the bug back or not.

XuetaoLiu-CLS: I saw you removed the `setDataRange()` function as well. We should test it on SXRMB, whether this will cause an input issue.

source/beamline/AMControl.cpp
```diff
@@ -54,3 +54,11 @@ bool AMControl::hasChildControl(AMControl *control) const
 
 	return result;
 }
+
+void AMControl::removeChildControl(AMControl *control)
```
XuetaoLiu-CLS: This implementation is alike with the ones in the subclass `AMPseudoMotorControl`. except that the subclass called `updateState()` after the child is removed.

I'm wondering whether we should change the signature of this function to return a bool value. Then in the subclass (`AMPseudoMotorControl`), we can call 
```
if (AMControll::removeChildControl(control)) { 
    updateStates()
}
````

File: source/beamline/AMSlit.cpp [hunk 3]
```diff
@@ -0,0 +1,352 @@
...
+
+	if (firstBlade_ && bladeOrientationMap_.contains(firstBlade_))
+		result = ( int(bladeOrientationMap_.value(firstBlade_)) == OpensPositively);
+
+	return false;
+}
+
+bool AMSlit::firstBladeOpensNegatively() const
+{
+	bool result = false;
```
XuetaoLiu-CLS: maybe it can be simplified as `return ! firstBladeOpensPositively()` ?

source/beamline/AMSlit.h
```diff
@@ -0,0 +1,144 @@
...
+#define AMSLIT_H
+
+#include "beamline/AMControl.h"
+#include "beamline/AMSlitControl.h"
+#include "beamline/AMControlSet.h"
+
+#define AMSLIT_CANNOT_OPEN 39810
+#define AMSLIT_CANNOT_CLOSE 39811
+
+class AMSlit : public AMControl
```
XuetaoLiu-CLS: is it possible for this to subclass from `AMSlitControl`, since it is a pseudo control and also has `firstBlader_` and `secondBlader_`?

source/beamline/BioXAS/BioXASSideBeamline.cpp
```diff
@@ -229,8 +229,13 @@ void BioXASSideBeamline::setupComponents()
 
 	// JJ slits.
 
-	jjSlits_ = new CLSJJSlits("JJSlits", "SMTR1607-6-I22-10", "SMTR1607-6-I22-09", "SMTR1607-6-I22-11", "SMTR1607-6-I22-12", this);
-	connect( jjSlits_, SIGNAL(connectedChanged(bool)), this, SLOT(updateConnected()) );
+	jjSlits_ = new CLSJJSlits("BioXASMainJJSlits", "SMTR1607-6-I22-10", "SMTR1607-6-I22-09", "SMTR1607-6-I22-11", "SMTR1607-6-I22-12", this);
+	jjSlits_->setVerticalSlitOpenValues(-15, 15);
```
XuetaoLiu-CLS: we should set the ranges for SXRMBBeamline as well.

source/ui/beamline/AMSlitView.cpp
```diff
@@ -0,0 +1,166 @@
...
+	connect( openButton_, SIGNAL(clicked(bool)), this, SLOT(onOpenButtonClicked()) );
+
+	closeButton_ = new QPushButton("Close");
+	connect( closeButton_, SIGNAL(clicked(bool)), this, SLOT(onCloseButtonClicked()) );
+
+	QHBoxLayout *buttonsLayout = new QHBoxLayout();
+	buttonsLayout->addWidget(openButton_);
+	buttonsLayout->addWidget(closeButton_);
+
+	firstBladeEditor_ = new AMExtendedControlEditor(0);
```
XuetaoLiu-CLS: do we need to check the value range for those editors?

XuetaoLiu-CLS:
Looks good. Some questions about code reuse.

We need to pay attention to the usage in SXRMB, and need to test it especially the value range of the slit.

Time Estimates: 30 minutes

source/application/SXRMB/SXRMBAppController.cpp
helfrij: That's a good idea. Do you want to check it out together?

source/beamline/AMControl.cpp
helfrij: I agree, and I would like addChild() and removeChild() to return bools. Not sure how best to go about it, as they are used and reimplemented quite a lot in AMPseudoMotorControl subclasses. Should that be done in this pull request, or as something separate?

source/beamline/BioXAS/BioXASSideBeamline.cpp
helfrij: That's a good idea. We'll have to find out what those values are.

source/ui/beamline/AMSlitView.cpp
helfrij: Right now, I don't think the editors here have limits on their ranges. The bug that made them necessary before should be taken care of.

source/beamline/AMSlit.cpp
helfrij: I suppose that is logically equivalent, isn't it? Ha. I can make the change.

source/beamline/AMSlit.h
helfrij: It's true that the two classes do share firstBlade_, secondBlade_, and orientation map class variables, and maybe it's worth modifying AMSlit to inherit from AMSlitControl to take advantage of this. Beyond having those variables in common, though, the purpose of each class diverges: AMSlitControl is meant to provide the equations for calculating gap and center for different blade orientations, while AMSlit is the main object that provides access to the gap, center, and blade controls. AMSlitControls are measurable, AMSlit isn't. To me, it makes a little more sense to keep them separate but an argument could be made either way.

source/beamline/AMControl.cpp
dretrex: As something separate.

dretrex:
Once you two do the test on SXRMB, let me know and I'll go over this and merge it in.

XuetaoLiu-CLS:
I made the changes to SXRMB. But there is one issue left: the inbound/outbound of horizontal slit are totally opposite. As a result, the gap of horizontal slit is negative

untitled

dretrex:
Is it possible that you have them backwards @XuetaoLiu-CLS?

XuetaoLiu-CLS:
@dretrex as you can see in the SXRMBBeamline.cpp, I switched the inboard/outboard blade controls. The reason for that change is that for the new CLSJJSlit, the "inboard" blade will move negatively, and the "outboard" blade will move positively. Therefore, I have to switch them so that we can use the "open"/"close" features. 

Unfortunately, it seems that the horizontal GAP PV is calculated in the opposite way. As talked with David B., either we or he should make the changes.

dretrex:
Okay.  Thanks for clearing that up @XuetaoLiu-CLS.

helfrij:
Hmmm. Here's a question: should we have a CLS JJ slits class, if there's no convention for blade orientation? Since BioXAS is moving away from using individual motor controls in favor of EPICS gap and center pseudomotors, SXRMB will be the only ones using this CLS class.

dretrex:
It might be that everyone has their own subclass (until such a time that there is a CLS wide solution), but everyone inherits from AMSlits?  That way we still get the opportunity to have a framework base but not limiting ourselves to these weird implementation cases.

helfrij:
I think that makes sense for now. I can put together JJ slits classes for SXRMB and BioXAS. @XuetaoLiu-CLS, would you review the SXRMB class once it's ready?

helfrij:
Hey @XuetaoLiu-CLS, I've pushed up changes that should be ready to test for SXRMB. In case I have the orientation backwards for the motors, the setters have an optional argument for specifying whether they open positively or negatively. Let me know if you run into any problems :)

helfrij:
Hey @dretrex. @XuetaoLiu-CLS and I have made some small changes to the JJ slits code, and tested them on BioXAS and SXRMB. These changes are ready for review/merge.

helfrij:
Hey @dretrex, merge conflicts resolved here as well.

dretrex:
Time Estimate: 20 minutes

Issue #1976, opened 2016-02-08
REIXS crashes in XES scan action controller


helfrij:
Friday, 08:19

0	raise	/lib64/libc.so.6	0	0x3939a2ffc5	
1	abort	/lib64/libc.so.6	0	0x3939a31a70	
2	qt_message_output	/home/beamline/tools/qt/qt-4.7.3-SL54-x86_64/lib/libQtCore.so.4	0	0x2b78a1f7bb18	
3	??	/home/beamline/tools/qt/qt-4.7.3-SL54-x86_64/lib/libQtCore.so.4	0	0x2b78a1f7bcbc	
4	qFatal	/home/beamline/tools/qt/qt-4.7.3-SL54-x86_64/lib/libQtCore.so.4	0	0x2b78a1f7be75	
5	QList<AMMeasurementInfo>::at	qlist.h	456	0x5bb0c4	
6	AMCDFDataStore::measurementAt	AMCDFDataStore.h	117	0xacece6	
7	AMRawDataSource	AMRawDataSource.cpp	27	0x5b40df	
8	REIXSXESScanActionController::buildScanController	REIXSXESScanActionController.cpp	103	0xa8261d	
9	REIXSXESScanConfiguration::createController	REIXSXESScanConfiguration.cpp	182	0x9feadf	
10	AMScanAction::startImplementation	AMScanAction.cpp	120	0x6e41e0	
11	AMAction3::start	AMAction3.cpp	93	0x4e8d52	
12	AMAction3::qt_metacall	moc_AMAction3.cpp	125	0xa96ff9	
13	AMScanAction::qt_metacall	moc_AMScanAction.cpp	111	0xacca4b	
14	QObject::event	/home/beamline/tools/qt/qt-4.7.3-SL54-x86_64/lib/libQtCore.so.4	0	0x2b78a2090abc	
15	QApplicationPrivate::notify_helper	/home/beamline/tools/qt/qt-4.7.3-SL54-x86_64/lib/libQtGui.so.4	0	0x2b78a107585f	
16	QApplication::notify	/home/beamline/tools/qt/qt-4.7.3-SL54-x86_64/lib/libQtGui.so.4	0	0x2b78a107ce32	
17	QCoreApplication::notifyInternal	/home/beamline/tools/qt/qt-4.7.3-SL54-x86_64/lib/libQtCore.so.4	0	0x2b78a207eb64	
[...]

PR #2017, opened 2016-02-12
Pull Request for Issue1976: Crash bug in REIXSXESScanActionController.

helfrij:
It looks like this bug could have been caused by asking the scan for its first raw data source, without checking to see whether the scan had at least one source. The trace is listed in #1976. Added an if-check.


helfrij:
Hey @XuetaoLiu-CLS, would you look over this pull request?

XuetaoLiu-CLS:
Looks good. I'm just wondering whether we should add the same/similar protection for the other ScanActionControllers.

Time Estimation: 1 minute

dretrex:
In general, yes we should detect if there are no detectors.  It isn't an issue for the most part because aside from the XES scan controller here and the generic scan controllers, all the beamline specific scan controllers always have detectors added to them.

Time Estimate: 1 minute

PR #2023, opened 2016-02-16
Pull Request for Issue2022: Scaler dark current uses 'stale' channel values

helfrij:
First attempt: 
Added a wait action (0.5 seconds) between the scaler dwell time finished action and when processing the dark current information proceeds. This gives the scaler channels enough time to update their values with the new measurement.

Second attempt:
Created new class CLSSIS3820ScalerTriggerAction/Info. This action triggers the scaler's trigger source, initiating the chain of events already implemented in CLSSIS3820Scaler for arming and taking a measurement with channel detectors. Modified the CLS dark current measurement action to use this new trigger action.

Created BioXAS-specific dark current measurement action class. This action inherits from the CLS DC measurement action, but additionally initializes the Zebra with the ion chamber detectors.


helfrij:

source/actions3/actions/CLSSIS3820ScalerDarkCurrentMeasurementAction.h
```diff
@@ -4,6 +4,8 @@
 #include "actions3/AMAction3.h"
 #include "actions3/actions/CLSSIS3820ScalerDarkCurrentMeasurementActionInfo.h"
 
+#define CLSSIS3820SCALERDARKCURRENTMEASUREMENTACTION_WAIT_SECONDS 0.5
```
iainworkman: I assume this flat wait value is long enough to consistently give enough time for the values to populate?

iainworkman:
One small question regarding the length of time to wait, and consistent performance.
Otherwise everything looks good.
Time Estimate: 2 minutes.

dretrex:
@helfrij and I had a chat about this and she is going to implement a different solution that doesn't require waiting for a period of time but instead the scaler channels.  Hopefully she can use something similar to what is used in scans already.

helfrij:
Hey @dretrex, I took a different direction for this pull request and I'm not sure that it's what's best. I wrote a new action for the scaler, similar to the dark current correction action, that triggers the scaler via the trigger source. This should initiate the chain of events already implemented in the scaler class for waiting for detectors to complete measurements. In this pull request, I've modified the dark current action to use this new trigger action, that's how I've been testing.

As written, the scaler triggers correctly on Main but the zebra complicates things a little on Side. To use this code on Side, the zebra would have to be pre-loaded with detector managers and detectors on startup, and XAS scan configurations would have to adjust according to what's needed in the scan in initialization and restore defaults on cleanup--something we don't have support for (yet?).

helfrij:
I'm taking another look at the solution in this pull request, and I've remembered the problematic Zebra involvement here, @dretrex. In a normal scaler, the action triggers the scaler's trigger source, and behaves as we would expect it to: the trigger source initiates acquisition on all scaler channel detectors, which trigger a scaler acquisition, and the source reports as succeeded when all scaler channel measurements are complete. On Main, this solution as written (while not pretty) works.

However, the BioXAS scaler uses the Zebra trigger source. Outside of scans, the Zebra trigger source is uninitialized--there are no detectors attached to it. On Side, the acquisition hangs.

helfrij:
The content of this pull request has changed substantially since it was first reviewed. @iainworkman, would you mind taking a look at this again?

source/actions3/CLS/CLSSIS3820ScalerTriggerAction.cpp
```diff
@@ -0,0 +1,143 @@
...
+	if (!supportedReadMode(infoMode)) {
+		QString message = QString("There was an error triggering the scaler. The acquisition read mode provided (%1) is either invalid or not supported.").arg(infoMode);
+		AMErrorMon::alert(this, CLSSIS3820SCALERTRIGGERACTION_INVALID_READ_MODE, message);
+		setFailed(message);
+		return;
+	}
+
+	// Make connections.
+
+	startedMapper_->setMapping(triggerSource, triggerSource);
```
iainworkman: Are the signal mappers used here as a method of ensuring the trigger source which emits the signal ends up as an argument to the slots?

source/beamline/AMDetectorTriggerSource.cpp
```diff
@@ -121,11 +121,13 @@ void AMZebraDetectorTriggerSource::onDetectorArmed(QObject *detector)
 {
 	AMDetector *asDetector = qobject_cast<AMDetector*>(detector);
 	if(asDetector){
+		qDebug() << "Detector armed:" << asDetector->name();
```
iainworkman: Are these debugs still needed?

iainworkman:
Looks good to me.
Time Estimate: 15 minutes.

source/beamline/AMDetectorTriggerSource.cpp
helfrij: Nope, those QDebugs aren't needed anymore. Removed.

source/actions3/CLS/CLSSIS3820ScalerTriggerAction.cpp
helfrij: Yup, that's right. If we don't care about disconnecting from the trigger source once the action is complete, I guess those mappers are actually a little unnecessary. We could probably simplify that code by just connecting the trigger source's signals to the action's slots directly.

source/actions3/BioXAS/BioXASSIS3820ScalerDarkCurrentMeasurementAction.cpp
```diff
@@ -0,0 +1,93 @@
...
+			AMDetector *i1Detector = BioXASBeamline::bioXAS()->i1Detector();
+			if (i1Detector)
+				triggerSource->addDetector(i1Detector);
+
+			AMDetector *i2Detector = BioXASBeamline::bioXAS()->i2Detector();
+			if (i2Detector)
+				triggerSource->addDetector(i2Detector);
+
+			CLSSIS3820Scaler *scaler = BioXASBeamline::bioXAS()->scaler();
+			if (scaler)
```
dretrex: This is unnecessary because you already have a valid scaler pointer.  You would not be able to reach this code if that wasn't the case.

source/actions3/CLS/CLSSIS3820ScalerTriggerAction.h
```diff
@@ -0,0 +1,67 @@
...
+	void onSucceeded();
+
+protected:
+	/// Returns true if the given value is a supported detector read mode, false otherwise.
+	bool supportedReadMode(int mode) const;
+
+	/// This function is called from the Starting state when the implementation should initiate the action.
+	virtual void startImplementation();
+	/// For actions which support pausing, this function is called from the Pausing state when the implementation should pause the action.
+	virtual void pauseImplementation() { setPaused(); }
```
dretrex: Pedantically speaking, since `canPause()` and `canSkip()` are false, these could be left as {} since you will never enter the implementation methods.

source/actions3/BioXAS/BioXASSIS3820ScalerDarkCurrentMeasurementAction.cpp
helfrij: How do you mean?

source/ui/CLS/CLSSIS3820ScalerDarkCurrentWidget.cpp
```diff
@@ -74,23 +77,43 @@ void CLSSIS3820ScalerDarkCurrentWidget::setScaler(CLSSIS3820Scaler *newScaler)
...
+void CLSSIS3820ScalerDarkCurrentWidget::updateTimeBox()
 {
-	if (scaler_ && !scaler_->isScanning()) {
+	if (scaler_ && scaler_->isConnected()) {
+
+		timeBox_->blockSignals(true);
+		timeBox_->setEnabled(true);
+		timeBox_->setRange(0, INT_MAX);
+		timeBox_->setSuffix(" ms");
+		timeBox_->setValue(int(scaler_->dwellTime() * 1000));
```
dretrex: Out of curiosity, why not make the time be in seconds rather than milliseconds?

dretrex:
Time Estimate: 10 minutes

source/ui/CLS/CLSSIS3820ScalerDarkCurrentWidget.cpp
helfrij: I kept it as ms so that it matches the time units in the scaler view. Is it better to have seconds?

source/actions3/BioXAS/BioXASSIS3820ScalerDarkCurrentMeasurementAction.cpp
helfrij: Ah I see. I'll take that out.

source/ui/CLS/CLSSIS3820ScalerDarkCurrentWidget.cpp
dretrex: I'd say so.  Everything in AM uses seconds as its base.  Even the scaler model.  We should probably endeavour to get the scaler view in line as well.

helfrij: Okay, I'll make some changes.

helfrij:
Hey @dretrex. Looking over this pull request, I think there are a couple of things that I've just now noticed are missing. One of them is that while the Ge detector is never added to the trigger source as a detector or manager, it's pulse 3 block is still accepting input from pulse 1, meaning it's likely still triggering with every dark current acquisition. This will somehow have to be taken into account in the dark current action initialization.

dretrex:
Okay.  Let me know when the code is where you want it and I'll go over it again.

PR #2027, opened 2016-02-18
Pull Request for Issue1981: AMScanView tools crash

helfrij:
The crashes were caused by deleting tools that have already been deleted by the plot. Here, I've added checks to each view destructor so that only the non-selected tools are deleted, trusting the plot to take care of the rest.


helfrij:
Hey @XuetaoLiu-CLS, would you have a look?

XuetaoLiu-CLS:
Looks good to me. Agree with @dretrex that we need to check MPlot. Ideally, it should be the creator/owner (AMScanView for this case) to delete the instances instead of the users (the MPlot)

Time Estimate: 5 minutes

dretrex:
Looks good.
Time Estimate: 3 minutes

PR #2028, opened 2016-02-18
Pull Request for Issue484: Add Range Changed Signals to AMControl

iainworkman:
- Moved maximumValueChanged() and minimumValueChanged() up from AMPseudoMotorControl to AMControl.
- Added a function enableLimitsMonitoring() to AMPVControl, which connects to the correct PV fields depending on the PV type and listens for their value changed signals.
- Altered AMPVControl classes so that the low and high limits are stored in the class. Otherwise the valueChanged signals came through for the limits, but calling maximumValue or minimumValue still obtained the old value.

Some Notes:
- With this being a change for AMPVControl I was pretty rigorous in my testing relating to the limits:
    - I ensured that on construction, but before connection, the min and max values returned -1
    - I ensured that on connection the values which came through were those expected
    - I ensured that pushing values to both limit fields of a given PV produced the expected signal
This was done for all PV types except the waveform ones.
- I initially wanted to avoid the enableLimitsMonitoring() function, and just have it as an optional constructor parameter. However because all the PV classes are part of a hierarchy, calling a virtual function to set up the right PVs in a constructor would not have worked correctly.

### DIFF
--- a/source/beamline/AMPVControl.cpp
+++ b/source/beamline/AMPVControl.cpp
@@ -34,6 +34,9 @@ AMReadOnlyPVControl::AMReadOnlyPVControl(const QString& name, const QString& rea
 	wasConnected_ = false;
 	readPV_ = new AMProcessVariable(readPVname, true, this);
 
+	lowLimitPV_ = 0;
+	highLimitPV_ = 0;
+
 	connect(readPV_, SIGNAL(valueChanged(double)), this, SIGNAL(valueChanged(double)));
 	connect(readPV_, SIGNAL(alarmChanged(int,int)), this, SIGNAL(alarmChanged(int,int)));
 	connect(readPV_, SIGNAL(readReadyChanged(bool)), this, SLOT(onPVConnected(bool)));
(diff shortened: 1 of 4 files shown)


iainworkman:
Hey @helfrij, would you mind looking over this one?
Thanks.

source/beamline/AMPVControl.cpp
```diff
@@ -50,6 +56,17 @@ AMReadOnlyPVControl::AMReadOnlyPVControl(const QString& name, const QString& rea
 
 }
 
+
+void AMReadOnlyPVControl::enableLimitMonitoring()
```
helfrij: Just curious, is there ever a case when we wouldn't want PV controls to monitor limits?

iainworkman: If the PV does not have a HOPR LOPR (for read-only PVS) or a DRVH DRVL (for writable PVS) then there is nothing to monitor, and attempting to connect to these will result in the usual ``` AMProcessVariable: channel connect timed out for blah blah blah``` messages.

For example all the enum PV types (mbbi records) don't support anything but HOPR.

dretrex: Unfortunately, @helfrij, since limits are things you need to trust are implemented at the EPICS layer it is actually the reverse case where you'd only enable limit monitoring for controls you know have limits.

helfrij:
Looks good, @iainworkman!

Time Estimate: 10 minutes.

dretrex:
Looks good to me too.
Time Estimate: 3 minutes

PR #2042, opened 2016-02-19
Pull Request for Issue2036: Fixing valves moving state

helfrij:
First stab:
I've reimplemented updateMoving() in BioXASValves so that the valves never report as moving. The valves are currently reporting as moving erroneously because there are valves closed on Imaging while the Kohzu is vented.

Death blow:
Reversed first stab and just commented out the offending valves.

### DIFF
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -251,6 +251,7 @@ void BioXASAppController::setupUserInterface()
 	addComponentView(BioXASBeamline::bioXAS()->standardsWheel(), "Standards Wheel");
 	addComponentView(BioXASBeamline::bioXAS()->cryostatStage(), "Cryostat Stage");
 	addComponentView(BioXASBeamline::bioXAS()->filterFlipper(), "Filter Flipper");
+	addComponentView(BioXASBeamline::bioXAS()->detectorStageLateral(), "Ge 32-el Stage");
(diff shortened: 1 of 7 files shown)


helfrij:
Hey @iainworkman, would you take a look?

dretrex:
Won't this affect actions that open and close the valves?

helfrij:
I don't think so. The state that's associated specifically with move actions is the 'move in progress' state, different than the 'is moving' state. This change should just impact the display aspect of moving. This is definitely something I can double-check on Monday though.

iainworkman:
The code looks fine. There is still the question as to whether the is moving state always being false impacts on the general behaviour, but this will be resolved with testing.
Time Estimate: 2 minutes

helfrij:
Tested this today by opening the valves in Acquaman when some valves on the beamline where closed. All valves that are able to be opened are, but the ones that can't be opened (the ones that are closed while the Imaging mono is vented) fail. The valves value displays as 'Unknown' and the valve status button is red. The moving status aspect of this problem is solved.

However, the valves and beam status still report as 'Unknown' or 'Off'. A better solution for now may actually be to comment out adding those valves to the beam status. This way, the beam status can report as On and the valves as okay.

helfrij:
Made the change commenting out the valves.

dretrex:
Looks good.
Time Estimate: 1 minute

PR #2045, opened 2016-02-20
Pull Request for Issue2037: generalizing BioXASBeamStatus

helfrij:
- Modified BioXASBeamStatus, it no longer has specific setters for particular beamline components. Instead, components can be added as needed. 
- As only one state needs to be specified for the beam status (ie. the beam on value), this control is truly a bi-state control, whereas other controls that inherit from BioXASBiStateGroup (ie BioXASShutters, BioXASValves, etc) are really more like 'tri-state' controls. They need two values, an open and a closed value (plus implicit between/unknown). Created new class BioXASTriStateControl for this purpose, that inherits from and extends BioXASBiStateControl. Not sure if this is a good way to go long-term.
- Modified the BioXASBeamStatusBar in response to the changes in BioXASBeamStatus.
- Encountered problem with the beam status, where it would report as 'not connected' even though it had an updateConnected() implementation in an ancestor class. Since these reimplementations are identical to the one now in AMPseudoMotorControl by default, I removed them. This seems to have fixed the problem.


helfrij:
Hey @iainworkman, would you take a look at this request?

source/beamline/BioXAS/BioXASBiStateGroup.cpp
```diff
@@ -107,31 +107,31 @@ bool BioXASBiStateGroup::canStop() const
 	return result;
 }
 
-void BioXASBiStateGroup::updateConnected()
-{
-	bool connected = false;
+//void BioXASBiStateGroup::updateConnected()
```
iainworkman: Do you wish to leave this commented code in place, or can it be removed?

source/beamline/BioXAS/BioXASBiStateGroup.h
```diff
@@ -21,13 +21,13 @@ class BioXASBiStateGroup : public AMEnumeratedControl
 	virtual bool canStop() const;
 
 protected slots:
-	/// Updates the connected state.
-	virtual void updateConnected();
+//	/// Updates the connected state.
+//	virtual void updateConnected();
```
iainworkman: Can this commented code be removed?

source/ui/BioXAS/BioXASControlEditor.cpp
```diff
@@ -407,9 +407,14 @@ AMNumber BioXASControlEditor::getCalibratedDoubleValue()
 
 	return result;
 }
-
+#include <QDebug>
 void BioXASControlEditor::onContextMenuRequested(const QPoint &clickPosition)
 {
+	if (control_) {
+		qDebug() << "\n\n";
```
iainworkman: Can these debugs be removed?

iainworkman:
Looks good to me.
Time Estimate: 15 minutes.

source/beamline/BioXAS/BioXASBiStateGroup.cpp
helfrij: I'll remove it.

source/beamline/BioXAS/BioXASBiStateGroup.h
helfrij: Will remove this too.

source/ui/BioXAS/BioXASControlEditor.cpp
helfrij: Yup, these are now removed too.

dretrex:
Looks good.
Time Estimate: 5 minutes

PR #2057, opened 2016-02-25
Pull Request for Issue2055: Moving BioXAS scan initialization and cleanup actions to BioXASBeamline

helfrij:
- Added public virtual getters to AMBeamline for scan initialization action, scan cleanup action, and building a scan. 
- Added calls to these methods to AMGenericStepScanController. 
- Reimplemented them in BioXASBeamline, moving the contents of the initialization actions, cleanup actions, and buildScan to this class from BioXASXASScanActionController, making a few small simplifications.


helfrij:

iainworkman:
The code looks good. The only issue I have with this is the movement of scan initialization to the AM Beamline level. This gets a little confusing if a given beamline has multiple scan types which all require different initialization steps.

helfrij:
I agree, that could be potentially confusing. It's possible to cast the given configuration to a particular type and check to see what's valid, and create the actions appropriate for that scan type from there, but it isn't as pretty as keeping everything separate.

dretrex:
@iainworkman @helfrij I think the more important thing about this is that even if different scans might have different initializations, in general there will be some common basic elements (such as configuring the scaler) which would be shared across scan types.  As always, inside of your specific scan controller, you can always still reimplement to add to, or completely ignore, the standard actions provided from the beamline.

iainworkman:
Maybe 'createPrepareForScanningActions()' might be a better name? That way
it clarifies the responsibility more. The beamline is responsible for
knowing what is required to put it in the right state for scanning, the
scan can then add to this if necessary.

dretrex:
We will need both initialization and cleanup actions.  Maybe,

```c++
createStandardInitializationActions();
createStandardCleanupActions();
```

Thoughts?  Maybe throw `Scan` in there somewhere?

source/acquaman/AMGenericStepScanController.cpp
```diff
@@ -71,88 +68,15 @@ void AMGenericStepScanController::onScanTimerUpdate()
...
 
 AMAction3 * AMGenericStepScanController::createCleanupActions()
 {
-	return 0;
+	return AMBeamline::bl()->createScanCleanupAction(configuration_);
 }
 
 void AMGenericStepScanController::buildScanControllerImplementation()
 {
-	QList<AMXRFDetector *> xrfDetectors;
```
dretrex: I'm not entirely sure about this part.  This is already quite general and didn't need to be moved as it wasn't getting in anyones way for anything.  Maybe I missed something?

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
```diff
@@ -247,233 +35,5 @@ void BioXASXASScanActionController::createScanAssembler()
...
-					normalizedRegion->setInputDataSources(QList<AMDataSource *>() << newRegion << normalizationSource);
-					normalizedRegion->setDataName(newRegion->name());
-					normalizedRegion->setNormalizationName(normalizationSource->name());
-
-					scan_->addAnalyzedDataSource(normalizedRegion, newRegion->name().contains(edgeSymbol), !newRegion->name().contains(edgeSymbol));
-				}
-			}
-		}
-	}
+	BioXASBeamline::bioXAS()->buildScan(scan_);
```
dretrex: This part I agree with @iainworkman where the different scan types can become an issue.

dretrex:
I think we might need to sit down and brainstorm how we want these to look like in the future.  I like the idea of generic scan initialization and cleanup, but typically the building of a scan would still live in the scan controller.

dretrex:
Time Estimate: 10 minutes

source/acquaman/AMGenericStepScanController.cpp
helfrij: The scan controller implementation on BioXAS contains the Zebra setup. This needs to be something that's done in BioXASBeamline.

dretrex: Hmm... that's a tricky one.  I'll have to think about that.  @iainworkman @XuetaoLiu-CLS @davidChevrier feel free to provide suggestions as well.  More brains on the matter would be helpful.

helfrij:
I think the word 'scan' is already included in the beamline action getters, they're called `createScanInitializationAction()` and `createScanCleanupAction()`. Unless I'm missing something, @dretrex?

The problem this pull request attempts to solve is to have the Zebra setup that happens in the BioXASXASScanActionController::buildScanControllerImplementation() happen for both the XAS scans and the commissioning tool scans. The `buildScan()` part needs to happen in the beamline class, for BioXAS anyways, because there are no actions for completing these steps. I don't think these changes stop other beamlines from implementing scan initialization, cleanup, and building the scan controller as they already do in their own classes.

Having the commissioning tool back up and running is a pretty high priority task for Side right now. Would having a sit-down discussion about these actions be able to happen soon? If not, perhaps this is best dealt with as a separate issue?

dretrex:
I would imagine the sit down discussion wouldn't happen until after I got back I'd imagine, which would make it late next week at the earliest.  

The main issue right now is that the changes and the addition of buildScan() are in AMBeamline, not BioXASBeamline.  Although technically not breaking anything, we should take the time to ensure that the pieces we add to the AM or CLS level have the broad implications we desire rather than solving an individual beamline issue.  For the time being, I would put the buildScan piece inside BioXASBeamline and, if necessary, subclass the commissioning tool to use the method.  We can then make separate issues that deal with desired code design and then refactor the BioXAS commissioning tool to use the new code.

helfrij:
I'm hesitant to subclass AMGenericStepScan for BioXAS now--won't that require a database upgrade later, if/when it's removed? If the real problems have to do with those Zebra initialization steps specifically, then perhaps the best solution is to investigate turning those into actions. Then they can happen in scan initialization, eliminating the need to have the `buildScan()` stuff in any beamline class.

dretrex:
You don't need to subclass AMGenericStepScanConfiguration, just the controller.  That doesn't require a database upgrade.

Configuring the Zebra as actions is also a possibility.

helfrij:
Oh, okay! Wait, don't you need a specific configuration to create and return a specific controller?

dretrex:
Ugh.  Yes, I believe you're correct.

dretrex:
This Zebra is creating quite the debacle.  I think the way that will get you to fixing the issue the quickest is making the ZebraConfigurationAction.  I'm not having any epiphanies for the other options at the moment.

helfrij:
I think so too. I'll create a new issue and start looking into how tricky that would be.

dretrex:
Okay.  Sounds good.

iainworkman:
Time Estimate: 10 minutes.

helfrij:
I've removed all of the changes that dealt with migrating `buildScanControllerImplementation()` to AMBeamline for both the AMGenericStepScanController and BioXASXASScanActionController. The only important changes that should remain in this branch now are the ones related to the scan initialization and cleanup actions. The specifics of the Zebra initialization should be handled in a different issue, with the current idea being in #2063.

dretrex:
Okay, sounds good @helfrij.  Thanks for spending time on this.  I'm sure we'll revisit this soon to really iron out a solid solution that we're all happy with.  I'll merge this in now.

Issue #2052, opened 2016-02-24
Allow AMPVControl Classes to Push Maximum and Minimum Values to the EPICS layer


iainworkman:
With the work done in #484 the range values in EPICS now populate up to the AMPVControl classes. However there is still a requirement to allow changes to the range at the AM level to be pushed down to EPICS. There will need to be some design work done here, in order to ensure that this only happens in cases where we want it (eg. only if the limit values exist in the EPICS layer, or when we definitely want the limits to persist).

This will close the loop on passing around these values, and mean that work can begin or tackling the #1569

dretrex:
This could likely be coupled in with the `allowLimitsMonitoring()` from #484 because then we'll only want to enable limits monitoring when it makes sense.  This way, everything about limits are tied to the same thing since I don't think it makes sense to talk about the direction of limits separately.

PR #2061, opened 2016-02-25
Pull Request for Issue2060: Modifying editors to listen for AMControl min/max value changes

helfrij:
- Modified AMReadOnlyPVControl and subclasses. The minimumValueChanged() and maximumValueChanged() signals should now be emitted whenever the low and high limit values have changed.
- Modified AMExtendedControlEditor::setControl(...) to listen for the the minimumValueChanged(...) and maximumValueChanged(...) signals.
- Modified BioXASControlEditor to do the same.


helfrij:
Hey @iainworkman, I know you worked on this recently. Would you take a look at this request?

iainworkman:
Looks good to me.
Time Estimate: 5 minutes.

source/beamline/AMPVControl.cpp
```diff
@@ -89,25 +92,40 @@ void AMReadOnlyPVControl::onReadPVError(int errorCode) {
 	}
 }
 
-
-void AMReadOnlyPVControl::onLowLimitPVValueChanged(double newLowLimit)
+void AMReadOnlyPVControl::setLowLimitValue(double newLowLimit)
 {
-	lowLimitValue_ = newLowLimit;
-	emit minimumValueChanged(lowLimitValue_);
+	if (lowLimitValue_ != newLowLimit) {
```
helfrij: There may be a case when we would want the signals to be emitted here but the new value fails this if-check--when the actual limit is equivalent to the default value. It may make sense to remove these if-checks from AMReadOnlyPVControl::setLowLimitValue() and AMReadOnlyPVControl::setHighLimitValue(). Then the signals would be emitted every time a value is set.

iainworkman: Originally that's the tack I'd taken. I thought that if ahigher level class wants to invite it, they can but they'll still receive the signal.

source/beamline/AMPVControl.cpp
```diff
@@ -188,8 +209,8 @@ void AMPVControl::onReadPVInitialized()
 }
 
 void AMPVControl::onWritePVInitialized() {
-	lowLimitValue_ = writePV_->lowerControlLimit();
-	highLimitValue_ = writePV_->upperControlLimit();
+	setLowLimitValue(writePV_->lowerControlLimit());
```
dretrex: Is this to ensure that the limit signals are emitted?

source/beamline/AMPVControl.cpp
```diff
@@ -459,8 +483,8 @@ void AMPVwStatusControl::onReadPVInitialized()
 }
 
 void AMPVwStatusControl::onWritePVInitialized() {
-	lowLimitValue_ = writePV_->lowerControlLimit();
-	highLimitValue_ = writePV_->upperControlLimit();
```
dretrex: This change will make AMPVwStatusAndUnitConversionControl not work because it only emits the signal in setLow/HighLimitValue and doesn't set the value..

source/beamline/AMPVControl.h
```diff
@@ -132,12 +132,16 @@ class AMReadOnlyPVControl : public AMControl {
 
 protected slots:
 	/// Handles the low limit PV changing. Converts the limit before forwarding the signal on.
-	virtual void onLowLimitPVValueChanged(double newLowLimit);
+	virtual void setLowLimitValue(double newLowLimit);
```
dretrex: Pure semantic question, but why change from onLimitChanged to setLimit?  From what I can tell this is being primarily called via signals which would use the onXChanged style.

source/ui/beamline/AMExtendedControlEditor.cpp
```diff
@@ -456,6 +462,20 @@ QSize AMExtendedControlEditor::sizeHint() const{
 	return newHint;
 }
 
+void AMExtendedControlEditor::onControlMinimumValueChanged()
+{
+	if (control_) {
```
dretrex: Will this cause any issues if `enableLimitsMonitoring()` is not set since you explicitly call it in the constructor?

dretrex:
Time Estimate: 10 minutes

source/beamline/AMPVControl.cpp
helfrij: Yes.

source/beamline/AMPVControl.h
helfrij: The implementation isn't different, but the usage is. Previously, the values were changed in `onWritePVInitialized` type slots, without the signals being emitted. Calling `onLowLimitPVValueChanged` in these cases isn't really representative of what's happening.

source/ui/beamline/AMExtendedControlEditor.cpp
helfrij: Ah, you're thinking because `enableLimitsMonitoring` is virtual?

dretrex: No, I'm thinking that since controls won't, in general, be emitting the min/max changed singals I'm worried that the default behaviour/values will cause issues in the editor.  Such as only allowing -1 or 0 or some such.

source/beamline/AMPVControl.h
dretrex: Aside from the `onWritePVInitialized()` calls, the method is typically called via the signal/slot mechanism from what I can tell.

source/beamline/AMPVControl.cpp
helfrij: I'm not sure I understand what you mean here, @dretrex. The low and high limit values look like they're set when the write PV reports as initialized.

dretrex: Unless I'm missing something here, you call `setLowLimitValue()` here and inside AMPVwStatusAndUnitConversionControl those methods only emit a signal.  They overload the standard call and only emit the signals and not update the `lowLimitValue_` and `highLimitValue_` members.

dretrex: See line 726/731 of AMPVControl.cpp I think.

helfrij: Ah I see. There should be a value set there. Will fix that.

source/ui/beamline/AMExtendedControlEditor.cpp
helfrij: That's the problem this pull request attempts to solve, it's the present behavior. With the changes to AMPVControl, controls should emit their min and max values when they connect. But in case the editor is created after they are connected, the editor should be able to ask a control for its min/max values.

helfrij:
Yup. Most AMPVControls on BioXAS are MAXvMotors, and the default minimum and maximum values they report when they connect are -DBL_MAX and DBL_MAX. These are the values the editors receive when the control connects with these changes. I would be interested to know if this is true for controls on other beamlines, and of other types. If this is a default behavior, then I think it makes more sense to let controls report their min/max values, rather than trusting in AM that -DBL_MAX and DBL_MAX are the limit values.

iainworkman:
@dretrex Not sure I understand. Do you mean pushing limit values on PV records even if they don't have them (eg. -DBL_MAX, DBL_MAX)?

iainworkman:
Jessie and I spent a little time on this. It looks like we pinned down a bit more of the cause of the underlying issue. I'm going to go through the pull request again, and make sure it should solve the causes we identified.

dretrex:
@iainworkman, no I meant if they were changed to some sort of real and meaningful values they should be propagated.  It was just a thought, but it isn't necessarily a good idea because we don't know if that will have other consequences since all things seem to have the assumption that nothing has limits.

dretrex:
I'm glad you two got a handle on the issue.

source/ui/beamline/AMExtendedControlEditor.cpp
iainworkman: When Jessie and I spoke today it seemed that this was the best way to ensure consistent behaviour.
The control emits the min/max changed signals once when the values are first initialized. If enableLimitsMonitoring is called on a PV control it will after that emit the signals as often as it needs to, however the editors can always expect the signal once (assuming the control is not connected when passed in).

The reason for this is that the min and max values did not always seem to be correct by the time AMExtendedControlEditor's onConnectedChanged slot was being invoked.

dretrex: This seems like a fundamental limitation then because if you pass a control to a control editor after it has connected you'd be left in limbo.  Wouldn't having the means of checking be a more robust solution?

iainworkman:
@dretrex There is an issue for allowing the limits to be propagated into the EPICs layer in #2052. However this would only be for those that had already enabled limits monitoring.

dretrex:
Right.

source/ui/beamline/AMExtendedControlEditor.cpp
iainworkman: When you pass it in after it has connected the correct value is obtained from the control_->minimumValue()

essentially the logic we'd end up with (simplified) in the editors is:
```
setControl(control):
...
    if (control is connected)
        maxValue_ = control->maxValue()

    connect(control, maxValueChanged(double), this, onControlMaxValueChanged(double))
...
onControlMaxValueChanged(newMaxValue):
    maxValue_ = newMaxValue
```

dretrex:
Okay!  After an incredibly thorough discussion, I believe we are all on the same page and the presented solution takes into account all of the appropriate things.  Unless there is any last changes you'd like to make @helfrij, I'll merge this in.  Thanks for your patience!

helfrij:
No further changes. Thanks @iainworkman @dretrex!

PR #2070, opened 2016-02-26
Pull Request for Issue2063: Zebra initialization actions

helfrij:


helfrij:
Hey @dretrex. I started looking into what we would need in order to have actions to perform some of the Zebra initialization and cleanup stuff. This pull request is what I've done this afternoon, and I was wondering if you could take a look when you have time.

So far, I've created an action that modifies the detectors through the Zebra interface, and I think it could work well. There may be a big hiccup with framing the problem this way, though, because we can't treat detector managers the same as detectors in actions--there's no such thing as a manager info. Ooops.

I'm open to suggestions if you happen to see a way to make this work.

PR #2082, opened 2016-03-01
Pull Request for Issue1992: BioXASBeamlineComponent to inherit from AMConnectedControl.

helfrij:
Modified BioXASBeamlineComponent to inherit from AMConnectedControl, instead of AMControl. Removed the class content related to getting and tracking connected state information, as this is taken care of in the new ancestor.


helfrij:
Hey @iainworkman, would you take a look at this request?

I noticed a potential problem with `AMConnectedControl`, it looked like there was a kind of connected state 'circular reference' previously: `updateConnected()` would call `isConnected()` and set the state to that value. Which would be its current state! If this was your trouble with the SGMVATValve, you would have broken that circular chain by reimplementing `isConnected()`. If you're happy with that solution, that's cool, but it may no longer be necessary.

iainworkman:
Looks good to me.
Time Estimate: 5 minutes.

dretrex:
Looks good to me too.
Time Estimate: 3 minutes

PR #2092, opened 2016-03-04
Pull Request for Issue2091: Duplicate regions of interest

helfrij:
Added a new method containsRegionOfInterest() to BioXASBeamline. Successfully used this method when a user configuration is loaded to prevent a ton of duplicate ROIs from being added to the detector and scan configuration.

### DIFF
--- a/source/beamline/AMXRFDetector.cpp
+++ b/source/beamline/AMXRFDetector.cpp
@@ -322,10 +322,9 @@ void AMXRFDetector::addRegionOfInterest(const AMEmissionLine &emissionLine)
 
 	addRegionOfInterest(region);
 }
-#include <QDebug>
+
 void AMXRFDetector::addRegionOfInterest(AMRegionOfInterest *newRegionOfInterest)
 {
-	qDebug() << "AMXRFDetector: adding ROI to detector.";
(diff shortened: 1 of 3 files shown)


helfrij:

dretrex:
Hey @helfrij, I know that what you've done technically fixes the problem, but I'd be interested to sit down with you sometime to figure out why it's adding duplicates in the first place because this could be happening on other beamlines.

source/beamline/AMXRFDetector.cpp
```diff
@@ -322,9 +322,10 @@ void AMXRFDetector::addRegionOfInterest(const AMEmissionLine &emissionLine)
 
 	addRegionOfInterest(region);
 }
-
+#include <QDebug>
```
XuetaoLiu-CLS: Do we want to keep this?

helfrij: No, I don't think so. I'll take it out.

XuetaoLiu-CLS:
Looks good to me. As @dretrex mentioned, this might not fix the root cause of the duplicated ROI. Hopefully we can find it.

Time Estimates: 5 minutes.

dretrex:
Looks good.  We will need to make an issue to look into the possibility that this is a more general problem.
Time Estimate: 2 minutes

PR #2097, opened 2016-03-05
Pull Request for Issue1993: Adding XAS detectors to BioXAS scan configuration by default

helfrij:
Added new detector set to BioXASBeamline, xasDetectors. This set is for the list of detectors that we'd like to be added to XAS scans by default once the detector set connects and whenever there are changes to the set contents. Modified BioXASSideBeamline and BioXASMainBeamline to add I0, I1, I2, and any Ge detectors to the set, though this can probably be refactored/simplified later. Added connect statement to BioXASAppController that listens to the connected state of the beamline's XAS detectors. The slot clears the configuration detectors and adds all valid, connected detectors from the set every time the set reports a connected change. Awaiting testing.


helfrij:
Better solution implemented as part of #2120.

PR #2103, opened 2016-03-07
Pull Request for Issue2101: removing fast shutter from Zebra trigger chain

helfrij:
- Added the ability to set a 'preferred' input value for Zebra logic blocks and pulse controls.
- Added some default 'preferred' input values for some Zebra controls in BioXASSideBeamline, including that the fast shutter should be disconnected from Pulse 1/scanning pulse.
- Added fast shutter getter to BioXASBeamline. This allows for the fast shutter to be accessible in scan initialization and cleanup actions.
- Added fast shutter initialization (opening shutter) and cleanup (closing shutter) to scan initialization and cleanup actions.

### DIFF
--- a/source/beamline/BioXAS/BioXASBeamline.cpp
+++ b/source/beamline/BioXAS/BioXASBeamline.cpp
@@ -139,6 +139,19 @@ AMAction3* BioXASBeamline::createScanInitializationAction(AMGenericStepScanConfi
 		if (standardsWheelInitialization)
 			initializationAction->addSubAction(standardsWheelInitialization);
 
+		// Initialize the fast shutter.
+
+		AMListAction3 *fastShutterInitialization = 0;
+		BioXASFastShutter *fastShutter = BioXASBeamline::bioXAS()->fastShutter();
+
+		if (fastShutter) {
+			fastShutterInitialization = new AMListAction3(new AMListActionInfo3("BioXAS fast shutter initialization", "BioXAS fast shutter initialization"));
+			fastShutterInitialization->addSubAction(AMActionSupport::buildControlMoveAction(fastShutter, BioXASFastShutter::Open));
+		}
+
+		if (fastShutterInitialization)
+			initializationAction->addSubAction(fastShutterInitialization);
+
 		// Complete action.
 
 		result = initializationAction;
--- a/source/beamline/BioXAS/BioXASBeamline.cpp
+++ b/source/beamline/BioXAS/BioXASBeamline.cpp
@@ -166,6 +166,19 @@ AMAction3* BioXASBeamline::createScanCleanupAction(AMGenericStepScanConfiguratio
 
 	AMListAction3 *result = new AMListAction3(new AMListActionInfo3("BioXAS scan cleanup actions", "BioXAS scan cleanup actions"), AMListAction3::Parallel);
 
+	// Create fast shutter cleanup actions.
+
+	AMListAction3 *fastShutterCleanup = 0;
+	BioXASFastShutter *fastShutter = BioXASBeamline::bioXAS()->fastShutter();
+
+	if (fastShutter) {
+		fastShutterCleanup = new AMListAction3(new AMListActionInfo3("BioXAS fast shutter initialization", "BioXAS fast shutter initialization"));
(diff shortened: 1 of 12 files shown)


helfrij:
Hey @dretrex. I've done a little work on this, and I've run into a bit of a problem. The fast shutter right now is triggered by soft input 2 or by pulse 1 (during a scan). Taking out the scan component means that the shutter is only triggered using soft input 2. I think this means that the fast shutter only triggers/stays open when the soft input 2 stays high and closes automatically afterwards... is that right?

dretrex:
That is correct, however that isn't an issue because Soft Input 2 doesn't get pulled down automatically.

helfrij:
Soft input 2 doesn't get pulled down automatically... I'm not sure I understand what that means.

dretrex:
The .HIGH field isn't set so it will stay high for as long as is required.  Or am misunderstanding what the problem is?

helfrij:
How do I check if that's the case?

dretrex:
It should be either -1 or 0.  I can't recall exactly off the top of my head.  You can check the current value for .HIGH for soft input 2 or you can check soft input 4.

helfrij:
All soft input .HIGH fields are either 0.01 or 0.1 right now. I'll try the 0 and -1 values and see if that makes a difference. Thanks!

helfrij:

source/beamline/BioXAS/BioXASBeamline.cpp
```diff
@@ -139,6 +139,19 @@ AMAction3* BioXASBeamline::createScanInitializationAction(AMGenericStepScanConfi
...
 			initializationAction->addSubAction(standardsWheelInitialization);
 
+		// Initialize the fast shutter.
+
+		AMListAction3 *fastShutterInitialization = 0;
+		BioXASFastShutter *fastShutter = BioXASBeamline::bioXAS()->fastShutter();
+
+		if (fastShutter) {
+			fastShutterInitialization = new AMListAction3(new AMListActionInfo3("BioXAS fast shutter initialization", "BioXAS fast shutter initialization"));
+			fastShutterInitialization->addSubAction(AMActionSupport::buildControlMoveAction(fastShutter, BioXASFastShutter::Open));
```
iainworkman: Just out of interest, do these calls to buildControlMoveAction generate a compile warning because of implicit cast from the enum type to double?

source/beamline/BioXAS/BioXASSideBeamline.cpp
```diff
@@ -502,11 +503,36 @@ void BioXASSideBeamline::setupComponents()
...
 
 	BioXASZebraSoftInputControl *softIn3 = zebra_->softInputControlAt(2);
 	if (softIn3)
-		softIn3->setTimeBeforeResetPreference(0.1);
+		softIn3->setTimeBeforeResetPreference(0.1); // The zebra has a good chance of missing triggers if this value is lower (definitely at 0.01).
+
+	BioXASZebraLogicBlock *fastShutterBlock = zebra_->andBlockAt(0);
+	if (fastShutterBlock) {
+		fastShutterBlock->setInputValuePreference(0, 61); // The fast shutter can be triggered using soft input 2 (#61).
+//		fastShutterBlock->setInputValuePreference(1, 52); // The fast shutter can be triggered using pulse 1 (#52).
```
iainworkman: Can this comment be removed?

iainworkman:
Couple of questions, but otherwise everything looks good.
Time Estimate: 5 minutes.

source/beamline/BioXAS/BioXASBeamline.cpp
helfrij: Not that I've noticed. I would guess there would be errors if the cast was in the other direction, though, from double to int/enum.

source/beamline/BioXAS/BioXASSideBeamline.cpp
helfrij: I would like to keep it for now. If we wanted to reintroduce the fast shutter into the trigger chain, this line would be uncommented.

dretrex:
Looks good.
Time Estimate: 5 minutes

PR #2109, opened 2016-03-08
Issue2022 zebra refactor -- creating Side-specific Zebra

helfrij:
- Moved the Zebra trigger source to be a class variable inside BioXASZebra.
- Created Side-specific Zebra subclass. This class is responsible for initializing the Zebra, and matching adding/removing detectors to making the appropriate logic block and pulse block connections.


helfrij:
Since #2023 has been merged into master, work done in this branch should be merged into master too. New request: #2132.

PR #2118, opened 2016-03-09
Pull Request for Issue2098: Diode/PIPS/Lytle detector dark current updates

helfrij:
Made a few small modifications to the BioXAS-specific scaler channel view. Added a detector as a class variable, the scaler channel's detector. This variable is updated every time the scaler channel reports a change in detector, as is what happens when the diode/PIPS/Lytle detectors are added/removed from the beamline. When the detector reports a change in the dark current value or state, the dark current display is updated.

### DIFF
--- a/source/ui/BioXAS/BioXASSIS3820ScalerChannelView.cpp
+++ b/source/ui/BioXAS/BioXASSIS3820ScalerChannelView.cpp
@@ -1,89 +1,70 @@
 #include "BioXASSIS3820ScalerChannelView.h"
+#include "beamline/CLS/CLSSIS3820Scaler.h"
 #include "beamline/CLS/CLSKeithley428.h"
 #include "ui/beamline/AMExtendedControlEditor.h"
-#include "ui/CLS/CLSSIS3820ScalerView.h"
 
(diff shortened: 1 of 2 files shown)


helfrij:

source/ui/BioXAS/BioXASSIS3820ScalerChannelView.cpp
```diff
@@ -1,89 +1,70 @@
...
 {
 	// Initialize member variables.
 
+	detector_ = 0;
+
 	biasEnabledVisible_ = false;
 	biasVisible_ = false;
 	darkCurrentVisible_ = false;
 
-	biasEnabledEditor_ = 0;
```
dretrex: These initializations should probably still be here I think.

dretrex:
One comment, otherwise looks good.
Time Estimate: 3 minutes

source/ui/BioXAS/BioXASSIS3820ScalerChannelView.cpp
helfrij: Ok! I've added them back.

PR #2120, opened 2016-03-10
Pull Request for Issue2054: exporting spectra data for Ge detectors

helfrij:
- Added getter to AMXRFDetector for the list of spectra controls.
- Added a map to BioXASBeamline, that maps detectors to the set of detector elements.
- Modified the add/remove/clearGeDetectors methods to also consider the detector's elements.
- Added two sets of detectors to BioXASBeamline, one for the detectors that should be added to scan configurations by default, and one for the detectors that should be shown as options (eg. scientific detectors + control emulators).

- setAccessAsDouble(), BioXASBeamline creating element emulators.


helfrij:
@dretrex, a note so I don't forget. As written now, the changes in this pull request cause the appropriate detector elements to be added to the BioXAS XAS scan configuration, and the proper number of files are exported after a scan. However, the files all seem to contain 0s.

dretrex:
Looks good.
Time Estimate: 5 minutes

PR #2125, opened 2016-03-12
Pull Request for Issue2124: Fixing BioXAS beam status view

helfrij:
- Created BioXASControlButton, a subclass of AMToolButton. This changes button color based on given control values.
- Created BioXASButtonBar, a widget that manages a collection of QAbstractButtons.
- Created BioXASControlButtonBar, a subclass of BioXASButtonBar that manages a collection of BioXASControlButtons based on supplied controls.
- Created BioXASBeamStatusButtonBar, a subclass of BioXASControlButtonBar that manages a collection of control buttons specific to the beam status.
- Modified BioXASBeamStatusView to use the new BioXASControlButtonBar, and to manage the selected component widgets.
- Removed old classes: BioXASBeamStatusBar, BioXASControlEditorBar. No longer used/necessary.

This fix ended up being a kind of refactor of this view. It doesn't look much different, but it is a little more easily extended:
- could show just the button bar in the persistent view.
- could develop similar view for the utilities.


source/ui/BioXAS/BioXASControlButton.cpp
```diff
@@ -0,0 +1,69 @@
...
+	control_ = 0;
+
+	greenValueSet_ = false;
+	greenValue_ = 0;
+
+	// Current settings.
+
+	setControl(control);
+
+	refresh();
```
XuetaoLiu-CLS: we called `refresh()` in `setControl`. I think we don't need to call this again.

source/ui/BioXAS/BioXASControlButton.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+	QColor color = QColor();
+
+	if (control_) {
+		if (greenValueSet_ && control_->isConnected() && qFuzzyCompare(control_->value(), greenValue_))
+			color = QColor(Qt::green);
+		else if (greenValueSet_ && control_->isConnected())
+			color = QColor(Qt::red);
+		else if (greenValueSet_)
+			color = QColor(Qt::yellow);
```
XuetaoLiu-CLS: what's the default color if `greenValueSet_` is `false`?

XuetaoLiu-CLS:
Looks good.

Time Estimates: 10 minutes

source/ui/BioXAS/BioXASControlButton.cpp
helfrij: It displays the 'default' color--the button has no color. Just looks like a regular button.

XuetaoLiu-CLS: git it. Thx

source/ui/BioXAS/BioXASControlButton.cpp
helfrij: That call to `refresh()` isn't executed if the control given in the constructor is 0. For some widgets, this is important, as `refresh()` helps make sure the view is consistent with the current information. For this particular case, for this widget, you may be right--it might not make a difference.

dretrex: Are you wanting to keep this refresh call or remove it?

helfrij: I'd like to keep it, if that's okay.

dretrex: It's alright with me.

dretrex:
Time Estimate: 5 minutes

PR #2144, opened 2016-03-17
Pull Request for Issue2128: Adding beam status bar to persistent view.

helfrij:
Added an instance of the BioXASBeamStatusButtonBar to BioXASPersistentView. The view now emits a signal when the selected control in the button bar changes. Modified the app controller to listen for this signal and change the current pane to the main beam status view in response. It additionally sets the current control to the newly selected control too.

### DIFF
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -154,6 +154,23 @@ void BioXASAppController::onRegionOfInterestBoundingRangeChanged(AMRegionOfInter
 		xasConfiguration_->setRegionOfInterestBoundingRange(region);
 }
 
+void BioXASAppController::goToBeamStatusView(AMControl *control)
+{
+	if (beamStatusView_) {
+
+		// Set the given control as the view's selected control.
+
+		beamStatusView_->setSelectedComponent(control);
+
+		// Set the beam status pane as the current pane.
+
+		QWidget *windowPane = viewPaneMapping_.value(beamStatusView_, 0);
+
+		if (windowPane)
+			mw_->setCurrentPane(windowPane);
+	}
+}
+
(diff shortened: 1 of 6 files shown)


helfrij:

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -154,6 +154,23 @@ void BioXASAppController::onRegionOfInterestBoundingRangeChanged(AMRegionOfInter
 		xasConfiguration_->setRegionOfInterestBoundingRange(region);
 }
 
+void BioXASAppController::goToBeamStatusView(AMControl *control)
```
XuetaoLiu-CLS: I feel this function name is not straight-forward. Maybe we should think about rename it?

XuetaoLiu-CLS:
Looks good to me.

Time Estimate: 5 minutes

source/application/BioXAS/BioXASAppController.cpp
helfrij: Okay! What's confusing about it?

XuetaoLiu-CLS: maybe sth like updateBeamStatusViewControl()?

helfrij: That name does describe part of what happens in the `goToBeamStatusView(AMControl*)`, the part where the selected component of the beam status view is set. But it doesn't describe the part where the current main window pane is set to the beam status view.

XuetaoLiu-CLS: sounds reasonable

dretrex: I would say this function name is descriptive enough.  If we think there is a better name at a later date we can change it then.

dretrex:
Time Estimate: 5 minutes

PR #2147, opened 2016-03-17
Pull Request for Issue2143: Creating BioXAS-specific commissioning tool

helfrij:
- Created new class BioXASGenericStepScanConfiguration, a subclass of AMGenericStepScanConfiguration.
- Created new class BioXASGenericStepScanController, a subclass of AMGenericStepScanController. Added the code to `buildScanControllerImplementation()` for initializing the Zebra trigger source.
- Added scan configuration and view to BioXASAppController.
- Added mechanism to BioXASBeamline and BioXASAppController so that default detectors can be added to the configuration as soon as they're connected. It remains commented out for now, though, as it appears AMGenericStepScanConfigurationView may not support showing 'pre-loaded' detectors. There is an existing issue to tackle implementing this behavior though (#2148).


helfrij:

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -783,9 +814,11 @@ void BioXASAppController::setupGenericStepScanConfiguration(AMGenericStepScanCon
 
 		// Set scan detectors.
 
-		AMDetector *i0Detector = BioXASBeamline::bioXAS()->i0Detector();
-		if (i0Detector)
-			configuration->addDetector(i0Detector->toInfo());
+//		connect( BioXASBeamline::bioXAS()->defaultGenericScanDetectors(), SIGNAL(connected(bool)), this, SLOT(updateGenericScanConfigurationDetectors()) );
```
iainworkman: Can these comments be removed or would you like to hang on to them?

source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
```diff
@@ -132,7 +132,6 @@ AMGenericStepScanConfigurationView::AMGenericStepScanConfigurationView(AMGeneric
 	moreLayout->addStretch();
 	moreLayout->addLayout(layout);
 	moreLayout->addLayout(detectorsAndI0Layout);
-	moreLayout->addWidget(i0ComboBox_);
```
iainworkman: So...was the i0ComboBox_ previously added to the layout twice?

source/beamline/BioXAS/BioXASBeamline.h
```diff
@@ -196,10 +196,15 @@ class BioXASBeamline : public CLSBeamline
 	/// Returns the detector for the given control, if one has been created and added to the control/detector map.
 	AMBasicControlDetectorEmulator* detectorForControl(AMControl *control) const;
 
-	/// Returns the set of default detectors used in XAS scans.
-	AMDetectorSet* defaultScanDetectors() const { return defaultScanDetectors_; }
-	/// Returns the set of detectors to use as options in scans.
-	AMDetectorSet* scanDetectorsOptions() const { return scanDetectorsOptions_; }
+	/// Returns the set of default detectors added to XAS scans, a subset of defaultXASScanDetectorOptions.
```
iainworkman: This subset status of default detectors vs. detector options in managed manually in specific functions for each detector, right? (eg. addGe32Detector() etc.)

iainworkman:
Couple of comments, but otherwise everything looks good.
Time Estimate: 15 minutes.

source/application/BioXAS/BioXASAppController.cpp
helfrij: I'd like to hang onto them, if that's okay. They are commented right now to avoid issues related to #2148 and I plan to use them when that issue is dealt with.

source/ui/acquaman/AMGenericStepScanConfigurationView.cpp
helfrij: Most of the time, the box would show up as normal in the configuration view, but every once in a while I noticed it showed up in kind of a funny place. It looks like it was added to two different layouts, and that was causing the funny behavior. I kept the one that I think makes sense (next to the I0 label), this is the one that was removed.

source/beamline/BioXAS/BioXASBeamline.h
helfrij: Yes. For each scaler channel and for each Ge detector, handled in BioXASBeamline. I'm not sure this is the best way to go about it--there's a lot of duplicate information.

dretrex:
Looks good.
Time Estimate: 5 minutes

Issue #2127, opened 2016-03-12
Refactoring BioXAS 'Group' classes


helfrij:
The BioXAS applications have these control 'group' classes that report a single value on behalf of all group members--eg. BioXASBiStateGroup, BioXASTriStateGroup. I would like to spend some time thinking about how it might be possible to better generalize these classes. 

These classes are key to how the beam status and utilities are implemented on BioXAS now; if a suitable generalization is found, it would be pretty simple to have something like an AMBeamStatus, or a general way of handling beamline utilities.

helfrij:
No longer necessary with changes in #2352.

PR #2149, opened 2016-03-18
Pull Request for Issue2088: Creating general temperature monitor controls

helfrij:
- Created new classes AMTemperatureMonitor, ~~AMTemperatureMonitors~~ AMTemperatureMonitorGroup, AMTemperatureMonitorGroupStatus
- Modified BioXASBeamline, BioXASUtilities to use the new temperature monitors classes.


helfrij:

source/beamline/AMTemperatureMonitor.h
```diff
@@ -0,0 +1,64 @@
...
+
+public slots:
+	/// Sets the status control.
+	void setStatusControl(AMControl *control, double goodValue, double badValue);
+	/// Sets the value control.
+	void setValueControl(AMControl *control);
+
+protected:
+	/// The good status value.
+	double goodValue_;
```
XuetaoLiu-CLS: I think it might be better if we name it `goodStatusValue_`

helfrij: Ha! Yes, you're probably right.

source/beamline/AMTemperatureMonitors.h
```diff
@@ -0,0 +1,43 @@
+#ifndef AMTEMPERATUREMONITORS_H
+#define AMTEMPERATUREMONITORS_H
+
+#include "beamline/AMConnectedControl.h"
+#include "beamline/AMTemperatureMonitorsStatus.h"
+
+class AMTemperatureMonitors : public AMConnectedControl
```
XuetaoLiu-CLS: I'm almost confused this class with `AMTemperatureMonitor`. Therefore, I'm wondering whether we could rename this class to `AMTemparatureMonitorGroup` or `AMTemparatureMonitorSet`, or sth alike?

source/beamline/BioXAS/BioXASUtilities.cpp
```diff
@@ -48,10 +49,13 @@ BioXASUtilities::BioXASUtilities(const QString &name, QObject *parent) :
...
-	temperatureMonitors_ = new BioXASUtilitiesGroup(QString("%1%2").arg(name).arg("TemperatureMonitors"), this);
-	addControl(temperatureMonitors_);
+	temperatureMonitors_ = new AMTemperatureMonitors(QString("%1%2").arg(name).arg("TemperatureMonitors"), this);
+	addControl(temperatureMonitors_->statusControl(), AMTemperatureMonitorsStatus::Bad, AMTemperatureMonitorsStatus::Good);
 
-	connect( temperatureMonitors_, SIGNAL(valueChanged(double)), this, SIGNAL(temperatureMonitorsValueChanged(double)) );
+//	temperatureMonitors_ = new BioXASUtilitiesGroup(QString("%1%2").arg(name).arg("TemperatureMonitors"), this);
+//	addControl(temperatureMonitors_);
+
+//	connect( temperatureMonitors_, SIGNAL(valueChanged(double)), this, SIGNAL(temperatureMonitorsValueChanged(double)) );
```
XuetaoLiu-CLS: I guess these codes will be removed eventually

helfrij: Yes, I forgot to remove earlier. Will remove now.

XuetaoLiu-CLS:
Looks good to me other than some concerns on names.

Time Estimate: 20 minutes

source/beamline/AMTemperatureMonitors.h
helfrij: I agree that the names are close enough to be confusing. Because the temperature monitors share so much in common with other controls like ion pumps, flow switches, pressure monitors, and flow transducers I actually don't expect this class to stick around for very long. I expect in #2127 this class will be entirely replaced with something more general, like `AMUtilitiesControlGroup`... or something. I'm having trouble coming up with a good name for that though. Something that's general enough to apply to all of those groups of controls, but not confusing.

source/beamline/AMTemperatureMonitorGroupStatus.cpp
```diff
@@ -0,0 +1,158 @@
...
+	// all are in a good state.
+
+	QList<AMControl*> monitors = childControls();
+
+	if (monitors.count() > 0) {
+
+		for (int i = 0, count = monitors.count(); i < count && !badMonitorFound; i++) { // we want to stop if any one child isn't in state 1.
+			AMTemperatureMonitor *monitor = qobject_cast<AMTemperatureMonitor*>(monitors.at(i));
+
+			if (monitor && !monitor->isGood())
```
dretrex: Does this imply that `!monitor_->isGood()` is different from `monitor_->isBad()`?  Conversely, if `isBad()` is any monitor being false, wouldn't it be sufficient for `bool isGood() const { return !isBad(); }`?

source/beamline/BioXAS/BioXASUtilities.h
```diff
@@ -69,7 +72,7 @@ class BioXASUtilities : public BioXASUtilitiesGroup
 	/// Returns the pressure monitors control.
 	BioXASUtilitiesGroup* pressureMonitors() const { return pressureMonitors_; }
 	/// Returns the temperature monitors control.
-	BioXASUtilitiesGroup* temperatureMonitors() const { return temperatureMonitors_; }
+	AMTemperatureMonitorGroup* temperatureMonitors() const { return temperatureMonitors_; }
```
dretrex: Since you changed the name of the class, should the getter should reflect that too?

dretrex:
Couple comments.  Looks good otherwise.
Time Estimate: 10 minutes

source/beamline/BioXAS/BioXASUtilities.h
dretrex: Actually, I take it back.  The name is totally fine.

source/beamline/AMTemperatureMonitorGroupStatus.cpp
helfrij: I think what you describe is true for controls of this particular type. Since `caget` for the values of the status only shows `Good` or `Bad`, as defined there this control will only ever be one or the other. I'm not sure if that's true for all controls. Is it better to keep it simple?

source/beamline/BioXAS/BioXASUtilities.h
helfrij: Yes, I'll make that change.

source/beamline/AMTemperatureMonitorGroupStatus.cpp
dretrex: I think that while the interface being presented is essentially boolean (i.e.: Good or Bad) that one should have the necessary logic and then the other be the negation of that.  The way that you implemented isBad, I think isGood could just be the negation.

Conversely, if you had implemented isGood such that every monitor needed to be good, then isBad() would just be !isGood().

helfrij: Okay, I'll take a look.

acquamanCommon.pri
```diff
@@ -468,7 +468,10 @@ HEADERS += \
 	source/beamline/AMSlitCenter.h \
 	source/ui/beamline/AMSlitView.h  \
 	source/beamline/AMSlits.h \
-	source/ui/beamline/AMSlitsView.h
+	source/ui/beamline/AMSlitsView.h \
+	source/beamline/AMTemperatureMonitor.h \
+    $$PWD/source/beamline/AMTemperatureMonitorGroup.h \
+    $$PWD/source/beamline/AMTemperatureMonitorGroupStatus.h
```
XuetaoLiu-CLS: is it possible to rmove the `$$PWD`?

source/beamline/AMTemperatureMonitor.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+bool AMTemperatureMonitor::isBad() const
+{
+	bool result = false;
+
+	if (statusControl_ && statusControl_->canMeasure())
+		result = (statusControl_->value() == badStatusValue_);
+
+	return result;
+}
```
XuetaoLiu-CLS: just pop up a question: is there a state of `in-between`(is not `good` and  not `bad`?

acquamanCommon.pri
helfrij: Yup, I can take it out. It's added there automatically for classes that I create on my laptop, it seems.

source/beamline/AMTemperatureMonitor.cpp
helfrij: For these controls, there actually isn't. @dretrex has suggested simplifying the logic somewhat so the control can only be one or the other.

helfrij:
More comments, @XuetaoLiu-CLS @dretrex?

dretrex:
Nope.

PR #2179, opened 2016-03-24
Pull Request for Issue2174: Initial PGM project

dretrex:
This combines two issues together and does the initial project setup for PGM.  It contains the beamline, app controller, and main files.

### DIFF
--- a/source/application/AMDatamanAppController.cpp
+++ b/source/application/AMDatamanAppController.cpp
@@ -873,6 +873,8 @@ bool AMDatamanAppController::startupPopulateNewDatabase()
 	bioXASImaging.storeToDb(db);
 	AMFacility sxrmb("SXRMB", "CLS SXRMB Beamline", ":/clsIcon.png");
 	sxrmb.storeToDb(db);
+	AMFacility pgm("PGM", "CLS PGM Bealine", ":/clsIcon.png");
+	pgm.storeToDb(db);
 
 	return true;
 }
--- a/source/application/PGM/PGMAppController.cpp
+++ b/source/application/PGM/PGMAppController.cpp
@@ -1,0 +1,134 @@
+/*
+Copyright 2010-2012 Mark Boots, David Chevrier, and Darren Hunter.
+Copyright 2013-2014 David Chevrier and Darren Hunter.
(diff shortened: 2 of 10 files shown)


source/application/PGM/PGMAppController.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include "util/AMPeriodicTable.h"
+
+#include "ui/util/AMChooseDataFolderDialog.h"
+#include "ui/AMMainWindow.h"
+#include "ui/dataman/AMGenericScanEditor.h"
+#include "ui/acquaman/AMScanConfigurationViewHolder3.h"
+
+PGMAppController::PGMAppController(QObject *parent)
+	: AMAppController(parent)
```
helfrij: I think there is a CLSAppController class now; do you prefer to use AMAppController here?

helfrij:
Looks good! One minor question about inheritance of app controller class.

Time Estimate: 10 minutes.

source/application/PGM/PGMAppController.cpp
helfrij: Nope, I take it back. I was thinking of the changes in #1968, which haven't been merged yet.

source/application/AMDatamanAppController.cpp
```diff
@@ -873,6 +873,8 @@ bool AMDatamanAppController::startupPopulateNewDatabase()
 	bioXASImaging.storeToDb(db);
 	AMFacility sxrmb("SXRMB", "CLS SXRMB Beamline", ":/clsIcon.png");
 	sxrmb.storeToDb(db);
+	AMFacility pgm("PGM", "CLS PGM Bealine", ":/clsIcon.png");
+	pgm.storeToDb(db);
```
XuetaoLiu-CLS: @dretrex I know this is closed. Since the #483 is still under development, this changes here will have the same issue as what we had for the other beamline since we didn't define the facility id for PGM .

dretrex:
You are quite correct.  We will need update #483 with this extra facility.  I wanted to have this in master so that @dleacock could start on some of the PGM issues.

PR #2186, opened 2016-03-28
Pull Request for Issue2182: Using name instead of configuration user scan name.

helfrij:
- Modified the view classes for BioXAS XAS scan configurations to use the scan name instead of the user scan name.
- I see a lot about these classes that could stand to be refactored. Another issue.

### DIFF
--- a/source/ui/BioXAS/BioXASXASScanConfigurationView.cpp
+++ b/source/ui/BioXAS/BioXASXASScanConfigurationView.cpp
@@ -20,8 +20,8 @@ const AMScanConfiguration* BioXASXASScanConfigurationView::configuration() const
 
 void BioXASXASScanConfigurationView::setConfigurationName(BioXASXASScanConfiguration *configuration, const QString &newName)
 {
-	if (configuration && configuration->userScanName() != newName)
-		configuration->setUserScanName(newName);
+	if (configuration)
+		configuration->setName(newName);
 }
 
 void BioXASXASScanConfigurationView::setConfigurationEdge(BioXASXASScanConfiguration *configuration, const AMAbsorptionEdge &newEdge)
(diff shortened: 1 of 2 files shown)


helfrij:
Hey @dretrex. I think these changes solve the 'editing scan configuration name' problem. I've tested this so far by modifying a scan configuration in the view, adding the configuration to the workflow, and verifying visually that the workflow configuration has the new name. Is there another way that I should test this, maybe that would be specific to the 'new' way Ian C. wants to use scan configurations?

dretrex:
The last check would be to start a scan and see if the name on the left side bar matches the name that you set.  Once that's done you are good to go.

helfrij:
Okay, thanks @dretrex. I'll check.

helfrij:

source/ui/BioXAS/BioXASXASScanConfigurationView.cpp
```diff
@@ -20,8 +20,8 @@ const AMScanConfiguration* BioXASXASScanConfigurationView::configuration() const
 
 void BioXASXASScanConfigurationView::setConfigurationName(BioXASXASScanConfiguration *configuration, const QString &newName)
 {
-	if (configuration && configuration->userScanName() != newName)
-		configuration->setUserScanName(newName);
+	if (configuration)
```
iainworkman: Is there a reason why the second part of the check was removed along with the change from using name() instead of userScanName()?

iainworkman:
Small question. Otherwise looks good.
Time Estimate: 2 minutes.

source/ui/BioXAS/BioXASXASScanConfigurationView.cpp
helfrij: No major reason. The same check is performed in `AMDbObject::setName()` already, don't need to have a duplicate check here.

dretrex:
Time Estimate: 1 minute

PR #2187, opened 2016-03-29
Pull Request for Issue2160: general beamline controls

helfrij:
- Created general beamline control class, AMBeamlineControl. This class has a control that reports a value, and a good/bad status control as well.
- Created general beamline control group class, AMBeamlineControlGroup. Beamline controls can be added to and removed from this class, it acts like a container.
- Created general beamline control group status class, AMBeamlineControlGroupStatus. Beamline controls can be added to and removed from this class. Its value is the overall status for all added controls.
- Created CLS-level beamline control class, CLSBeamlineControl. This class has the enumerated state convention used for all beamline controls at the CLS.
- Created CLS-level classes for the ion pumps, flow switches, flow transducers, pressure monitors, and temperature monitors. Each class inherits from CLSBeamlineControl.
- Modified the BioXAS beamline classes to use the new CLS beamline controls, instead of the utilities class. The BioXASUtilities class hasn't been deleted yet, as it's still used for the shutters and valves. Deleting it, and the BioXASBiState- and BioXASTriStateControls will be handled as a separate issue.


helfrij:

acquamanCommon.pri
```diff
@@ -467,9 +467,9 @@ HEADERS += \
 	source/ui/beamline/AMSlitView.h  \
 	source/beamline/AMSlits.h \
 	source/ui/beamline/AMSlitsView.h \
-	source/beamline/AMTemperatureMonitor.h \
-	source/beamline/AMTemperatureMonitorGroup.h \
-	source/beamline/AMTemperatureMonitorGroupStatus.h
+    $$PWD/source/beamline/AMBeamlineControlGroupStatus.h \
+    $$PWD/source/beamline/AMBeamlineControlGroup.h \
+    $$PWD/source/beamline/AMBeamlineControl.h
```
XuetaoLiu-CLS: I'm wondering whether the `$$PWD/` is mandatory. If not, I prefer to remove it to be consistent with the other header/source files.

source/beamline/AMBeamlineControlGroup.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+bool AMBeamlineControlGroup::isGood() const
+{
+	return status_->isGood();
+}
+
+bool AMBeamlineControlGroup::hasControl(AMControl *control) const
+{
+	return status_->hasChildControl(control);
```
XuetaoLiu-CLS: Is this equivalent to `children_->contains(control)` ?

source/beamline/AMBeamlineControlGroup.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+bool AMBeamlineControlGroup::addControl(AMBeamlineControl *control)
+{
+	bool result = false;
+
+	if (control && !children_.contains(control)) {
+		addChildControl(control);
```
XuetaoLiu-CLS: Just a thought for discussion. I think it might be ideal if we can change the signature of `addChildControl()`/`removeChildControl()` to return True/False and then this function can be revised as:
```
if (addChildControl(control)) {
   status_->addControl(control);
   result = true;
}

Comment (XuetaoLiu-CLS) [inline source/beamline/AMBeamlineControlGroup.cpp]: If we did this, then the implementation of `addControl()` / `removeControl()` of `AMBeamlineControlGroupStatus` can be removed.

**Comment by XuetaoLiu-CLS** [comment]
Look good. Nice code refactor.

Time Estimate: 20 minutes

File: acquamanCommon.pri (hunk 1 above)
Comment (helfrij) [inline acquamanCommon.pri]: When I create new classes or modify class names using Creator on an OPI, the `$$PWD` doesn't appear, but it seems to be added by default for the same actions using Creator on my Mac.

File: source/beamline/AMBeamlineControlGroup.cpp (hunk 3 above)
Comment (helfrij) [inline source/beamline/AMBeamlineControlGroup.cpp]: I agree, that's a change that definitely makes sense. We could probably also rename those methods, as we're not actually adding/removing a 'child' in the formal sense--not reparenting anything, I mean.

File: acquamanCommon.pri (hunk 1 above)
Comment (helfrij) [inline acquamanCommon.pri]: I imagine there must be a setting for this somewhere. I can't see anything in Creator preferences yet, and I'm having trouble finding something relevant on Google too.

File: source/beamline/AMBeamlineControlGroup.cpp (hunk 2 above)
Comment (helfrij) [inline source/beamline/AMBeamlineControlGroup.cpp]: It should be, at least how I've defined add/removeControl(). I'll change it to `children_->contains(control)`, as that's a little more clear.

Comment (helfrij) [inline source/beamline/AMBeamlineControlGroup.cpp]: Ah, there's already an `AMControl::hasChildControl()`. I'll use that instead.

File: source/beamline/AMBeamlineControlGroup.cpp [hunk 4]
```diff
@@ -0,0 +1,62 @@
...
+
+	return result;
+}
+
+bool AMBeamlineControlGroup::removeControl(AMBeamlineControl *control)
+{
+	bool result = false;
+
+	if (control && children_.contains(control)) {
+		removeChildControl(control);
```
dretrex: Does `addChildControl()` and `removeChildControl()` do the `control && children_.contains(control)` check?  If so, the add/remove control could just be in the if statement like @XuetaoLiu-CLS suggested.

dretrex:
Just the one comment.  Looks good!

Time Estimate: 10 minutes

source/beamline/AMBeamlineControlGroup.cpp
helfrij: It doesn't, but I can add that now. Those checks seem like a good thing to have, generally.

helfrij: `AMControl::addChildControl(...)` and `AMControl::removeChildControl(...)` don't return a success bool right now, though, so I don't think they can replace the if-check conditional just yet.

dretrex: Alright, sounds good.

PR #2189, opened 2016-03-29
Pull Request for Issue2183: EXAFS scans need different data collection paradigm

dretrex:
I modified the BioXASXASScanController to use all the individual ROIs.  It still uses the summed spectra for the full ROI.  I needed to pass in a means of updating the axis info of the 1D detector emulator because the controls don't have the in depth knowledge of the detector.

I'm sure there are some ways to make this more general (for instance, using the createXYZAB methods in AMGenericStepScanController), but for the time being this will suffice.


helfrij:
Looks good, @dretrex.

Time Estimate: 15 minutes.

PR #2198, opened 2016-03-31
Pull Request for Issue2191: Adding 'all good events' to Ge detector ICRs.

helfrij:
- Modified the dead time button to be able to display counts information if the ICR sources are provided.
- Modified the XRF detailed detector view to use the ICR-based dead time buttons, if the detector has ICR sources.
- Added the 'all good events' controls to the Side and Main inboard Ge detectors.
- Cleaned up some stray cases in AMXRFDetector and AMXRFScanController where it was assumed that if there are ICRs then there are also OCRs--not true for the Xspress3 detectors on BioXAS.

### DIFF
--- a/source/acquaman/AMXRFScanController.cpp
+++ b/source/acquaman/AMXRFScanController.cpp
@@ -97,7 +97,7 @@ void AMXRFScanController::onDetectorAcquisitionFinished()
 		}
 	}
 
-	else {
+	else if (!detector_->inputCountSources().isEmpty() && !detector_->outputCountSources().isEmpty()){
 		for (int i = 0, elements = detector_->elements(); i < elements; i++){
 
 			detector_->rawSpectrumSources().at(i)->values(AMnDIndex(0), AMnDIndex(detector_->size(0)-1), spectrum.data());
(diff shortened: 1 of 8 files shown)


helfrij:

source/acquaman/AMXRFScanController.cpp
```diff
@@ -97,7 +97,7 @@ void AMXRFScanController::onDetectorAcquisitionFinished()
 		}
 	}
 
-	else {
+	else if (!detector_->inputCountSources().isEmpty() && !detector_->outputCountSources().isEmpty()){
```
XuetaoLiu-CLS: I think we don't need `!detector_->inputCountSources().isEmpty()`, since the `detector_->inputCountSources()` is for sure not empty in the else branch

XuetaoLiu-CLS:
Looks good other than an optimization suggestions.

Time Estimate: 10 minutes

source/acquaman/AMXRFScanController.cpp
helfrij: Okay, I'll remove the condition.

dretrex:
Looks good.
Time Estimate: 2 minutes

PR #2204, opened 2016-04-01
Pull Request for Issue2202: dark current check in scan initialization

helfrij:
Changes:
- There wasn't an explicit getter for the SOE shutter previously, so added class variable, getter, signal, setter for the SOE shutter to BioXASSideBeamline, BioXASMainBeamline, and BioXASBeamline.
- Added method to AMDetector that returns the dark current valid state for the given acquisition time.
- Added `BioXASBeamline::createDarkCurrentMeasurementAction(double)`, returns an action that performs a dark current measurement if the beamline has a valid scaler and SOE shutter to use.
- Added check to the scan initialization actions in BioXASBeamline. If one of the scaler channel detectors requires a dark current measurement, a call to `createDarkCurrentMeasurementAction(double)` is added to the initialization actions.

Testing:
- [x] No DC measurement (startup), XAS scan 1s --> should measure DC in scan init
- [x] DC measurement 1s, XAS scan 1s --> should NOT measure DC in scan init
- [x] DC measurement 1s, change gain, XAS scan 1s --> should measure DC in scan init
- [x] DC measurement 1s, XAS scan 2s --> should measure DC in scan init
- [x] DC measurement 2s, XAS scan 1s --> should NOT measure DC in scan init
- [x] DC measurement 1s, EXAFS scan (some points > 1s) --> should measure DC in scan init

In all cases, the DC corrected data sources should be included in scan data.

### DIFF
--- a/source/beamline/BioXAS/BioXASBeamline.cpp
+++ b/source/beamline/BioXAS/BioXASBeamline.cpp
@@ -85,9 +85,32 @@ AMAction3* BioXASBeamline::createScanInitializationAction(AMGenericStepScanConfi
 		AMListAction3 *scalerInitialization = 0;
 		CLSSIS3820Scaler *scaler = BioXASBeamline::bioXAS()->scaler();
 
-		if (BioXASBeamlineSupport::usingScaler(configuration)) {
+		if (scaler && BioXASBeamlineSupport::usingScaler(configuration)) {
 			scalerInitialization = new AMListAction3(new AMListActionInfo3("BioXAS scaler initialization", "BioXAS scaler initialization"));
 			scalerInitialization->addSubAction(scaler->createContinuousEnableAction3(false)); // Check that the scaler is in single shot mode and is not acquiring.
+
+			// Determine whether a dark current measurement is needed. If so, add measurement to scaler initialization.
+
+			bool requiresDarkCurrentMeasurement = false;
+
+			for (int i = 0, count = scaler->channels().count(); i < count && !requiresDarkCurrentMeasurement; i++) {
+				AMDetector *channelDetector = 0;
+
+				if (scaler->channelAt(i))
+					channelDetector = scaler->channelAt(i)->detector();
+
+				requiresDarkCurrentMeasurement = (channelDetector && !channelDetector->darkCurrentValidState());
+			}
+
+			if (requiresDarkCurrentMeasurement && BioXASBeamline::bioXAS()->soeShutter()) {
+				AMListAction3* scalerDarkCurrentMeasurementAction = new AMListAction3(new AMListActionInfo3("BioXAS scaler dark current measurement", "BioXAS scaler dark current measurement action"), AMListAction3::Sequential);
+
+				scalerDarkCurrentMeasurementAction->addSubAction(AMActionSupport::buildControlMoveAction(BioXASBeamline::bioXAS()->soeShutter(), BioXASShutters::Closed));
+				scalerDarkCurrentMeasurementAction->addSubAction(scaler->createMeasureDarkCurrentAction(configuration->scanAxisAt(0)->regionAt(0)->regionTime()));
+				scalerDarkCurrentMeasurementAction->addSubAction(AMActionSupport::buildControlMoveAction(BioXASBeamline::bioXAS()->soeShutter(), BioXASShutters::Open));
+
+				scalerInitialization->addSubAction(scalerDarkCurrentMeasurementAction);
+			}
 		}
 
 		if (scalerInitialization)
--- a/source/beamline/BioXAS/BioXASBeamline.cpp
+++ b/source/beamline/BioXAS/BioXASBeamline.cpp
@@ -118,30 +118,15 @@ AMAction3* BioXASBeamline::createScanInitializationAction(AMGenericStepScanConfi
 			scalerInitialization = new AMListAction3(new AMListActionInfo3("BioXAS scaler initialization", "BioXAS scaler initialization"));
 			scalerInitialization->addSubAction(scaler->createContinuousEnableAction3(false)); // Check that the scaler is in single shot mode and is not acquiring.
 
-			// Determine the longest region time for the scan.
+			// Determine the longest region time of all axis in the scan.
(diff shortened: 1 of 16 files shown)


helfrij:

source/beamline/BioXAS/BioXASBeamline.cpp
```diff
@@ -44,9 +45,37 @@ bool BioXASBeamline::isConnected() const
...
+
+		// Return shutter to initial settings.
+
+		if (shutterOpen)
+			result->addSubAction(AMActionSupport::buildControlMoveAction(soeShutter, CLSExclusiveStatesControl::Open));
+	}
+
+	return result;
+}
+#include <QDebug>
```
iainworkman: Is this include still required?

helfrij: Nope, I'll remove.

iainworkman:
Looks good to me. 
Time Estimate: 10 minutes

source/beamline/BioXAS/BioXASBeamline.cpp
```diff
@@ -88,6 +117,45 @@ AMAction3* BioXASBeamline::createScanInitializationAction(AMGenericStepScanConfi
 		if (BioXASBeamlineSupport::usingScaler(configuration)) {
 			scalerInitialization = new AMListAction3(new AMListActionInfo3("BioXAS scaler initialization", "BioXAS scaler initialization"));
 			scalerInitialization->addSubAction(scaler->createContinuousEnableAction3(false)); // Check that the scaler is in single shot mode and is not acquiring.
+
+			// Determine the longest region time for the scan.
+
+			double maxRegionTime = configuration->scanAxisAt(0)->regionAt(0)->regionTime();
+
+			for (int axisIndex = 0, axisCount = configuration->scanAxes().count(); axisIndex < axisCount; axisIndex++) {
```
dretrex: Refactoring supposition: would this be something that maybe a scan axis should return?

helfrij: A getter for the longest dwell time, you mean?

source/beamline/BioXAS/BioXASBeamline.cpp
```diff
@@ -88,6 +117,45 @@ AMAction3* BioXASBeamline::createScanInitializationAction(AMGenericStepScanConfi
...
+
+					if (maxRegionTime < regionDwell)
+						maxRegionTime = regionDwell;
+				}
+			}
+
+			// Determine whether a dark current measurement is needed, by asking each scaler channel detector
+			// if its dark current value would be valid for the longest region time.
+
+			bool requiresDarkCurrentMeasurement = false;
```
dretrex: Is being the same as the maximum time in a region sufficient time for a dark current?  Would we need something like:

```c++
double darkCurrentTime = maxRegionTime > 10 ? maxRegionTime : 10;
createDarkCurrentAction(darkCurrentTime);
```

source/beamline/BioXAS/BioXASBeamline.cpp
dretrex: Yeah, on a per axis basis.

dretrex:
Time Estimate: 5 minutes

source/beamline/BioXAS/BioXASBeamline.cpp
helfrij: The channel detectors' dark current value is valid as long as the scaler dwell time in a scan is less than or equal to the dark current time. It doesn't really help us to set a maximum on the dark current time, I think that's what's effectively done above. The dark current measurement must happen for a time that's equal to or longer than the dwell time at any point in the scan. I think.

dretrex: But is there a minimum amount of time that should be used to ensure proper sampling of the dark current?  For instance, is it sufficient that you only have 1 second of dark current for a 1 second data collection?

source/beamline/BioXAS/BioXASBeamline.cpp
helfrij: Are you thinking that the max time should be a cached value, or would the loop just be shuffled over to the AMScanAxis class?

dretrex: Looping is fine :)

source/beamline/BioXAS/BioXASBeamline.cpp
helfrij: Oh wait, yes, I misread what you wrote. I think all that's really necessary for a default XANES scan, for example, is a dark current measurement of 1s. However, a user can still take a dark current measurement for as long as they'd like before a scan.

dretrex: Okay, sounds good.

source/beamline/BioXAS/BioXASBeamline.cpp
helfrij: I'll make the change. :+1:

source/dataman/AMScanAxis.cpp
```diff
@@ -369,3 +370,49 @@ QString AMScanAxis::toString(const QString &units) const
...
+
+double AMScanAxis::minimumRegionTime() const
+{
+	double minTime = 0;
+
+	// Iterate through the regions, identify the smallest region time.
+
+	for (int i = 0, count = regionCount(); i < count; i++) {
+
+		double regionDwell = regionAt(i)->regionTime();
```
dretrex: I believe the time needs to be casted.  `double regionDwell = double(regionAt(i)->regionTime());`

Also, I think the conditional needs to be a bit more strict:

```c++
if (minTime > regionDwell && regionDwell >= 0)  // This ensures that the times can't be negative if the region hasn't been fully specified.
    minTime = regionDwell;
```

Finally, the `minTime` needs to be initialized to positive infinite or the first region's time because this function will always return 0 unless the time is negative.

helfrij: Sounds good. I'll make the changes.

Issue #2167, opened 2016-03-24
PGM Components - BPMs


dretrex:
A list of BPMs is required but these will be read only controls and will need a means of displaying the value.

dretrex:
This will also need a red light/green light heads up display.

XuetaoLiu-CLS:
Time Estimate: 5 hours (including @dleacock 's time)

PR #2213, opened 2016-04-04
Pull Request for Issue2167:  PGM Components - BPMs

dleacock:
#2167 

Added BPM pv's and means of display.

**16/4** ToDo:
- [x] Change PGMBpmControl inheritence relationship to a composition one. 

**15/4** ToDo: 
- [x] Create PGMBeamStatusView for beam current and lifetime
- [x] Create PGMBpmControl for bpm reading,logic, constants and status of a particular bpm.

ToDo:
- [x] Add green light / light
- [x] What values are good for red light/green light?


dleacock:
What would constitute green or red values?

dretrex:
You'll need to ask Lucia.

dleacock:
Ring current changes red if dipped below 100mA.   BPM values also change red if +/- 50 um (as specified by Lucia). 

Screenshot: http://imgur.com/a/2VF1l

source/ui/PGM/PGMPersistentView.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "beamline/PGM/PGMBeamline.h"
+
+#include <QGridLayout>
+#include <QLabel>
+#include <QGroupBox>
+
+PGMPersistentView::PGMPersistentView(QWidget *parent) :
+    QWidget(parent)
+
+{
```
iainworkman: Might be an idea to have all these controls in a PGMBeamStatusView, and then have the persistent view contain one of those. That will help with the potential for the persistent view code from becoming fairly cluttered in the future.

source/ui/PGM/PGMPersistentView.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+}
+
+void PGMPersistentView::onRingCurrentChange(double value){
+    // If beam has tripped of dipped too low, mark text as red.
+    if(value < 100){
+        ringCurrentValue_->setStyleSheet( "color : red"  );
+        ringCurrentValue_->setText(QString("%1 mA").arg(value, 0, 'f', 2));
+    } else {
+        ringCurrentValue_->setText(QString("%1 mA").arg(value, 0, 'f', 2));
```
iainworkman: This will also need a 
```c++
ringCurrentValue_->setStyleSheet( "color : black" );
```
... in order to flip it back once the ring current returns.

source/ui/PGM/PGMPersistentView.h
```diff
@@ -0,0 +1,61 @@
...
+class PGMPersistentView : public QWidget
+{
+    Q_OBJECT
+public:
+    explicit PGMPersistentView(QWidget *parent = 0);
+
+    /// Storing the average (x, y) values for the beam position for 10ID, 11ID-1 and 11ID-2
+    /// When the bpm changes, they are allowed to change within +/- 50 or a flag will be raised (color change in text).
+    int avg10IDx_, avg10IDy_, avg11ID1x_, avg11ID1y_, avg11ID2x_, avg11ID2y_;
+    int bpmVariance_;
```
iainworkman: I notice that they represent constants, so I would suggest either making them const ints or using #defines
Either way they should be protected instead of public

source/ui/PGM/PGMPersistentView.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+void PGMPersistentView::onBPM11ID1ValueXChanged(double value){
+
+    if( value > (avg11ID1x_ + bpmVariance_) || value < (avg11ID1x_ - bpmVariance_) ){
+        bpm11ID1valueX_->setStyleSheet( "color : red"  );
+        bpm11ID1valueX_->setText(QString("%1 um").arg(value, 0, 'f', 0));
+    } else {
+        bpm11ID1valueX_->setStyleSheet( "color : black"  );
+        bpm11ID1valueX_->setText(QString("%1 um").arg(value, 0, 'f', 0));
```
iainworkman: In fact, all these bpm controls share a commonality.
* They have an average value and a variance.
* Getting variance distance from the average puts them in a bad state
* Otherwise they are in a good state.

I think it might be an idea to wrap these controls in a class, which contains the logic for determining whether they are in a good state or not. Then all the view needs to do is listen for a 'inValidStateChanged(bool)' type signal and update itself accordingly.

iainworkman:
A couple of questions and recommendations, otherwise looks good.
Time Estimate: 20 minutes

dleacock:
@iainworkman I made some changes based on your suggestions, thanks for your help!

dretrex:
Hey @dleacock, I think your BPM control would probably work better as a _has a_ relationship with the control rather than an _is a_ relationship.

dleacock:
Updated it. 

5421a984a5ef75

iainworkman:
Looks great.

dleacock:
Time Estimate: 6 hours

source/beamline/PGM/PGMBpmControl.h
```diff
@@ -0,0 +1,55 @@
+#ifndef PGMBPMCONTROL_H
+#define PGMBPMCONTROL_H
+
+#include <QObject>
+#include "beamline/AMPVControl.h"
+
+/// Class to hold the Beam Position Monitor class.
+class PGMBpmControl : public QObject
```
dretrex: Naming convention thing: PGMBPMControl

source/beamline/PGM/PGMBpmControl.h
```diff
@@ -0,0 +1,55 @@
...
+public:
+    /*!
+    \param name - A unique description of this control
+    \param pvName - The EPICS channel-access name for this Process Variable
+    \param avgValue - average beam position value, meaured in um. Provided by beamline staff.
+    \param variance - allowed value variance. Provided by beamline staff.
+    \param parent - QObject parent class
+    */
+    /// Constructor
+    PGMBpmControl(const QString &name, const QString &pvName, const int avgValue, const int variance, QObject *parent = 0);
```
dretrex: `const int x` is not necessary.

source/beamline/PGM/PGMBeamline.cpp
```diff
@@ -77,7 +91,10 @@ void PGMBeamline::setupControlsAsDetectors()
 
 void PGMBeamline::setupExposedControls()
 {
-
+    // Storage ring current
+    addExposedControl(ringCurrent_);
```
dretrex: These are not typically exposed controls.  Exposed controls are things that you plan to expose to scan controllers to change.

source/beamline/PGM/PGMBpmControl.cpp
```diff
@@ -0,0 +1,36 @@
...
+    variance_ = variance;
+
+    bpmPV_ = new AMReadOnlyPVControl(name, pvName, this);
+
+    // Get updated pv value from AMReadOnlyPVControl.
+    connect(bpmPV_, SIGNAL(valueChanged(double)), this, SLOT(setCurrentValue(double)));
+
+}
+
+void PGMBpmControl::isValid(){
```
dretrex: This function doesn't really make sense.  `isValid()` should return a bool and if you want the valid to change state you should use a more traditional signal/slot mechanism to update it.  Something like:

```c++
// In constructor
connect(bpmPV_, SIGNAL(valueChanged(double)), this, SLOT(updateValidity(double)));


// Elsewhere in the class
bool PGMBPMControl::updateValidity(double newValue)
{
   int upperBound = averageValue_ + variance_;
   int lowerBound = averageValue_ - variance_;
   bool newState = (newValue < upperBound || newValue > lowerBound);

   if (newState != state_){
       
      state_ = newState;
      emit stateValidityChanged(state_);
   }
}
```

source/beamline/PGM/PGMBpmControl.h
```diff
@@ -0,0 +1,55 @@
+#ifndef PGMBPMCONTROL_H
+#define PGMBPMCONTROL_H
+
+#include <QObject>
```
dretrex: QObject include is technically unnecessary because you are already inheriting it from AMPVControl.

source/ui/PGM/PGMPersistentView.cpp
```diff
@@ -0,0 +1,178 @@
...
+    connect(PGMBeamline::pgm()->bpm10IDxControl(), SIGNAL(onValueChanged(QString)), bpm10IDvalueX_, SLOT(setText(QString)));
+    connect(PGMBeamline::pgm()->bpm10IDxControl(), SIGNAL(isValidStateChanged(bool)), this, SLOT(onBPM10IDxStatusChanged(bool)));
+    // BPM 10ID-x connections
+    connect(PGMBeamline::pgm()->bpm10IDyControl(), SIGNAL(onValueChanged(QString)), bpm10IDvalueY_, SLOT(setText(QString)));
+    connect(PGMBeamline::pgm()->bpm10IDyControl(), SIGNAL(isValidStateChanged(bool)), this, SLOT(onBPM10IDyStatusChanged(bool)));
+    // BPM 10ID-x connections
+    connect(PGMBeamline::pgm()->bpm11ID1xControl(), SIGNAL(onValueChanged(QString)), bpm11ID1valueX_, SLOT(setText(QString)));
+    connect(PGMBeamline::pgm()->bpm11ID1xControl(), SIGNAL(isValidStateChanged(bool)), this, SLOT(onBPM11ID1xStatusChanged(bool)));
+    // BPM 10ID-x connections
+    connect(PGMBeamline::pgm()->bpm11ID1yControl(), SIGNAL(onValueChanged(QString)), bpm11ID1valueY_, SLOT(setText(QString)));
```
dretrex: Does this actually work?  Controls only emit a valueChanged(double) signal.  I'm not sure where your `onValueChanged(QString)` is coming from.

dretrex:
Time Estimate: 10 minutes

source/ui/PGM/PGMPersistentView.cpp
dleacock: It comes from PGMBPMControl
`void PGMBPMControl::setCurrentValue(double value){

    currentValue_ = value;
    isValid();
    emit onValueChanged(QString("%1 um").arg(value, 0, 'f', 0));
}`

source/beamline/PGM/PGMBeamline.cpp
```diff
@@ -91,10 +91,7 @@ void PGMBeamline::setupControlsAsDetectors()
 
 void PGMBeamline::setupExposedControls()
 {
-    // Storage ring current
-    addExposedControl(ringCurrent_);
-    // Beam lifetime
-    addExposedControl(beamLifetime_);
```
dleacock: Also, removed the exposed controls.

source/beamline/PGM/PGMBPMControl.cpp
```diff
@@ -0,0 +1,38 @@
...
+        state_ = newState;
+        emit stateValidityChanged(state_);
+    }
+}
+
+void PGMBPMControl::setCurrentValue(double value){
+
+    currentValue_ = value;
+    updateValidity(currentValue_);
+    emit onBPMValueChanged(QString("%1 um").arg(value, 0, 'f', 0));
```
dretrex: Oh, shoot.  Sorry @dleacock.  I think I misunderstood your original signal so you'll need to make one more change. 

In general, there is no need for a control to emit a value changed signal as a string like this (there are some exceptions, but this is not one of them).  The model should emit the value as a double (which is the real data) and then the view would have the `QString("%1 um").arg(value, 0, 'f', 0)` call inside it's `onBPMValueChanged(double)` slot.  The main reason for this is you make it harder for models to talk to each other by having your model return a string as the value.

dleacock: Got it.  My thinking was to keep as much out of the persistent view as possible. Would you like me to change it?

dretrex: Things associated with display should be in the view regardless of which view it is. Conversely, you could have a BPM view that does this for you.

Sent from my iPhone

On Apr 27, 2016, at 10:19 PM, David Leacock <notifications@github.com<mailto:notifications@github.com>> wrote:


In source/beamline/PGM/PGMBPMControl.cpp<https://github.com/acquaman/acquaman/pull/2213#discussion_r61371842>:

> +    int lowerBound = averageValue_ - variance_;
> +
> +    bool newState = (newValue < upperBound || newValue > lowerBound);
> +
> +    if (newState != state_){
> +
> +        state_ = newState;
> +        emit stateValidityChanged(state_);
> +    }
> +}
> +
> +void PGMBPMControl::setCurrentValue(double value){
> +
> +    currentValue_ = value;
> +    updateValidity(currentValue_);
> +    emit onBPMValueChanged(QString("%1 um").arg(value, 0, 'f', 0));


Got it. My thinking was to keep as much out of the persistent view as possible. Would you like me to change it?

dleacock: Duh, not sure why I didn't do that sooner. I put it all in PGMBPMStatusView.  Thank you.

source/beamline/PGM/PGMBPMControl.cpp
```diff
@@ -0,0 +1,41 @@
+#include "beamline/PGM/PGMBPMControl.h"
+
+PGMBPMControl::PGMBPMControl(const QString &name, const QString &pvName, int avgValue, int variance, QObject *parent, const QString description)
```
iainworkman: Can you use averageValue here instead of avgValue.

source/beamline/PGM/PGMBPMControl.cpp
```diff
@@ -0,0 +1,41 @@
...
+    averageValue_ = avgValue;
+    variance_ = variance;
+	validityState_ = true;
+
+	// Get updated value from AMReadOnlyPVControl.
+	connect(this, SIGNAL(valueChanged(double)), this, SLOT(onPVValueChanged(double)));
+
+}
+
+QString PGMBPMControl::valueStr() const
```
iainworkman: Can this also be valueString() rather than valueStr()

source/beamline/PGM/PGMBPMControl.h
```diff
@@ -0,0 +1,53 @@
...
+public:
+    /*!
+    \param name - A unique description of this control
+    \param pvName - The EPICS channel-access name for this Process Variable
+    \param avgValue - average beam position value, meaured in um. Provided by beamline staff.
+    \param variance - allowed value variance. Provided by beamline staff.
+    \param parent - QObject parent class
+    */
+    /// Constructor
+	PGMBPMControl(const QString &name, const QString &pvName, int avgValue, int variance, QObject *parent = 0, const QString description = "");
```
iainworkman: Same here with the avgValue

source/beamline/PGM/PGMBPMControl.cpp
iainworkman: In fact I'm wondering if we need to do the string conversion here at all. Shouldn't it be up to the receiver of the usual valueChanged() signal to do the conversion if they wish to?

source/ui/PGM/PGMBPMStatusView.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    setLayout(mainLayout);
+
+	// Set connections for PGMBpmContols
+	setupBPMConnections();
+}
+
+QLayout * PGMBPMStatusView::setupBPMDisplay(){
+
+    bpm10IDvalueX_ = new QLabel("   um");
```
iainworkman: I wonder if we could do this with a bunch of AMExtendedControlEditors, set to read only, rather than manually building the display ourselves?

source/ui/PGM/PGMBeamStatusView.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <QGridLayout>
+#include <QLabel>
+#include <QGroupBox>
+
+
+PGMBeamStatusView::PGMBeamStatusView(QWidget *parent) :
+    QWidget(parent)
+{
+    ringCurrentLabel_ = new QLabel("Ring Current: ");
```
iainworkman: Again, I'm wondering if we could do all this display with just a few generic widgets, rather than manually setting all the labels and values?

source/ui/PGM/PGMBPMStatusView.cpp
XuetaoLiu-CLS: @iainworkman it could be. This was the original changes @dleacock made. It would be great if we can use some standard AM UI components. But we can live like this so far.

iainworkman:
Few comments. 
Time Estimate: 20 minutes.

source/beamline/PGM/PGMBPMControl.cpp
XuetaoLiu-CLS: This would be a helper function for the receiver's convenience, if they want the value + unit format. If they want some thing else, they can listen to the valueChanged() signal and do it themselves.

source/ui/PGM/PGMBeamStatusView.cpp
XuetaoLiu-CLS: @iainworkman I would like to do more code refactor in #2355 to make this ring current display stuff universal to all the beamlines.

If you are not against it, we will live with this changes and refactor it later.

source/beamline/PGM/PGMBPMControl.cpp
```diff
@@ -1,10 +1,10 @@
 #include "beamline/PGM/PGMBPMControl.h"
 
-PGMBPMControl::PGMBPMControl(const QString &name, const QString &pvName, int avgValue, int variance, QObject *parent, const QString description)
+PGMBPMControl::PGMBPMControl(const QString &name, const QString &pvName, int averagValue, int variance, QObject *parent, const QString description)
```
iainworkman: typo in averageValue

PR #2214, opened 2016-04-05
Pull Request for Issue2089: Photon Shutter 2 button does becomes disabled after injection

dretrex:
AMProcessVariablePrivate has two connected signals and isn't using the "important" one for the `checkReadWriteReady()` method.


iainworkman:
Looks good to me.
Time Estimate: 1 minute.

I know this is difficult to test in relation to this particular issue, but is it worth testing to see if it has any adverse effects on PVs in the common case?

dretrex:
Since the check in readWriteCheck is wrapped in the:

```c++
if (value_ != newValue)
    value_ = newValue;
```

It shouldn't have any new effects due to extra signals firing.

PR #2215, opened 2016-04-06
Pull Request for Issue2199: VESPERS Mono can get stuck in Move Active

dretrex:
Implemented a fix that can get the VESPERS mono out of it's stalled state.

I also added an error monitor to AMDatabase to maybe catch the crashing bug that has been plaguing VESPERS for months.  Hopefully this will give us a clue to the root cause.

### DIFF
--- a/source/beamline/VESPERS/VESPERSAddOnsCoordinator.cpp
+++ b/source/beamline/VESPERS/VESPERSAddOnsCoordinator.cpp
@@ -188,6 +188,12 @@ VESPERSAddOnsCoordinator::VESPERSAddOnsCoordinator(QObject *parent)
 	oldRealWireXStatusControl_ = new AMReadOnlyPVControl("oldRealWireXFeedback", "SVM1607-2-B21-02:status", this);
 	oldRealWireYStatusControl_ = new AMReadOnlyPVControl("oldRealWireYFeedback", "SVM1607-2-B21-03:status", this);
 	oldRealWireZStatusControl_ = new AMReadOnlyPVControl("oldRealWireZFeedback", "SVM1607-2-B21-01:status", this);
+\
+	// Mono
+
+	monoFeedbackControl_ = new AMReadOnlyPVwStatusControl("monoFeedback", "SMTR1607-1-B20-20:mm:sp", "SMTR1607-1-B20-20:status", this);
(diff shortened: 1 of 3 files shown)


source/beamline/VESPERS/VESPERSAddOnsCoordinator.cpp
```diff
@@ -247,6 +253,7 @@ VESPERSAddOnsCoordinator::VESPERSAddOnsCoordinator(QObject *parent)
 
 	// Atto H, V, N
 
+/*
```
XuetaoLiu-CLS: do we want to keep those codes in the coordinator?

dretrex: Unfortunately, yes :(

source/beamline/VESPERS/VESPERSAddOnsCoordinator.cpp
```diff
@@ -1836,3 +1854,24 @@ void VESPERSAddOnsCoordinator::restoreAddOnsWireVAndNStatus()
 	if(!addOnsWireNStatusControl_->withinTolerance(finalStatus))
 		addOnsWireNStatusControl_->move(finalStatus);
 }
+
+void VESPERSAddOnsCoordinator::onMonoStatusChanged(bool isMoving)
+{
+	if (isMoving){
+
+		monoWaitTimer_->start(30000);
+		connect(monoFeedbackControl_, SIGNAL(valueChanged(double)), monoWaitTimer_, SLOT(start()));
```
XuetaoLiu-CLS: Question here. Should we connect to the `start()` signal or the `stop()` signal? I have thought that when Mono works as expected, it will always emit the valueChanged() signal.

XuetaoLiu-CLS:
Looks good other than a clarification.

Time Estimate: 5 minutes.

source/beamline/VESPERS/VESPERSAddOnsCoordinator.cpp
dretrex: Nope.  While the move is active, we need to restart the timer every time we get a value changed signal.  Otherwise, we could conceivably stop a valid move (if a move took longer than 30 seconds).  Since we know that the move is working correctly if the value changed signal is being emitted, we can just have the timer restart it's timer (since start() is used as a restart in the Qt API).

PR #2220, opened 2016-04-11
Pull Request for Issue2216: Scans stalling sometimes after first point

dretrex:
Removed some extra code that is not necessary for the proper initialization of scans for VESPERS anymore.

### DIFF
--- a/source/acquaman/VESPERS/VESPERSScanController.cpp
+++ b/source/acquaman/VESPERS/VESPERSScanController.cpp
@@ -46,22 +46,7 @@ AMAction3 *VESPERSScanController::buildBaseInitializationAction(double firstRegi
 	AMSequentialListAction3 *initializationAction = new AMSequentialListAction3(new AMSequentialListActionInfo3("VESPERS Scan Initialization Actions", "VESPERS Scan Initialization Actions"));
 	CLSSIS3820Scaler *scaler = VESPERSBeamline::vespers()->scaler();
 
-	AMListAction3 *stage1 = new AMListAction3(new AMListActionInfo3("VESPERS Initialization Stage 1", "VESPERS Initialization Stage 1"), AMListAction3::Sequential);
-	stage1->addSubAction(scaler->createContinuousEnableAction3(false));
-
-	AMListAction3 *stage2 = new AMListAction3(new AMListActionInfo3("VESPERS Initialization Stage 2", "VESPERS Initialization Stage 2"), AMListAction3::Sequential);
-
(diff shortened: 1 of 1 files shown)


source/acquaman/VESPERS/VESPERSScanController.cpp
```diff
@@ -46,22 +46,7 @@ AMAction3 *VESPERSScanController::buildBaseInitializationAction(double firstRegi
...
 	CLSSIS3820Scaler *scaler = VESPERSBeamline::vespers()->scaler();
 
-	AMListAction3 *stage1 = new AMListAction3(new AMListActionInfo3("VESPERS Initialization Stage 1", "VESPERS Initialization Stage 1"), AMListAction3::Sequential);
-	stage1->addSubAction(scaler->createContinuousEnableAction3(false));
-
-	AMListAction3 *stage2 = new AMListAction3(new AMListActionInfo3("VESPERS Initialization Stage 2", "VESPERS Initialization Stage 2"), AMListAction3::Sequential);
-
-	stage2->addSubAction(scaler->createStartAction3(false));
-	stage2->addSubAction(scaler->createScansPerBufferAction3(1));
-	stage2->addSubAction(scaler->createTotalScansAction3(1));
```
XuetaoLiu-CLS: We are also using those in the other beamlines. Won't it cause the same issue? 
And in SXRMB, the order of the actions are:
```
	scaler->createStartAction3(false);
	scaler->createContinuousEnableAction3(false);
	scaler->createScansPerBufferAction3(1);
	scaler->createTotalScansAction3(1);
```

Will this make the behavior different?

XuetaoLiu-CLS: And in REIXS, we didn't use `createStartAction3(false)` at all.

dretrex: Probably.  We should probably update the other applications as well.  Since @helfrij implemented the ContinuousMode control a long time ago, it should make this more robust overall.

dretrex: But that's fine because it is called in mode control action list.

XuetaoLiu-CLS: @dretrex are we going to update the other applications in this PR, or you prefer to a new issue?

dretrex: New issue.

XuetaoLiu-CLS:
Looks good. 

Time Estimate: 10 minutes

PR #2225, opened 2016-04-11
PullRequest for Issue483: remove the faciltyId for AMRun table

XuetaoLiu-CLS:
- AMRun table will not have the facilityId anymore
- AMFacility table will have only one entry
- db updgrade for AMFacility and AMRun

Waiting for Issue2218 to be merged...


XuetaoLiu-CLS:
@helfrij do you mind re-review this one? Thanks.
It has the changes of AMFacility and the DB upgrade to existing database

source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
```diff
@@ -0,0 +1,118 @@
...
+	return QStringList();
+}
+
+bool CLSDbUpgrade1Pt1::upgradeNecessary() const
+{
+	// run update when:
+	//
+	//   - the AMRun table has the "facilityId" column
+	//   - The AMFacility table has more than one record, or
+	//   - the name of the only facility doesn't match with the given beamline name, or
```
helfrij: This is kind of a roundabout question. For BioXAS, there are three separate facilities--BioXASSide, BioXASMain, BioXASImaging. Right now, I can sometimes run the same database using different facilities. Would that cause a problem with this database upgrade?

helfrij:
One minor request for clarification. Otherwise looks good!

Time Estimate: 10 minutes.

source/application/CLS/CLSAppController.h
```diff
@@ -10,37 +11,16 @@ class CLSAppController : public AMAppController
...
 	virtual void shutdown();
 
-public slots:
-	/// populate the data for some tables (AMFacility, AMRun) of the user database
-	virtual bool startupPopulateUserDBTable(AMDatabase* userDb); ///< Run on first time only
-
 protected:
-	/// returns the beamline name of a given beamline Id
-	QString beamlineName(CLSAppController::CLSBeamlineID beamline);
+	virtual AMFacility facility() const	{ return clsFacility_; }
```
dretrex: Spacing is weird here.

source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
dretrex: I believe we were thinking of making BioXAS just one facility a while ago.  Would it make sense?

To answer your question though, I don't believe it would matter for this database upgrade the way it is organized.

dretrex:
Overall, looks good.  I just wanted to double check the testing on this.  After you do the upgrade, have you exported scans, and created runs that make sense?
Time Estimate: 10 minutes

source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
helfrij: I think it would eventually make sense for BioXAS to be a single facility. With my question, was mostly concerned with whether what I was doing was going to break something, or if I should be more careful.

XuetaoLiu-CLS:
@helfrij @dretrex I made more changes to this PR and tested on SXRMB. 
It works as expected: 
1) when launching apps, the runs are loaded properly
2) create new run works as expected

source/dataman/AMScanDictionary.cpp
```diff
@@ -218,11 +218,8 @@ QString AMScanDictionary::krFacilityName(const QString& arg) {
 	if(!scan_)
 		return "[??]";
 
-	AMRun r;
-	r.loadFromDb(scan_->database(), scan_->runId());
-
 	AMFacility f;
-	f.loadFromDb(scan_->database(), r.facilityId());
+	f.loadFromDb(scan_->database(), 1); //AMFacility table will have only one entry, which is the definition of this facility
```
dretrex: Pedantically, this is not correct.  Wouldn't you want to get the id of the facility from the name (it is technically possible to have multiple facilities in a database (especially true with the current BioXAS setup I think).

source/dataman/export/AMExporter.cpp
```diff
@@ -293,11 +293,8 @@ QString AMExporter::krFacilityName(const QString& arg) {
 	if(!currentScan_)
 		return "[??]";
 
-	AMRun r;
-	r.loadFromDb(currentScan_->database(), currentScan_->runId());
-
 	AMFacility f;
-	f.loadFromDb(currentScan_->database(), r.facilityId());
+	f.loadFromDb(currentScan_->database(), 1); //AMFacility table will have only one entry, which is the definition of this facility
```
dretrex: Same thing here?

dretrex: Or use your new method in AMDbObject?

dretrex:
Some comments about loading the facility from the database.

source/dataman/AMScanDictionary.cpp
XuetaoLiu-CLS: @dretrex I think from this point forward, we will have only one facility in each DB and this is also what we are trying to do here (if this assumption doesn't stand, then we need to rework on the DBUpgrade and this piece of code).

Regarding BioXAS, I talked with @helfrij and just pushed one changes 29f48754de090a351824e540c2f339c2fe5449b4, which will make the BioXAS only one facility.

dretrex: It is probably fine for the time being.

dretrex:
So I think this is essentially good to go now, right @XuetaoLiu-CLS?  It's taken a little while to get there, but I think it's in a good place.  If you don't have any more comments then I'll merge it in.

XuetaoLiu-CLS:
@dretrex I have no comments :)

PR #2227, opened 2016-04-12
Pull Request for Issue2194: Prevent raw data sources from being exported by default

helfrij:
Changes:
- Modified the argument for `AMExporterOptionXDIFormat::setIncludeAllDataSources(...)` from true to false in BioXAS.h. We don't want to include all data sources. Commented out the call to `AMExporterOptionXDIFormat::setFirstColumnOnly(...)` as I'm not sure this does anything if the includeAllDataSources flag is false.
- Added method to AMExporterOptionGeneral to clear all data sources. This is used in BioXASXASScanActionController to prevent "old" data sources from being exported.
- Added calls throughout `BioXASXASScanActionController::buildScanControllerImplementation()` to add the appropriate data sources to the exporter option.

I think it's likely that the beamline staff will want some additional changes:
- The dark current corrected data sources have names that still have _DarkCorrect appended to the end.

These changes will be handled in a separate issue as they come up.


helfrij:

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
```diff
@@ -158,6 +174,8 @@ void BioXASXASScanActionController::buildScanControllerImplementation()
 		connect( BioXASBeamline::bioXAS()->i0Detector(), SIGNAL(darkCurrentValueChanged(double)), i0CorrectedDetectorSource, SLOT(setDarkCurrent(double)) );
 
 		scan_->addAnalyzedDataSource(i0CorrectedDetectorSource, true, false);
+
+		exportXDI->addDataSource(i0CorrectedDetectorSource->name(), false);
```
iainworkman: Just making sure... you did want the omit axis value argument for this one to be false, but all the others true?

iainworkman:
One small question. otherwise looks good to me.
Time Estimate: 10 minutes

source/acquaman/BioXAS/BioXASXASScanActionController.cpp
helfrij: Yes. What I'd wanted to achieve here was having the first column in the exported file be the X-axis value for the first data source (in this case the I0 corrected source). There is an exporter option setting for this behavior (`setFirstColumnOnly(bool)`?), but it doesn't appear to apply in cases where `includeAllDataSources()` is false, as it is here.

A better solution might be to add an implementation for this particular case.

source/application/BioXAS/BioXAS.h
```diff
@@ -32,8 +32,8 @@ namespace BioXAS
 		xasDefault->setColumnHeaderDelimiter("");
 		xasDefault->setSectionHeader("");
 		xasDefault->setSectionHeaderIncluded(true);
-		xasDefault->setIncludeAllDataSources(true);
-		xasDefault->setFirstColumnOnly(true);
+		xasDefault->setIncludeAllDataSources(false);
+		//xasDefault->setFirstColumnOnly(true); // Doesn't mean anything if all data sources aren't included.
```
dretrex: Probably don't need this comment anymore, right?

dretrex:
Looks good.  Just a trivial change.

Time Estimate: 1 minute

source/application/BioXAS/BioXAS.h
helfrij: Ha, that was a mental note to myself. Would it be a good idea to develop that 'setFirstColumnOnly' capability for the case where all data sources are NOT included?

dretrex: Yes, I would totally agree.  New issue?

helfrij: Sounds good.

PR #2236, opened 2016-04-14
Pull Request for Issue2235: Add storage ring current to persistent view.

helfrij:
Added control editor to the persistent view for the storage ring current.

### DIFF
--- a/source/ui/BioXAS/BioXASPersistentView.cpp
+++ b/source/ui/BioXAS/BioXASPersistentView.cpp
@@ -42,6 +42,7 @@ BioXASPersistentView::BioXASPersistentView(QWidget *parent) :
 
 	BioXASControlEditor *sr1CurrentEditor = new BioXASControlEditor(CLSStorageRing::storageRing()->ringCurrentControl());
 	sr1CurrentEditor->setTitle("SR1 Current");
+	sr1CurrentEditor->setReadOnly(true);
 
 	layout->addWidget(sr1CurrentEditor);
 
--- a/source/ui/BioXAS/BioXASPersistentView.cpp
+++ b/source/ui/BioXAS/BioXASPersistentView.cpp
@@ -38,6 +38,13 @@ BioXASPersistentView::BioXASPersistentView(QWidget *parent) :
 
 	setLayout(layout);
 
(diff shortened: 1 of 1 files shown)


helfrij:
Hey @XuetaoLiu-CLS, would you take a look?

source/ui/BioXAS/BioXASPersistentView.cpp
```diff
@@ -38,6 +38,13 @@ BioXASPersistentView::BioXASPersistentView(QWidget *parent) :
 
 	setLayout(layout);
 
+	// Create SR1 current view.
+
+	BioXASControlEditor *sr1CurrentEditor = new BioXASControlEditor(CLSStorageRing::storageRing()->ringCurrentControl());
+	sr1CurrentEditor->setTitle("SR1 Current");
```
XuetaoLiu-CLS: the default value of `readOnly_` is false. Do we need to set it to true?

XuetaoLiu-CLS:
Looks good other than a question on whether we should set the control to readonly.

Time Estimate: 5 minutes

source/ui/BioXAS/BioXASPersistentView.cpp
dretrex: Good point.

dretrex:
Time Estimate: 1 minute

source/ui/BioXAS/BioXASPersistentView.cpp
helfrij: I checked, because I know we've had issues with this in the past. This control editor is read-only. :)

dretrex: I can't seem to find where read-only is set to true.  Is there another mechanism that can set it to true?

helfrij: I'm not sure there needs to be an explicit setting in this case. The ring current control is a read-only PV control in CLSStorageRing, and the control editor checks whether the control can move before displaying the 'edit dialog'. If the control can't move, the editor is effectively read-only. Is the worry that read-only PV controls may inappropriately report as being able to move?

PR #2245, opened 2016-04-18
Pull Request for Issue2243: adding control move progress to control editors

helfrij:
- Created new class BioXASValueProgressLabel. This class inherits from QLabel with a couple of class variables related to displaying move progress: progressMin, progressValue, progressMax, displayProgress.
- Added BioXASValueProgressLabel to BioXASValueEditor. Added class variables for the progress values, with getters, setters, and signals.
- Added methods to BioXASControlEditor that update the progress min, max, value, and display flag in response to changes in the move status of the control.


helfrij:

source/ui/BioXAS/BioXASValueProgressLabel.cpp
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	painter.drawText(labelRect, Qt::AlignCenter, text());
+
+	painter.end();
+}
+
+double BioXASValueProgressLabel::progressPercentage(double min, double value, double max) const
+{
+	return abs(value - min) / abs(max - min);
```
dretrex: For information only! You could use qAbs() which would technically give you a platform independent abs function.

dretrex: Oh! You implemented your own abs.  I'd recommend just using qAbs().  Way simpler.

source/ui/BioXAS/BioXASControlEditor.cpp
```diff
@@ -262,6 +291,32 @@ void BioXASControlEditor::updateUnits()
...
+void BioXASControlEditor::updateProgressValueMaximum()
+{
+	if (control_ && useControlMovingAsProgress_ && control_->moveInProgress())
+		BioXASValueEditor::setProgressValueMaximum(control_->setpoint());
+}
+
+void BioXASControlEditor::updateProgressValue()
+{
+	if (control_ && useControlMovingAsProgress_ && control_->moveInProgress())
+		BioXASValueEditor::setProgressValue(control_->value());
```
dretrex: Why do you need to specifically scope the BioXASValueEditor?  I didn't notice any inheritance or conflicts in your changes.

dretrex:
There are merge conflicts in this guy after all the merging I did.  Looks good other than a couple small comments.

Time Estimate: 5 minutes

source/ui/BioXAS/BioXASValueProgressLabel.cpp
helfrij: Kk! I'll make the change.

source/ui/BioXAS/BioXASControlEditor.cpp
helfrij: There's no real need to scope it here. But the majority of methods are reimplemented between BioXASControlEditor and BioXASValueEditor. In this case, I think, it helps to be specific.

helfrij:
Thanks @dretrex. Merge conflicts should be resolved.

source/ui/BioXAS/BioXASControlEditor.cpp
dretrex: Ah, okay.

PR #2246, opened 2016-04-18
PullRequest for Issue2244: reorder the call-order of setupUserConfiguration()

XuetaoLiu-CLS:
@dretrex 
I changed the order back to what it was. For example, in VESPERS, it should be called after the setupUserInterface.


dretrex:
Thanks @XuetaoLiu-CLS.  Could you make a new issue to properly fix these?  This gets us back to the right place, but I'm sure we can have a cleaner solution.

Time Estimate: 1 minute

Issue #2059, opened 2016-02-25
IDEAS: Upgrade bugs and cleanup


epengr:
One issue to take care of a few little things that have cropped up after bringing IDEAS back up to Master branch:

- [x] AMExtendedControlEditors on IDEASPersistentView all allow only '-1' as input (done in #2060)
- [x] AMExtendedControlEditors on Sample alignment all allow only '-1' as input, except vacuum stage only allows 150 (done in #2060)
- [x]  Mono settling time needs tweaking to resolve stutter/noise for 1s dwell XANES
- [x] Not convinced that AM1DCalibrationAB is properly saving all properties (moved to #2256)
- [x] Generic scan needs cleanup actions
- [x] Transmission and reference checkboxes should be independent
- [x] EXAFS energy range label shows k values for end point (with eV as units)
- [x] Scaler not resetting properly after XANES scans. It's entering continuous mode, but the dwell time is not changed.
- [x] Scaler dwell for continuous mode needs to be 250 ms, not 100 ms - IOC is having issues with that speed
- [x] Default XANES region sets near edge too short
- [x] First point of (KETEK!) dwell time channel in XANES scans is the previous dwell time setting
- [x] "Beam is on!"/"Beam is off!" label on persistent view not responding - signal not generated? (Moved to #2110)
- [x] The scan select dialog presented when "Export Scan" from the KETEK panel is not populated, the filter has to be cleared for scans to appear (filter values need to be fixed, I assume) (Moved to #2106)
- [x] The peaking time drop box on the KETEK panel is not labelled.
- [x] XRF acquisition status on sidebar
- [x] Turn off higher order datasets for default export
- [x] XRF detector status in sidebar

dretrex:
For the generic scan needs cleanup actions, see #2055 and maybe have a chat with @helfrij.

helfrij:
Hey @epengr, I'm in the BioXAS main users area today if you want to chat.

epengr:
I think I'm on to something or the -1's.  My beamline energy control is reporting

canMove()=  false

so the AMExtendedControlEditor is setting itself read only.

epengr:
`canWrite()` comes down to `isConnected() && ca_write_access(chid_)`, and `isConnected() `is reporting false at the time the control is created.

helfrij:
[...]

Issue #2256, opened 2016-04-21
AM1DCalibrationAB is not saving all properties


epengr:
The AB has the following QProperties:
```
	Q_PROPERTY(QString dataName READ dataName WRITE setDataName)
	Q_PROPERTY(QString NormalizationName READ NormalizationName WRITE setNormalizationName)
	Q_PROPERTY(double energyCalibrationOffset READ energyCalibrationOffset WRITE setEnergyCalibrationOffset)
	Q_PROPERTY(double energyCalibrationScaling READ energyCalibrationScaling WRITE setEnergyCalibrationScaling)
	Q_PROPERTY(double energyCalibrationReference READ energyCalibrationReference WRITE setEnergyCalibrationReference)
	Q_PROPERTY(bool isTransmission READ isTransmission WRITE setIsTransmission)
	Q_PROPERTY(bool toEdgeJump READ toEdgeJump WRITE setToEdgeJump)
	Q_PROPERTY(int preEdgePoint READ preEdgePoint WRITE setPreEdgePoint)
	Q_PROPERTY(int postEdgePoint READ postEdgePoint WRITE setPostEdgePoint)
```

But only `isTransmission` and `toEdgeJump` are being saved/reloaded properly.

QDebug shows that the editor's constructor is trying to reload them properly, but the proper values aren't there.

epengr:
@dretrex, can you have a look at this - I'm at a loss.

epengr:
I stumbled across a couple scans that _do_ save these properties.  Other scan in the same user folder/DB don't work.

Also, if I add a new AM1DCalibrationAB to a scan, it will save it's properties _that one time_ but after reloading it it no longer saves any further changes.

Very weird.

@dretrex, any thoughts?

Issue #2055, opened 2016-02-25
Add AMBeamline getter for scan initialization and cleanup actions


helfrij:
As discussed in #2053, it could be really handy for beamline classes to provide the necessary actions for initializing and cleaning up after a scan. This could be implemented as a virtual getter in AMBeamline and reimplemented for each particular beamline subclass. The generic scan could then easily access and execute these actions.

helfrij:
Done in #2057.

Time Estimate: 1 hour.

Issue #1158, opened 2015-01-22
AM1DCalibrationAB normalization glitch


epengr:
I have to investigate further, but the edge jump normalization seems to have a glitch where it where it will happily normalize by a negative value when the tail of the scan is lower than the start.  This can cause some wild swings during the pre-edge, and can invert the odd EXAFS spectra.  

Also, the plot axis aren't updating properly, there may be some missing signals.

dretrex:
@iainworkman @XuetaoLiu-CLS @helfrij this issue also belongs to IDEAS work.  It isn't flagged as such since the issue is on an AM class, but it was built for IDEAS.

epengr:
Thanks @dretrex, no reason not to flag it as IDEAS since that's where this glitch shows up, but I use this AB on REIXS as well.   It's getting pretty slick - it may be of interest to others...

epengr:
completed in #2059

PR #2257, opened 2016-04-21
Pull Request for Issue2059: IDEAS: Upgrade bugs and cleanup

epengr:
Numerous little bug fixes and tweaks.

See #2059 for list.


epengr:
Note:  #2055 was merged into this branch as a Hotfix

epengr:
@helfrij, I think you helped with several of these, would you mind looking it over for me?

source/acquaman/IDEAS/IDEASXASScanActionController.cpp
```diff
@@ -232,7 +238,9 @@ AMAction3* IDEASXASScanActionController::createCleanupActions()
 {
 	AMListAction3 *cleanupActions = new AMListAction3(new AMListActionInfo3("IDEAS XAS Cleanup Actions", "IDEAS XAS Cleanup Actions"));
 
-	cleanupActions->addSubAction(IDEASBeamline::ideas()->scaler()->createDwellTimeAction3(0.1));
+	cleanupActions->addSubAction(new AMWaitAction(new AMWaitActionInfo(IDEASBeamline::ideas()->scaler()->dwellTime())));
```
helfrij: Just curious--what's the reason for adding the extra waits here?

source/analysis/AM1DCalibrationAB.cpp
```diff
@@ -358,7 +358,7 @@ void AM1DCalibrationAB::computeCachedValues() const
 			scale = 1;
 
 		    else
-			scale = (data.at(postEdgePointValue)/normalizer.at(postEdgePointValue)) - offset;
+			scale = qAbs((data.at(postEdgePointValue)/normalizer.at(postEdgePointValue)) - offset);
```
helfrij: Is this change to prevent the scale from being negative?

source/ui/IDEAS/IDEASPersistentView.cpp
```diff
@@ -73,8 +73,11 @@ IDEASPersistentView::IDEASPersistentView(QWidget *parent) :
 
 	monoCrystal_ = new QLabel("Crystal");
 	monoCrystal_->setAlignment(Qt::AlignCenter);
-	monoEnergyRange_ = new QLabel(" eV - eV");
-	monoEnergyRange_->setAlignment(Qt::AlignCenter);
+
+	//*****************TEST: MONO TROUBLESHOOTING TOOL, WILL NEED IN FUTURE
```
helfrij: Would you like to keep these QDebugs for now?

source/ui/IDEAS/IDEASPersistentView.cpp
```diff
@@ -226,15 +237,22 @@ IDEASPersistentView::IDEASPersistentView(QWidget *parent) :
...
 	QVBoxLayout *mainPanelLayout = new QVBoxLayout;
 	mainPanelLayout->addWidget(ringCurrent_);
 	mainPanelLayout->addLayout(beamChangeLayout);
 	mainPanelLayout->addWidget(beamStatusLabel_, 0, Qt::AlignCenter);
 	mainPanelLayout->addLayout(monoEnergyLayout);
-	mainPanelLayout->addWidget(monoCrystal_);
-	mainPanelLayout->addWidget(monoEnergyRange_);
+	mainPanelLayout->addLayout(KETEKStatusLayout);
+
+	//mainPanelLayout->addWidget(monoSettlingTime_); //MONO TROUBLESHOOTING TOOL, WILL NEED IN FUTURE
```
helfrij: Same QDebug question here too.

source/ui/IDEAS/IDEASPersistentView.cpp
```diff
@@ -323,13 +340,19 @@ void IDEASPersistentView::onCalibrateClicked()
 void IDEASPersistentView::onScalerContinuousButtonClicked()
 {
 	AMListAction3 *scalerContinuousEnableActions = new AMListAction3(new AMListActionInfo3("Enable Scaler Continuous Mode", "Enable Scaler Continuous Mode"));
-	scalerContinuousEnableActions->addSubAction(new AMWaitAction(new AMWaitActionInfo(0.25)));
+	scalerContinuousEnableActions->addSubAction(new AMWaitAction(new AMWaitActionInfo(0.5)));
```
helfrij: Are these waits a necessary step for changing the scaler single shot/continuous state?

helfrij:
Couple of minor questions.

Time Estimate: 10 minutes.

source/acquaman/IDEAS/IDEASXASScanActionController.cpp
epengr: The scaler was usually not re-entering continuous mode after a scan and this resolved that issue.  I don't know why, but it works.

source/analysis/AM1DCalibrationAB.cpp
epengr: When the first point is higher than the last point, the spectra end up inverted.  This was a failed attempt to resolve that issue.  I didn't realize that I had left it in here.  I was just about to tackle this problem in #1158. I'll maybe have a quick look at it here.

source/ui/IDEAS/IDEASPersistentView.cpp
epengr: Once CID has time to investigate this odd mono behavior with me, I will likely need this troubleshooting tool.  It's handy to leave it here.

source/analysis/AM1DCalibrationAB.cpp
helfrij: Ah okay.

source/ui/IDEAS/IDEASPersistentView.cpp
helfrij: Sounds good.

source/ui/IDEAS/IDEASPersistentView.cpp
epengr: yep.

source/ui/IDEAS/IDEASPersistentView.cpp
epengr: Only to get back into continuous, the seem to be needed for some reason.

source/acquaman/IDEAS/IDEASXASScanActionController.cpp
helfrij: Hmmm. This could indicate a problem with how the change between these states is implemented in CLSSIS3820Scaler.

epengr: I thought @dretrex mentioned that this wasn't the only implementation that still needed strategic waits even after implementing the new continuous mode switching method?  I could be wrong.  But that would just support your suggestion that something needs to be tweaked in CLSSIS3820Scaler.

source/analysis/AM1DCalibrationAB.cpp
epengr: Thanks for drawing this to my attention!  The only reason this did solve the problem is that I had fixed the fluorescent mode algorithm and didn't fox the transmission mode one.  I did so and pushed up that change.

source/analysis/AM1DCalibrationAB.h
```diff
@@ -40,7 +40,7 @@ class AM1DCalibrationAB : public AMStandardAnalysisBlock
 	Q_OBJECT
 
 	Q_PROPERTY(QString dataName READ dataName WRITE setDataName)
-	Q_PROPERTY(QString normalizationName READ normalizationName WRITE setNormalizationName)
+	Q_PROPERTY(QString NormalizationName READ normalizationName WRITE setNormalizationName)
 	Q_PROPERTY(double energyCalibrationOffset READ energyCalibrationOffset WRITE setEnergyCalibrationOffset)
```
epengr: Found a fixed one bug that only showed up when testing on an existing DB. changing the case of this name triggers a DB upgrade that fails because the name already exists (I wouldn't actually want a DB upgrade anyways - I can live with the capital), causing AM to exit with code 255. 

@dretrex, is that a bug in the DB upgrade method?  Should be be case smart?  Or maybe it should just be used by competent coders...

source/acquaman/IDEAS/IDEASXASScanActionController.cpp
helfrij: I know the BioXAS scaler uses strategic waits when manipulating that PV, but for us it toggles between 'Armed/Disarmed' states instead of 'SingleShot/Continuous.' But the wait is still there. Maybe we should take another look at this scaler behavior, as part of another issue?

epengr:
I'm happy with this now, if you are.

Time estimate: 1 hour

source/analysis/AM1DCalibrationAB.h
```diff
@@ -40,7 +40,7 @@ class AM1DCalibrationAB : public AMStandardAnalysisBlock
 	Q_OBJECT
 
 	Q_PROPERTY(QString dataName READ dataName WRITE setDataName)
-	Q_PROPERTY(QString NormalizationName READ NormalizationName WRITE setNormalizationName)
+	Q_PROPERTY(QString NormalizationName READ normalizationName WRITE setNormalizationName)
```
helfrij: I like that these properties are now the proper camel-case. Did this change solve a particular problem (I'm thinking of #2256)?

source/acquaman/IDEAS/IDEASXASScanActionController.cpp
dretrex: Strategic waits make me go 😢

source/analysis/AM1DCalibrationABEditor.cpp
```diff
@@ -91,7 +91,7 @@ AM1DCalibrationABEditor::AM1DCalibrationABEditor(AM1DCalibrationAB *analysisBloc
 	if (analysisBlock_->inputDataSourceCount() > 0){
 
 		dataNames_->setCurrentIndex(dataNames_->findData(analysisBlock_->dataName()));
-		NormalizationNames_->setCurrentIndex(NormalizationNames_->findData(analysisBlock_->NormalizationName()));
+		NormalizationNames_->setCurrentIndex(NormalizationNames_->findData(analysisBlock_->normalizationName()));
```
dretrex: Camel case issue with variable name.

source/ui/IDEAS/IDEASPersistentView.cpp
dretrex: Not necessary for this pull request, but maybe we should try looking into that again?

dretrex:
Couple small comments.  Looks good otherwise.
Time Estimate: 10 minutes

source/analysis/AM1DCalibrationAB.h
epengr: No, but when I check to make sure they didn't, I learned something odd - possibly helpful?

See #2256

source/analysis/AM1DCalibrationABEditor.cpp
epengr: Fixed.

epengr:
@dretrex, read to go, I believe.

Time Estimate: 5 min

PR #2259, opened 2016-04-22
Pull Request for Issue2107: Adding output controls to BioXASZebra.

helfrij:
- Created new class BioXASZebraOutputControl, contains PV controls for the output value and status.
- Created new class BioXASZebraOutputControlView, a visualization for the output control.
- Created new class BioXASSideZebra and added to BioXASSideBeamline. This class has the output controls used on Side.
- Created new class BioXASMainZebra and added to BioXASMainBeamline. This class has the output controls used on Main, different from Side.
- Modified BioXASSide- and BioXASMainBeamlines to include the preferred values for the output controls, set on startup and enforced when the connected state of the control changes.

### DIFF
--- a/source/beamline/BioXAS/BioXASMainBeamline.cpp
+++ b/source/beamline/BioXAS/BioXASMainBeamline.cpp
@@ -255,7 +255,7 @@ bool BioXASMainBeamline::useLytleDetector(bool useDetector)
 
 	return result;
 }
-#include <QDebug>
+
 void BioXASMainBeamline::setupComponents()
 {
 	// SOE shutter.
@@ -343,46 +343,6 @@ void BioXASMainBeamline::setupComponents()
 
 	zebra_ = new BioXASMainZebra("TRG1607-701", this);
 	connect(zebra_, SIGNAL(connectedChanged(bool)), this, SLOT(updateConnected()));
-
-	BioXASZebraPulseControl *pulse1 = zebra_->pulseControlAt(0);
-	if (pulse1)
-		pulse1->setEdgeTriggerPreference(0);
-
-	BioXASZebraPulseControl *pulse3 = zebra_->pulseControlAt(2);
-	if (pulse3)
-		pulse3->setEdgeTriggerPreference(0);
-
-	BioXASZebraSoftInputControl *softIn1 = zebra_->softInputControlAt(0);
-	if (softIn1)
-		softIn1->setTimeBeforeResetPreference(0.01);
-
-	BioXASZebraSoftInputControl *softIn3 = zebra_->softInputControlAt(2);
-	if (softIn3)
-		softIn3->setTimeBeforeResetPreference(0.01);
-
-	BioXASZebraOutputControl *out1TTL = zebra_->outputControlAt(0);
-	if (out1TTL)
-		out1TTL->setOutputValuePreference(54); // The OUT1TTL output should connect to PULSE3, or the Ge detector.
-	else
-		qDebug() << "OUT1TTL control is invalid, cannot set value preference.";
-
-	BioXASZebraOutputControl *out1NIM = zebra_->outputControlAt(1);
-	if (out1NIM)
-		out1NIM->setOutputValuePreference(37); // The OUT1NIM output should connect to OR2, or the scaler.
-	else
-		qDebug() << "OUT1NIM control is invalid, cannot set value preference.";
-
-	BioXASZebraOutputControl *out2TTL = zebra_->outputControlAt(2);
-	if (out2TTL)
-		out2TTL->setOutputValuePreference(54); // The OUT2TTL output should connect to PULSE3, or the Ge detector.
-	else
-		qDebug() << "OUT2TTL control is invalid, cannot set value preference.";
-
-	BioXASZebraOutputControl *out3TTL = zebra_->outputControlAt(3);
-	if (out3TTL)
-		out3TTL->setOutputValuePreference(32); // The OUT3TTL output should connect to AND1, or the fast shutter.
-	else
-		qDebug() << "OUT3TTL control is invalid, cannot set value preference.";
 
 	// The Zebra trigger source.
 
--- a/source/beamline/BioXAS/BioXASMainBeamline.cpp
+++ b/source/beamline/BioXAS/BioXASMainBeamline.cpp
@@ -341,7 +341,7 @@ void BioXASMainBeamline::setupComponents()
 
 	// Zebra.
 
-	zebra_ = new BioXASZebra("TRG1607-701", this);
+	zebra_ = new BioXASMainZebra("TRG1607-701", this);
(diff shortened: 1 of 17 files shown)


helfrij:

source/beamline/BioXAS/BioXASMainBeamline.cpp
```diff
@@ -255,7 +255,7 @@ bool BioXASMainBeamline::useLytleDetector(bool useDetector)
 
 	return result;
 }
-
+#include <QDebug>
```
XuetaoLiu-CLS: Do we want to keep this?

source/beamline/BioXAS/BioXASMainBeamline.cpp
```diff
@@ -360,6 +360,30 @@ void BioXASMainBeamline::setupComponents()
 	if (softIn3)
 		softIn3->setTimeBeforeResetPreference(0.01);
 
+	BioXASZebraOutputControl *out1TTL = zebra_->outputControlAt(0);
```
XuetaoLiu-CLS: is it possible to encrypt these configuration inside `BioXASMainZebra`? Those configurations look like specific to the BioXASMainZebra.

source/beamline/BioXAS/BioXASMainBeamline.cpp
helfrij: Nope, I'll remove.

source/beamline/BioXAS/BioXASMainBeamline.cpp
helfrij: That's a good idea. I'll move them inside, and do the same for the Side Zebra too.

source/beamline/BioXAS/BioXASSideBeamline.cpp
```diff
@@ -444,6 +444,18 @@ void BioXASSideBeamline::setupComponents()
 	if (xspress3Pulse)
 		xspress3Pulse->setInputValuePreference(52); // The Xspress3 detector can be triggered using pulse 1 (#52).
 
+	BioXASZebraOutputControl *out1TTL = zebra_->outputControlAt(0);
```
XuetaoLiu-CLS: same question here. I think they should be part of `BioXASSideZebra`

source/beamline/BioXAS/BioXASZebra.cpp
```diff
@@ -131,8 +146,12 @@ void BioXASZebra::onConnectedChanged()
 	foreach (AMControl *orBlock, orBlocks_)
 		connected &= orBlock->isConnected();
 
+	foreach (AMControl *outputControl, outputControls_)
+		connected &= outputControl->isConnected();
+
 	if (connected_ != connected){
 		connected_ = connected;
+		qDebug() << (connected_ ? "Zebra is connected." : "Zebra is NOT connected.");
```
XuetaoLiu-CLS: if we really want this message, it will be better to use AMErrorMon::debug()

source/beamline/BioXAS/BioXASZebra.cpp
```diff
@@ -131,8 +146,12 @@ void BioXASZebra::onConnectedChanged()
 	foreach (AMControl *orBlock, orBlocks_)
 		connected &= orBlock->isConnected();
 
+	foreach (AMControl *outputControl, outputControls_)
+		connected &= outputControl->isConnected();
```
XuetaoLiu-CLS: Just a suggestion. Will it make the codes looks better if we use `AMControlSet` here? If we put those controls into different AMControlSet, then we can use the `isConnected()` method and the `connected()` signal of AMControlSet

helfrij: That would simplify some things, for sure. Right now `BioXASZebraPulseControl` does not inherit from AMControl, so we'd have to refactor that first.

source/ui/BioXAS/BioXASZebraView.cpp
```diff
@@ -3,9 +3,10 @@
 #include "ui/BioXAS/BioXASZebraPulseControlView.h"
 #include "ui/beamline/AMExtendedControlEditor.h"
 #include "ui/BioXAS/BioXASZebraLogicBlockView.h"
+#include "ui/BioXAS/BioXASZebraOutputControlView.h"
 
 #include <QGridLayout>
-
+#include <QDebug>
```
XuetaoLiu-CLS: this can be removed for sure.

source/beamline/BioXAS/BioXASZebra.cpp
XuetaoLiu-CLS: I see. I don't about that.

XuetaoLiu-CLS:
Time Estimate: 30 minutes

source/beamline/BioXAS/BioXASSideBeamline.cpp
```diff
@@ -403,67 +403,14 @@ void BioXASSideBeamline::setupComponents()
...
-
 	// The Zebra trigger source.
 
 	zebraTriggerSource_ = new AMZebraDetectorTriggerSource("ZebraTriggerSource", this);
 	zebraTriggerSource_->setTriggerControl(zebra_->softInputControlAt(0));
 
 	// Scaler.
 
-	scaler_ = new BioXASSIS3820Scaler("MCS1607-601:mcs", softIn3, this);
+	scaler_ = new BioXASSIS3820Scaler("MCS1607-601:mcs", zebra_->softInputControlAt(2), this);
```
XuetaoLiu-CLS: it might be more readable if we can have a function like `zebra_->fastShutterSoftInputControl()`.

helfrij: That could definitely make the function calls more intuitive. I'll have to think about that. If it's okay, I think this change would be best handled in a separate issue, as part of a future Zebra refactor.

XuetaoLiu-CLS: sure. we can change it later.

dretrex: As it stands right now, the zebra is made to be about as general as possible.  This would be something where we would have a BioXAS subclass that would have convenience getters.

source/beamline/BioXAS/BioXASMainBeamline.h
```diff
@@ -112,7 +113,7 @@ class BioXASMainBeamline : public BioXASBeamline
 	virtual BioXAS32ElementGeDetector* ge32DetectorOutboard() const { return 0; }
 
 	/// Returns the zebra control box.
-	virtual BioXASZebra *zebra() const { return zebra_; }
+	virtual BioXASMainZebra *zebra() const { return zebra_; }
```
dretrex: Or specifically in these subclasses you could have the convenience getters (if you were to do it in a separate issue).

dretrex:
Looks good to me.
Time Estimate: 10 minutes

PR #2266, opened 2016-04-25
Pull Request for Issue2264: Adding limits to select motors on Side

helfrij:
Changes:
- Moved `AMPseudoMotorControl::setMinimum(...)` and `setMaximum(...)` methods to public slots, from protected slots.
- Added `AMReadOnlyPVControl::setMinimum(...)` and `setMaximum(...)` methods. These call the existing protected `setLowLimitValue(...)/setHighLimitValue(...)` slots, and additionally set a flag that causes limit updates from PVs to be rejected.
- Added limits.

Testing involved attempting to input values inside and outside the value ranges using editors. Control editors should reject input for values outside the desired range.
- [x] M1 mirror PV controls
- [x] M1 mirror pseudomotors
- [x] Mono PV controls
- [x] M2 mirror PV controls
- [x] M2 mirror pseudomotors

### DIFF
--- a/source/beamline/AMPVControl.cpp
+++ b/source/beamline/AMPVControl.cpp
@@ -37,6 +37,9 @@ AMReadOnlyPVControl::AMReadOnlyPVControl(const QString& name, const QString& rea
 	lowLimitPV_ = 0;
 	highLimitPV_ = 0;
 
+	allowLowLimitValuePVUpdates_ = true;
+	allowHighLimitValuePVUpdates_ = true;
+
 	lowLimitValue_ = -1;
 	highLimitValue_ = -1;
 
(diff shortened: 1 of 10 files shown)


helfrij:
Hey @iainworkman, would you take a look?

source/beamline/AMPVControl.h
```diff
@@ -114,6 +114,12 @@ class AMReadOnlyPVControl : public AMControl {
 	///  This extra signal is specialized to report on PV channel connection status.  You should be free to ignore it and use the signals defined in AMControl.
 	void readConnectionTimeoutOccurred();
 
+public slots:
+	/// Sets the minimum value.
+	void setMinimumValue(double newValue);
+	/// Sets the maximum value.
+	void setMaximumValue(double newValue);
```
iainworkman: I do wonder if these functions should be given a name which more reflects what is going on here. Essentially these functions are used for acquaman to manually take control of the limits. Once these have been called acquaman will now ignore the limits which come in from epics land. 

Maybe
```C++
void setMinimumValueOverride(double newValue);
void setMaximumValueOverride(double newValue);
```
Or something of that nature???

source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -299,6 +299,22 @@ AMControl::FailureExplanation AMPseudoMotorControl::calibrate(double oldValue, d
 	return AMControl::NoFailure;
 }
 
+void AMPseudoMotorControl::setMinimumValue(double newValue)
+{
+	if (minimumValue_ != newValue) {
+		minimumValue_ = newValue;
```
iainworkman: Should there be a check here if the manually set override minimum is greater than any value which has come from epics? It seems like it might be a bad idea to allow acquaman to override a limit with a value which is less constraining than the one which comes from epics?

I mean, the motor still won't get there if we tell it to go beyond the limits. The setpoint will get pushed to epics, which will simply stop it at the limit value. Maybe an AMWarning?

source/beamline/AMPVControl.cpp
```diff
@@ -92,6 +95,18 @@ void AMReadOnlyPVControl::onReadPVError(int errorCode) {
 	}
 }
 
+void AMReadOnlyPVControl::setMinimumValue(double newValue)
+{
+	setLowLimitValue(newValue);
```
iainworkman: Should there be a check here if the manually set override minimum is less than any value which has come from epics? It seems like it might be a bad idea to allow acquaman to override a limit with a value which is less constraining than the one which comes from epics?

I mean, the motor still won't get there if we tell it to go beyond the limits, so it's not all that bad. The setpoint will get pushed to epics, which will simply stop it at the limit value. Maybe an AMWarning?

source/beamline/AMPVControl.cpp
```diff
@@ -108,19 +123,23 @@ void AMReadOnlyPVControl::onReadPVInitialized() {
...
+	if (allowHighLimitValuePVUpdates_)
+		setHighLimitValue(readPV_->upperGraphicalLimit());
 }
 
-void AMReadOnlyPVControl::onLowLimitPVInitialized()
+void AMReadOnlyPVControl::onLowLimitPVValueChanged()
 {
-	if (lowLimitPV_ && lowLimitPV_->isInitialized())
+	if (lowLimitPV_ && lowLimitPV_->isInitialized() && allowLowLimitValuePVUpdates_)
 		setLowLimitValue(lowLimitPV_->getDouble());
```
iainworkman: Again, I wonder if these should check for the case that allowLowLimitValuePVUpdates_ is set to false (ie we've manually set the limits) but the incoming value had been set to a more constraining value?

iainworkman:
Looks good. Couple of questions regarding how to handle manual limits being less constraining than the epics ones.
Time Estimate: 15 minutes

source/beamline/AMPVControl.cpp
helfrij: Ah, that's interesting. I hadn't thought of that!

source/beamline/AMPVControl.h
helfrij: Those names are probably clearer, especially as we aren't actually pushing limits to the EPICS layer. There would need to be a distinction between setting the Acquaman/softer-limits and the EPICS/soft-limits, whenever we develop that capability.

dretrex:
Looks good.
Time Estimate: 5 minutes

PR #2273, opened 2016-04-26
Pull Request for Export Issue 2106: XRF button bring up unpopulated scan browser

epengr:
Changed AMXRFDetailedDetectorView::onSaveButtonClicked() so that RegEx filters on scan name instead of type and only shows XRF scans with the default name "XRF Scan - <detector>" (i.e. unexported
scans)

### DIFF
--- a/source/ui/beamline/AMXRFDetailedDetectorView.cpp
+++ b/source/ui/beamline/AMXRFDetailedDetectorView.cpp
@@ -534,8 +534,8 @@ void AMXRFDetailedDetectorView::onSaveButtonClicked()
 		connect(chooseScanDialog_, SIGNAL(accepted()), this, SLOT(exportScan()));
 	}
 
-	chooseScanDialog_->setFilterKeyColumn(4);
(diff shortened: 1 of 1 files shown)


epengr:
@iainworkman, you're the obvious choice to review this - would you mind?

iainworkman:
I assume this name applies to all XRF scans?
Other than that question, looks good.
Time Estimate: 2 minutes.

dretrex:
I believe that will be alright.  We'll just blame @epengr if it doesn't work 😉 
Time Estimate: 1 minute

epengr:
AMXRFDetailedDetectorView auto-generates that exact default name based on the detector it's given, so this will only _not_ work if someone has re-implemented it and fiddled with things.  

If it doesn't work, I'm not the one that merged it  ;)

dretrex:
That's not how shifting blame works around here ;)

epengr:
So FIFO, not LIFO?

PR #2280, opened 2016-04-29
Pull Request for Issue1852: REIXS In-Plane Sample Motor

iainworkman:
- Created a new class REIXSSampleMotor, which rotates its coordinate system as the Z axis rotates, to create in plane sample motions.
- Added two instances of the REIXSSampleMotor (for sample horizontal and sample normal) to the REIXSSampleChamber class within REIXSBeamline.h/cpp
- Altered the REIXSSampleChamberButtonPanel to connect the in/out and right/left buttons to the new sample horizontal and normal controls.

Few changes that came along for the ride:
- Fixed a bug in the new build system where qmake syntax quirks were causing the include qmobility aspect to not function correctly.
- Altered the REIXSSampleChamberButtonPanel so that buttons are disabled when their control is moving.
- Fixed a bug where the AMPVwStatusAndUnitConversionControl was using the write converter instead of the read converter when fetching min and max values from the PV.


iainworkman:
Hey @XuetaoLiu-CLS,
Would you mind having a look at this for me? 
Thanks

source/beamline/REIXS/REIXSSampleMotor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+	angleOffset_ = 0;
+	connectedOnce_ = false;
+
+	addChildControl(horizontalTranslationControl_);
+	addChildControl(normalTranslationControl_);
+	addChildControl(verticalRotationControl_);
+
+	connect(verticalRotationControl_, SIGNAL(valueChanged(double)),
+	        this, SLOT(onVerticalAxisRotated(double)));
```
XuetaoLiu-CLS: You might NOT need this signal/slot connection. In function `addChildControl()`, we connected signal `valueChanged()` with slot `updateValue()`.

source/beamline/REIXS/REIXSSampleMotor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+double REIXSSampleMotor::totalRotation() const
+{
+	if(connected_) {
+
+		double combinedRotation = verticalRotationControl_->value() + angleOffset_;
+
+		while(combinedRotation > 360) {
+			combinedRotation -= 360;
+		}
```
XuetaoLiu-CLS: this sounds like a `fmod()` function call, like  `combinedRotation = fmod(combinedRotation, 360);`. I'm not sure whether we can use `fmod()` in qt easily.

source/beamline/REIXS/REIXSSampleMotor.cpp
```diff
@@ -0,0 +1,377 @@
...
+{
+	AMPseudoMotorControl::onMoveSucceeded(action);
+	updateMinimumAndMaximum();
+}
+
+AMAction3 * REIXSSampleMotor::createMoveAction(double setpoint)
+{
+
+	AMListAction3* moveAction = new AMListAction3(new AMListActionInfo3(QString("In-plane sample move"),
+	                                                                   QString("In-place sample move")),
```
XuetaoLiu-CLS: `In-plane` or `in-place`?

source/beamline/REIXS/REIXSSampleMotor.cpp
```diff
@@ -0,0 +1,377 @@
...
+	case AMMotorGroupObject::VerticalMotion:
+		return vector.z();		// This shouldn't happen, but we include it anyway.
+	case AMMotorGroupObject::NormalMotion:
+		return vector.y();
+	}
+
+	return 0;
+}
+
+double REIXSSampleMotor::radiansToDegrees(double radians)
```
XuetaoLiu-CLS: BioXAS Main and BioXAS Side also have the same implementation. Maybe we should create the utility class `AMTransformer`?

source/beamline/REIXS/REIXSSampleMotor.cpp
```diff
@@ -0,0 +1,377 @@
...
+}
+
+bool REIXSSampleMotor::validValue(double value) const
+{
+	QVector3D globalVector = createGlobalMovementVector(value);
+
+	return globalVector.x() >= REIXS_SAMPLE_HORIZONTAL_MIN &&
+	        globalVector.x() <= REIXS_SAMPLE_HORIZONTAL_MAX &&
+	        globalVector.y() >= REIXS_SAMPLE_NORMAL_MIN &&
+	        globalVector.y() <= REIXS_SAMPLE_NORMAL_MAX;
```
XuetaoLiu-CLS: when looking into the `source/util` folder, I found that we have `AMRange`, which has a function `withinRange()` for this :8ball:

source/beamline/REIXS/REIXSSampleMotor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+			max1 = (REIXS_SAMPLE_HORIZONTAL_MAX - currentPrimePosition.x() * cosRotation) / sinRotation;
+			max2 = (REIXS_SAMPLE_NORMAL_MAX + currentPrimePosition.x() * sinRotation) / cosRotation;
+		}
+
+	}
+
+	// The limit is the more restrictive for each value
+	setMaximumValue(qMin(max1, max2));
+	setMinimumValue(qMax(min1, min2));
```
XuetaoLiu-CLS: as mentioned above, it might be better to make use of the AMRange util class

source/ui/REIXS/REIXSSampleChamberButtonPanel.cpp
```diff
@@ -59,10 +57,69 @@ REIXSSampleChamberButtonPanel::~REIXSSampleChamberButtonPanel()
...
 {
 	REIXSSampleChamber* chamber = REIXSBeamline::bl()->sampleChamber();
-	chamber->x()->stop();
-	chamber->y()->stop();
-	chamber->z()->stop();
-	chamber->r()->stop();
+	chamber->beamNormalTranslation()->stop();
+	chamber->beamHorizontalTranslation()->stop();
+	chamber->beamVerticalTranslation()->stop();
+	chamber->beamVerticalRotation()->stop();
```
XuetaoLiu-CLS: why don't we add a "stop()" function for `REIXSSampleChamber`, which will call the stop functions of those motors?

XuetaoLiu-CLS:
Looks good other than some code refactor comments

Time Estimate: 30 minutes

source/beamline/REIXS/REIXSSampleMotor.cpp
iainworkman: There is a class in beamline/camera called AMAngle but it would bloat the code a little in here if I was to use it. To be honest I think this is a case where a namespace header file with a bunch of angle helper functions would be more useful than a class.

source/beamline/REIXS/REIXSSampleMotor.cpp
iainworkman: This is connected up here because the recalculation of the coordinate system is only required when the verticalRotationControl_ alters its value. If I put the same code it would also happen when the either the horizontal and normal motors performed a translation (which doesn't require a recalculation of the system rotation).

source/beamline/REIXS/REIXSSampleMotor.cpp
iainworkman: I don't think AMRange would help here. I'd still have to check which is the more limiting for each of the potential min/max values. I'd end up with something like:
```C++
    ...
    AMRange newLimits(qMax(min1, min2), qMin(max1, max2));
    setMaximumValue(newLimits.minimum(), newLimits.maximum());
```

which doesn't really make things that much clearer.

source/ui/REIXS/REIXSSampleChamberButtonPanel.cpp
iainworkman: Good idea.

source/beamline/REIXS/REIXSSampleMotor.cpp
XuetaoLiu-CLS: "a namespace header file" should work or a utility class with static methods only

iainworkman:
Okay, I've fixed this up and made a few more changes.
This should be good for review again.

XuetaoLiu-CLS:
Still looks good.

dretrex:
Looks good.
Time Estimate: 10 minutes

Issue #2194, opened 2016-03-31
Add option to add detector to a configuration without exporting it 


helfrij:
The BioXAS team wants to be able to add I0, I1, I2 detectors to a scan configuration and have the ability to use processed I0, I1, and I2 data in the scan, instead of the raw data.

dretrex:
Hey @helfrij, I renamed the issue to better reflect what they meant.  Since it isn't possible to include a detector without having it as a data source (you'd never be able to use analysis blocks that way), it seems like they just don't want non-dark current corrected datasources as columns in the exported file.

helfrij:
Sounds good @dretrex.

helfrij:
It looks like there is an AMExporterOptionGeneral method `includeAllDataSources()` that is normally set to true, but can be set to false. That's probably the first step to solving this problem, and would likely be done in BioXAS.h.

There are other AMExporterOptionGeneral methods for adding and removing data sources, and setting whether or not they are 'required': `addDataSource(...)`. This may be the second part. I guess I wonder, would these data sources be set in BioXAS.h as well, @dretrex?

dretrex:
They could be.  You could also have it dynamically filled as you add the corresponding data sources in the scan controller since when you call BioXAS::buildExporterOption it returns you a valid option object.  I'd say you probably want to manipulate that but have the `includeAllDataSources()` set to false in the BioXAS.h method.

helfrij:
That makes sense. Thanks, @dretrex.

helfrij:
Done in #2227.

Time Estimate: 1 day.

PR #2281, opened 2016-04-29
Pull Request for Issue2148: generic scan configuration to show detectors already added.

helfrij:
Basic changes:
- Modified the generic step scan configuration to use the existing generic step scan configuration detectors view. This view handles disabling detector options if they are not connected, checking them if they are already added, etc.
- Uncommented calls in BioXASAppController to use existing methods to update generic configurations.
- Modified the detectors view to show the detector descriptions instead of names, if they are available.

During testing, I ran into a couple of unrelated problems that I fixed in this pull request:
- The changes in #2194 meant that only sources that are explicitly added to the exporter option are actually exported. Sources were specified for XAS scans but not for BioXAS generic scans--no data sources were being exported. Modified the BioXAS generic scans to export raw I0, I1, I2 values if they are available.
- Previously, the generic scan controller only added the I0-normalized data to a scan if there was an XRF detector added to the scan. Moved this code out of the XRF detector if-check so that normalized values are provided whenever I0 is provided.

### DIFF
--- a/source/acquaman/AMGenericStepScanConfiguration.cpp
+++ b/source/acquaman/AMGenericStepScanConfiguration.cpp
@@ -283,16 +283,18 @@ void AMGenericStepScanConfiguration::setRegionOfInterestBoundingRange(AMRegionOf
 
 void AMGenericStepScanConfiguration::setI0(const AMDetectorInfo &info)
 {
+	bool validInfo = (info.name() != "Invalid detector");
 	bool nameInDetectorList = false;
 
 	for (int i = 0, size = detectorConfigurations_.count(); i < size && !nameInDetectorList; i++)
 		if (info.name() == detectorConfigurations_.at(i).name())
 			nameInDetectorList = true;
 
-	if (nameInDetectorList){
-
+	if ((validInfo && nameInDetectorList) || !validInfo) {
 		i0_ = info;
 		setModified(true);
+
+		emit i0DetectorChanged();
 	}
 }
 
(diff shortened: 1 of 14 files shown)


helfrij:

source/acquaman/AMGenericStepScanConfiguration.cpp
```diff
@@ -283,16 +283,18 @@ void AMGenericStepScanConfiguration::setRegionOfInterestBoundingRange(AMRegionOf
 
 void AMGenericStepScanConfiguration::setI0(const AMDetectorInfo &info)
 {
+	bool validInfo = (info.name() != "Invalid detector");
```
XuetaoLiu-CLS: I know it's not a big deal. But personally, I would prefer to have a helper function in AMDetectorInfo called `isValidInfo()`. The reason for that is, the caller of AMDetectorInfo won't need to know how the valid/invalid name is definition.

helfrij: That's a good idea. Would it live in the configuration or in AMDetectorInfo?

XuetaoLiu-CLS: It should be in the class of `AMDetectorInfo`, since only `AMDetectorInfo` knows what is valid or invalid.

XuetaoLiu-CLS:
Looks good.

Time Estimate: 10 minutes

source/acquaman/AMGenericStepScanConfiguration.cpp
helfrij: Done.

dretrex:
Looks good.
Time Estimate: 5 minutes

Issue #2253, opened 2016-04-20
Acquaman crash Issue


XuetaoLiu-CLS:
The user reported that SXRMB Acquaman crashed several times in the past weekend. In the crash reports, there are four crashes. But only one of it has some stack information.

/home/acquaman/AcquamanApplicationCrashReports/SXRMB/2016/report_SXRMBAcquaman_OPI1606-602_04162016_144134_23696.txt

XuetaoLiu-CLS:
This looks like a framework issue. I saw similar stack trace on the other beamlines as well.
file:///home/acquaman/AcquamanApplicationCrashReports/VESPERS/2016/report_VESPERSAcquaman_OPI1607-003_04202016_093646_8132.txt

XuetaoLiu-CLS:
Both crashes happened when the exporter is trying to generate the ROI header strings.

:thought_balloon: My guess for the root cause is that when a scan was going, the user/staff changed the ROI of the XRF detector. After the scan is done, we tried to loop through the ROIs for header string. Unfortunately, the ROIs were changed and caused a crash.

I need to verify the hypothesis though :game_die:

dretrex:
That shouldn't be a problem, but it could be possible that something has already deleted the regions of interest associated with the scan configuration (since the regions of interest on the detector are not relevant).

XuetaoLiu-CLS:
@dretrex I just verified this on SXRMB. This crash will happen when you add/remove ROI during a scan. I think this is a good candidate to introduce Michael our system, since we need to touch detector, scan configuration, scans, exports. I will work with Michael tomorrow.

XuetaoLiu-CLS:
Time Estimates: 3 days.

PR #2288, opened 2016-05-03
PullRequest for Issue2253: when adding ROI, XRFDetector create a copy instead of the original instance

XuetaoLiu-CLS:
Reported in Issue #2253, when we start to do a scan, we added the ROIs in the scan configuration to the XRFDetectors in function `buildScanControllerImplementation()`. Unfortunately, we are using the ROI instances of the current scan configuration.

Therefore, if the user removed ROIs from XRFDetectors when a scan is running, we deleted the ROI instance which is associated with the current scan configuration and will cause a crash when we are trying to export the scan.
 
====
Solution:
  - create a copy of ROI when adding the ROI to the XRFDetector
  - small code refactor for `buildScanControllerImplementation()`

### DIFF
--- a/source/acquaman/AMGenericContinuousScanController.cpp
+++ b/source/acquaman/AMGenericContinuousScanController.cpp
@@ -111,7 +111,7 @@ void AMGenericContinuousScanController::buildScanControllerImplementation()
 				scan_->addAnalyzedDataSource(newRegion, true, false);
 			}
 			// This is sufficient to add a region of interest on all detectors as they should by synchronized via AMBeamline::synchronizeXRFDetectors.
-			detector->addRegionOfInterest(region);
+			detector->addRegionOfInterest(region->createCopy());
 		}
 
 		if (configuration_->hasI0()){
--- a/source/acquaman/AMGenericContinuousScanController.cpp
+++ b/source/acquaman/AMGenericContinuousScanController.cpp
@@ -101,6 +101,9 @@ void AMGenericContinuousScanController::buildScanControllerImplementation()
 
 		foreach (AMRegionOfInterest *region, configuration_->regionsOfInterest()){
 
+			// This is sufficient to add a region of interest on all detectors as they should by synchronized via AMBeamline::synchronizeXRFDetectors.
+			detector->addRegionOfInterest(region);
+
 			AMRegionOfInterestAB *regionAB = (AMRegionOfInterestAB *)region->valueSource();
 
 			for(int x = 0, size = spectrumSources.count(); x < size; x++){
@@ -110,8 +113,6 @@ void AMGenericContinuousScanController::buildScanControllerImplementation()
 
 				scan_->addAnalyzedDataSource(newRegion, true, false);
 			}
-			// This is sufficient to add a region of interest on all detectors as they should by synchronized via AMBeamline::synchronizeXRFDetectors.
-			detector->addRegionOfInterest(region->createCopy());
 		}
 
 		if (configuration_->hasI0()){
--- a/source/acquaman/IDEAS/IDEAS2DScanActionController.cpp
+++ b/source/acquaman/IDEAS/IDEAS2DScanActionController.cpp
@@ -106,14 +106,7 @@ void IDEAS2DScanActionController::buildScanControllerImplementation()
     transmissionAB->setNormalizationName("Sample");
     scan_->addAnalyzedDataSource(transmissionAB);
 
-    AMXRFDetector *detector = 0;
-
-	if (configuration_->fluorescenceDetector().testFlag(IDEAS::Ketek))
-		detector = qobject_cast<AMXRFDetector *>(AMBeamline::bl()->exposedDetectorByName("KETEK"));
-
-	else if (configuration_->fluorescenceDetector().testFlag(IDEAS::Ge13Element) && IDEASBeamline::ideas()->ge13Element()->isConnected())
-		detector = qobject_cast<AMXRFDetector *>(AMBeamline::bl()->exposedDetectorByName("13-el Ge"));
-
+	AMXRFDetector *detector = IDEASBeamline::ideas()->xrfDetector(configuration_->fluorescenceDetector());
 	if (detector){
 
 		detector->removeAllRegionsOfInterest();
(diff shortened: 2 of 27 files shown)


source/acquaman/AMGenericContinuousScanController.cpp
```diff
@@ -101,6 +101,9 @@ void AMGenericContinuousScanController::buildScanControllerImplementation()
 
 		foreach (AMRegionOfInterest *region, configuration_->regionsOfInterest()){
 
+			// This is sufficient to add a region of interest on all detectors as they should by synchronized via AMBeamline::synchronizeXRFDetectors.
+			detector->addRegionOfInterest(region);
```
helfrij: If the detector ROI information isn't changed during a scan, the detector ROIs should be the same as the configuration ROIs right? Is it possible to 'double add' regions of interest here?

source/acquaman/VESPERS/VESPERSSpatialLineScanActionController.cpp
```diff
@@ -264,21 +264,12 @@ VESPERSSpatialLineScanActionController::VESPERSSpatialLineScanActionController(V
...
 
-	else if (xrfDetector.testFlag(VESPERS::Ge13Element))
-		detector = qobject_cast<AMXRFDetector *>(VESPERSBeamline::vespers()->exposedDetectorByName("Ge13El"));
+	AMXRFDetector *xrfDetector = VESPERSBeamline::vespers()->xrfDetector(xrfDetectorConfig);
+	if (xrfDetector){
 
-	if (detector){
-
-		detector->removeAllRegionsOfInterest();
+		xrfDetector->removeAllRegionsOfInterest();
```
helfrij: I think this is the only call to removeAllRegionsOfInterest() that I've seen in these changes so far--is this intentional?

source/beamline/SXRMB/SXRMBBeamline.cpp
```diff
@@ -621,7 +634,7 @@ void SXRMBBeamline::setupComponents()
 	jjSlits_->setHorizontalSlitClosedGapValue(0);
 
 	//energy_ = new AMPVwStatusControl("Energy", "BL1606-B1-1:Energy:fbk", "BL1606-B1-1:Energy", "BL1606-B1-1:Energy:status", QString(), this, 0.1, 2.0, new AMControlStatusCheckerCLSMAXv());
-	energy_ = new AMPVwStatusControl("Energy", "BL1606-B1-1:AddOns:Energy:fbk", "BL1606-B1-1:AddOns:Energy", "BL1606-B1-1:AddOns:Energy:status", "BL1606-B1-1:AddOns:Energy:stop", this, 0.001, 2.0, new CLSMAXvControlStatusChecker());
+	energy_ = new AMPVwStatusControl("Energy", "BL1606-B1-1:AddOns:Energy:fbk", "BL1606-B1-1:AddOns:Energy", "BL1606-B1-1:AddOns:Energy:status", "BL1606-B1-1:AddOns:Energy:stop", this, 0.05, 2.0, new CLSMAXvControlStatusChecker());
```
helfrij: Was the tolerance too small before here?

helfrij:
Looks good, @XuetaoLiu-CLS!

Time Estimate: 10 minutes.

source/acquaman/AMGenericContinuousScanController.cpp
dretrex: @helfrij, there is a `detector->removeAllRegionsOfInterest()` on line 87 of this file so there shouldn't be an issue with double adding.  I believe @XuetaoLiu-CLS just moved this method.

source/acquaman/IDEAS/IDEAS2DScanActionController.cpp
```diff
@@ -106,14 +106,7 @@ void IDEAS2DScanActionController::buildScanControllerImplementation()
...
 
-    AMXRFDetector *detector = 0;
-
-	if (configuration_->fluorescenceDetector().testFlag(IDEAS::Ketek))
-		detector = qobject_cast<AMXRFDetector *>(AMBeamline::bl()->exposedDetectorByName("KETEK"));
-
-	else if (configuration_->fluorescenceDetector().testFlag(IDEAS::Ge13Element) && IDEASBeamline::ideas()->ge13Element()->isConnected())
-		detector = qobject_cast<AMXRFDetector *>(AMBeamline::bl()->exposedDetectorByName("13-el Ge"));
-
+	AMXRFDetector *detector = IDEASBeamline::ideas()->xrfDetector(configuration_->fluorescenceDetector());
```
dretrex: This only works while the Ketek and the 13 element are never used at the same time.  Since, practically, this is the case we are okay, but I wanted to bring it to your attention.

dretrex: After looking into it more, this is okay.  I was mixing up the check of whether or not there was an XRF detector at all vs. setting up the data sources properly if there are multiple XRF detectors.

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -81,7 +81,7 @@ void BioXASAppController::onUserConfigurationLoadedFromDb()
 
 				foreach (AMRegionOfInterest *region, userConfiguration_->regionsOfInterest()){
 					if (!containsRegionOfInterest(geDetector->regionsOfInterest(), region)) {
-						geDetector->addRegionOfInterest(region->createCopy());
```
dretrex: Are you sure about not having a copy here

source/beamline/AMXRFDetector.cpp
```diff
@@ -333,10 +333,13 @@ void AMXRFDetector::addRegionOfInterest(const AMEmissionLine &emissionLine)
 	AMRegionOfInterest *region = new AMRegionOfInterest(emissionLine.name(), energy, AMRange(energy*(1-halfWidth), energy*(1+halfWidth)), this);
 
 	addRegionOfInterest(region);
+	region->deleteLater();
 }
 
-void AMXRFDetector::addRegionOfInterest(AMRegionOfInterest *newRegionOfInterest)
+void AMXRFDetector::addRegionOfInterest(AMRegionOfInterest *regionOfInterest)
 {
+	AMRegionOfInterest *newRegionOfInterest = regionOfInterest->createCopy();
```
dretrex: By creating a copy, does this affect changing the bounding range synchronization between all the detectors and scan configurations?

dretrex:
Some comments.
Time Estimate: 10 minutes

source/acquaman/AMGenericContinuousScanController.cpp
XuetaoLiu-CLS: @helfrij @dretrex Yes, I just moved this LOC from Ln113 to here to make it clear that we added the ROI first

source/acquaman/VESPERS/VESPERSSpatialLineScanActionController.cpp
XuetaoLiu-CLS: We called  `removeAllRegionsOfInterest()` in all `buildScanControllerImplementation()`. The changes here is just to rename the local variable `detector` to `xrfDetector` to make it more specific

source/beamline/SXRMB/SXRMBBeamline.cpp
XuetaoLiu-CLS: Yes, it is. I should changed it in the other issues. I made the changes again to make sure I did that change

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: the default behavior of `addRegionOfIntere()` is to create a copy or the `region`. Therefore, we won't have to force a copy here. Otherwise, it will be a potential memory leak.

source/beamline/AMXRFDetector.cpp
XuetaoLiu-CLS: No, it won't.  After we added the ROI to the XRFDetector, we emit a signal to the scan configurations, which will re-listen to the changes from this ROI.

PR #2290, opened 2016-05-03
Pull Request for Issue2287: Add control information to XAS scan header

helfrij:
Modified BioXASBeamline: added virtual getters for the I0, I1, I2 Keithley amplifiers, and new getter for the list of control infos that would be useful to know about for scans (storage ring current, settling time, I0 gain, I1, gain, I2 gain). Removed the scanNotes() getter from BioXASBeamline--this functionality is now with the new control info getter. Modified BioXAS.h so that the exporter option now includes relevant control info information in the scan header.

### DIFF
--- a/source/beamline/BioXAS/BioXASBeamline.h
+++ b/source/beamline/BioXAS/BioXASBeamline.h
@@ -90,9 +90,6 @@ public:
 	/// Creates and returna an action that cleans up the beamline after a scan.
 	virtual AMAction3* createScanCleanupAction(AMGenericStepScanConfiguration *configuration);
 
-	/// Returns a string representation of the beamline settings to include in the scan notes.
-	virtual QString scanNotes() const;
-
 	/// Returns the beam status.
 	virtual BioXASBeamStatus* beamStatus() const { return beamStatus_; }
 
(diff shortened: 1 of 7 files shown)


helfrij:

source/beamline/BioXAS/BioXASBeamline.h
```diff
@@ -165,10 +162,16 @@ class BioXASBeamline : public CLSBeamline
 
 	/// Returns the scaler.
 	virtual CLSSIS3820Scaler* scaler() const { return 0; }
+	/// Returns the I0 Keithley amplifier.
+	virtual CLSKeithley428* i0Keithley() const { return 0; }
```
XuetaoLiu-CLS: I have a question here. Does the subclass must re-implement these functions? 
If yes, it will be better to define those functions as pure virtual functions. In that way, the compiler will tell us if we forget to implement those functions in the subclass.

XuetaoLiu-CLS:
Time Estimate: 5 minutes

source/beamline/BioXAS/BioXASBeamline.h
helfrij: I think it's better to keep them as they are for now. Imaging will likely follow the same paradigm (i0, i1, i2) and is a subclass of BioXASBeamline, but doesn't have Keithley objects to return yet.

dretrex:
Looks good.
Time Estimate: 1 minute

PR #2297, opened 2016-05-06
Pull request for Issue2069: Add scan serial number to AMScanThumbnailGridView scan items.

mButlerDM:
Added method to generate a QRect for the serial number. Adjusted name rectangle so it was shifted to the right so that the serial could paint in front of it. Added calls to paint serial along with other elements.

Since the serial QModelIndex value was used to paint the background previously, new methods had to be added to generate a rectangle for the scan item background and paint it, apart from the functions dependent on the model index values.

### DIFF
--- a/source/ui/dataman/AMScanThumbnailGridView.cpp
+++ b/source/ui/dataman/AMScanThumbnailGridView.cpp
@@ -4,423 +4,445 @@
 #include "ui/dataman/AMScanThumbnailGridInputManager.h"
 
 AMScanThumbnailGridView::AMScanThumbnailGridView(QWidget *parent) :
-	QAbstractItemView(parent)
+    QAbstractItemView(parent)
 {
-	geometryManager_ = new AMScanThumbnailGridGeometryManager(sizeHint().width());
-	// Wire up the input manager to this view
-	inputManager_ = new AMScanThumbnailGridInputManager();
-	connect(inputManager_, SIGNAL(itemSelected(int,QItemSelectionModel::SelectionFlags)), this, SLOT(onItemSelected(int,QItemSelectionModel::SelectionFlags)));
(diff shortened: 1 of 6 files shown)


mButlerDM:
@iainworkman Since you helped with this would you mind taking a look when you have a chance?

Thanks

source/ui/dataman/AMScanThumbnailGridView.cpp
```diff
@@ -4,415 +4,443 @@
 #include "ui/dataman/AMScanThumbnailGridInputManager.h"
 
 AMScanThumbnailGridView::AMScanThumbnailGridView(QWidget *parent) :
-	QAbstractItemView(parent)
```
iainworkman: Either your spacing is messed up now, or mine was messed up when I first wrote this code. My guess is the latter. You don't need to do anything with this code, but do remind me to check your settings.

source/ui/dataman/AMScanThumbnailGridView.cpp
```diff
@@ -4,415 +4,443 @@
...
-		case 2:
-			return geometryManager_->thumbnailTitleRectangle(rowContentRect);
-
-		default:
-			return QRect();
-		}
-	}
+    QRect rowContentRect;
+    /*
+    if(!index.isValid()){
```
iainworkman: Can this commented code be removed?

mButlerDM: @iainworkman Yes this should be removed, it was a left over fragment from that first solution attempt. I removed it, tested, and pushed the update.

iainworkman:
Sorry, forgot to leave the reviewed comment.
Looks good to me.
Time Estimate: 5 minutes.

dretrex:
Looks good.
Time Estimate: 3 minutes

PR #2302, opened 2016-05-09
Pull Request for Issue2299: Adding option to display dead time buttons with count rate.

helfrij:
Added getter to AMXRFDetector for the acquire time control. For the Xspress3 detectors on BioXAS, this control is the Zebra pulse width (in seconds) pseudomotor control. Modified AMDeadTimeButton to have a couple of different CountMode options: can view the counts information as a percentage, as flat counts, or as a count rate provided the button has the appropriate resources. Modified AMXRFDetailedDetectorView to use the new AMDeadTimeButton display changes.

### DIFF
--- a/source/ui/beamline/AMDeadTimeButton.cpp
+++ b/source/ui/beamline/AMDeadTimeButton.cpp
@@ -166,16 +166,16 @@ void AMDeadTimeButton::updateColorState()
 
 void AMDeadTimeButton::updateToolTip()
 {
-	if (countsMode_ == None)
-		setToolTip("");
-	else if (countsMode_ == Percent && canDisplayPercentage())
-		setToolTip(QString("%1%").arg(getPercent(), 0, 'f', 0));
-	else if (countsMode_ == Counts && canDisplayCounts())
-		setToolTip(QString("%1 counts").arg(getCounts()));
-	else if (countsMode_ == CountRate && canDisplayCountRate())
-		setToolTip(QString("%1 counts/s").arg(getCountRate()));
+	QString toolTip = "";
 
-	update();
+	if (countsMode_ == Percent && canDisplayPercentage())
+		toolTip = QString("%1%").arg(getPercent(), 0, 'f', 0);
+	else if (countsMode_ == Counts && canDisplayCounts())
+		toolTip = QString("%1 counts").arg(getCounts());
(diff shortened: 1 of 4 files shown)


helfrij:

source/ui/beamline/AMDeadTimeButton.cpp
```diff
@@ -142,3 +187,48 @@ bool AMDeadTimeButton::hasICRDataSource() const
...
+
+	return result;
+}
+
+double AMDeadTimeButton::getCountRate() const
+{
+	double result = 0;
+
+	if (canDisplayCountRate())
+		result = double(inputCountSource_->value(AMnDIndex())) / acquireTimeControl_->value();
```
XuetaoLiu-CLS: In real environment, the `acquireTimeControl_->value()` will never be zero. But for safety, should we add a protection here?

source/ui/beamline/AMDeadTimeButton.cpp
```diff
@@ -142,3 +187,48 @@ bool AMDeadTimeButton::hasICRDataSource() const
...
+{
+	return !(acquireTimeControl_ == 0);
+}
+
+double AMDeadTimeButton::getPercent() const
+{
+	double result = 0;
+
+	if (canDisplayPercentage())
+		result = 100*(1 - double(outputCountSource_->value(AMnDIndex()))/double(inputCountSource_->value(AMnDIndex())));
```
XuetaoLiu-CLS: Same here. I think we should always be cautious on div operation.

XuetaoLiu-CLS:
Looks good other than a question on whether we should have protection on the division operation

Time Estimate: 5 minutes

source/ui/beamline/AMDeadTimeButton.cpp
helfrij: `canDisplayCountRate()` checks for whether or not the acquireTimeControl_ is 0, but this check could definitely be clearer.

source/ui/beamline/AMDeadTimeButton.cpp
helfrij: Same thing here. `canDisplayPercentage()` checks whether the outputCountSource_ and inputCountSource_ are 0. I'll modify to make this explicit/more clear.

dretrex:
Looks good.
Time Estimate: 3 minutes

PR #2308, opened 2016-05-11
Pull Request for Issue2307: Solving problem with acquisition state.

helfrij:
Added getter to AMXRFDetector that returns the number of enabled elements. Added call to this getter to AMXspress3XRFDetector, replacing the old call to elements() when updating the acquisition state. Waiting on all elements to update was causing the acquisition to hang, because the wait is for enabled and disabled elements. Should only consider enabled elements.


helfrij:

iainworkman:
Looks good to me.
Time Estimate: 1 minute

dretrex:
Looks good.
Time Estimate: 1 minute

Issue #1445, opened 2015-06-15
Add ability to send motors to a given position by right clicking on scan data.


helfrij:
On VESPERS, a user can move motors to a given position by right clicking on a position in a scan view. The BioXAS team would like to see this in the BioXAS application as well.

helfrij:
Will likely need to finish #1444 before this one can be tackled in earnest. (so sayeth @dretrex )

dretrex:
This is possible to do now based on all the work @helfrij has done.  Maybe it can even be done in a general way that can be used across beamlines?

XuetaoLiu-CLS:
Time Estimate: 4 hours

XuetaoLiu-CLS:
It can't be generalized at this moment. Maybe in the future...

PR #2312, opened 2016-05-13
PullRequest for Issue1445: When openning a 2D generic scan, we will add a popup menu for the scan editor

XuetaoLiu-CLS:
#1445 

When openning a 2D generic scan, we will add a popup menu for the scaneditor editor, so that the user can move the controls to the corresponding
postition
- add axisControlInfoAt() helper function for AMScanConfiguration
- add refine some debugging log for the synchronization feature
- add the feature to pop up a menu to move the controls for 2D generic
scans


XuetaoLiu-CLS:
Hi @helfrij can you review? Thx

source/acquaman/AMContinuousScanActionController.cpp
```diff
@@ -60,8 +60,8 @@ void AMContinuousScanActionController::buildScanController()
 	for (int i = 0, axisCount = scan_->rawData()->scanAxesCount(); i < axisCount; i++){
 
 		int actualAxis = axisOrderMap_.value(scan_->rawData()->scanAxisAt(i).name);
-		scanAssembler_->insertAxis(i, AMBeamline::bl()->exposedControlByInfo(continuousConfiguration_->axisControlInfos().at(actualAxis)), continuousConfiguration_->scanAxisAt(actualAxis));
-		qDebug() << "Adding axis " << continuousConfiguration_->axisControlInfos().at(i).name();
+		scanAssembler_->insertAxis(i, AMBeamline::bl()->exposedControlByInfo(continuousConfiguration_->axisControlInfoAt(actualAxis)), continuousConfiguration_->scanAxisAt(actualAxis));
+		qDebug() << "Adding axis " << continuousConfiguration_->axisControlInfoAt(i).name();
```
helfrij: Would you like to keep these QDebugs?

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -251,12 +343,6 @@ void BioXASAppController::updateGenericScanConfigurationDetectors()
 	}
 }
 
-void BioXASAppController::initializeBeamline()
```
helfrij: Was this method moved, or replaced with something else?

source/util/AMRecursiveDirectoryCompare.cpp
```diff
@@ -215,17 +215,22 @@ bool AMRecursiveDirectoryCompare::compareOneLevel(const QString &path1, const QS
 	}
 	onlySide1SubDirectories = side1SubDirectories;
 
+
+	if (unknownFiles.count() > 0)
+		qDebug() << QString("Found (%1) files for which the more up-to-date version cannot be determined, leading with %2.").arg(unknownFiles.count()).arg(unknownFiles.at(0));
```
helfrij: QDebug?

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -67,6 +67,98 @@ void BioXASAppController::shutdown()
...
+	popup.addAction("Go to immediately");
+
+	temp = popup.exec(pos);
+	if (temp){
+		if (temp->text() == "Go to immediately")
+			moveImmediately(editor);
+	}
+}
+
+void BioXASAppController::moveImmediately(const AMGenericScanEditor *editor)
```
helfrij: Would it make sense to break this method down further into separate components: one for producing the axis control move action, and one for running it?

source/util/AMRecursiveDirectoryCompare.cpp
```diff
@@ -215,17 +215,22 @@ bool AMRecursiveDirectoryCompare::compareOneLevel(const QString &path1, const QS
...
 	for (int x = 0, size = unknownFiles.count(); x < size; x++) {
-		qDebug() << "Found files for which the more up-to-date version cannot be determined";
 		unknownFiles_.append(QDir::cleanPath(QString(".../%1").arg(unknownFiles.at(x))));
 	}
 
-	for(int x = 0, size = newerSide1Files.count(); x < size; x++){
+	if (newerSide1Files.count() > 0)
```
helfrij: Why are these changes here? Was there a problem with testing?

helfrij:
Looks good, @XuetaoLiu-CLS. Would you like to test these changes on the beamline?

Time Estimate: 10 minutes.

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: I make it a pure virtual function, which forces the subclass to implement it.

It is a misleading to have an implementation for BioXASBeamline, since in fact, we must instance the BioXASSideBeamline or BioXASMainBeamline.

A side note about removing this is that this implementation caused two hours yesterday to figure out why my codes doesn't work :-1:

source/util/AMRecursiveDirectoryCompare.cpp
XuetaoLiu-CLS: I'm not sure whether we want to keep it or not. It's legacy codes though

source/util/AMRecursiveDirectoryCompare.cpp
XuetaoLiu-CLS: Look at the original codes, the qDebug() is sitting inside the loop. That is, we will print out the same message `count()` times, which only gives us a single information.

The new changes will only qDebug() a single line.

XuetaoLiu-CLS: it is not a problem, just annoying.

XuetaoLiu-CLS:
@helfrij Somebody is using the side beamline the whole week. Therefore, I copied a group user scans to my local environment for the test. But I didn't test the real move though (I commented off `moveImmediatelyAction_->start();`).

If we trust the ActionList, then the codes should work :+1:

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: It is good to break it down for sure and I was planning to do it at the beginning.

As mentioned in the issue, we "think" it might be nicer to have a generalized solution for this so that all the beamlines can benefit from this. I found it is hard to do it since the feature relies on the "type" of the scan configuration. (Ln112). I think it might be better to do the code refactor later.

source/util/AMRecursiveDirectoryCompare.cpp
helfrij: Ah, okay.

source/application/BioXAS/BioXASAppController.cpp
helfrij: Okay, sounds good.

source/util/AMRecursiveDirectoryCompare.cpp
helfrij: Ah, okay.

source/application/BioXAS/BioXASAppController.h
```diff
@@ -76,6 +76,8 @@
 
 #include "ui/util/AMChooseDataFolderDialog.h"
 
+#define BIOXAS_APPCONTROLLER_INVALID_AXIS 201001001
```
dretrex: I like this error code for some reason...

dretrex:
Looks good.
Time Estimate: 5 minutes

PR #2314, opened 2016-05-16
PullRequest for Issue1628

XuetaoLiu-CLS:
- add displayMessageBox function for AMErrorMonitor so that theAMActionRunner3.cpp doesn't have to worry about UI header files imports
- add removeActionFromQueue() function to simply remove an action from
the queue


XuetaoLiu-CLS:
close this one since there is a new branch created

PR #2320, opened 2016-05-16
Pull Request for Issue2316: solution to CLSExclusiveStates move bug.

helfrij:
Changes:
- [x] Added AMErrorMon alert to AMControlWaitAction, so that users are notified if a wait action fails.
- [x] Modified `AMExclusiveStatesControl::setStatusControl(...)` so that it explicitly reimplements `AMSingleEnumeratedControl::setStatusControl(...)`.
- [x] Added wait action to the move action generated by `CLSExclusiveStatesControl::createMoveAction(...)` that waits for the status control to achieve the setpoint for the desired index. This was missing before and is the source of the original bug.

The problem with the shutters control appears to be solved with these changes. Waiting on being able to open the front-end shutters to test the shutters in the context of scan initialization.


helfrij:

iainworkman:
Looks good to me.
Time Estimate: 5 minutes

dretrex:
Looks good.
Time Estimate: 2 minutes

PR #2315, opened 2016-05-16
Pull Request for Issue1451: updating mono region control.

helfrij:
- Updated the SSRL mono classes: variable names, macro names, using AMActionSupport to create move actions. Replaced the CLSMAXvMotors in the Side and Main mono classes used in the crystal change with BioXASMAXvMotors.
- Modified the 'move to limit' action to be able to handle when the motor reports a move done before the limit control reports that the motor is at a limit.
- Modified the CLSControlEditor to include a flag for whether or not the editor is the one that initiated a control's current move.

PR #2323, opened 2016-05-17
Pull Request for Issue2319: adding 'move to limit' capability to select motors on BioXAS

helfrij:
Changes:
- Created subclass of CLSMAXvMotor, BioXASMAXvMotor. This class provides a basic interface for setting limit setpoints and moving to them.
- Created new action and info, BioXASMAXvMotorMoveToLimitAction/Info. These classes will call `BioXASMAXvMotor::moveToLimit(...)` and check whether or not the move to the limit has succeeded.

Tested with a few motors on BioXAS (mono mask upper and lower blades, mono paddle) and seems to work as expected, both when the motor starts at the limit and when it needs to travel to the appropriate limit.


helfrij:

source/actions3/BioXAS/BioXASMAXvMotorMoveToLimitAction.cpp
```diff
@@ -0,0 +1,129 @@
...
+	if (atLimit()) {
+		setProgress(100, 100);
+		setSucceeded();
+
+	} else {
+		setFailed(QString("Failed to move motor '%1' to '%2' limit. The move failed without reaching the limit.").arg(control_->name()).arg(moveToLimitInfo()->limitSetpointToString(moveToLimitInfo()->limitSetpoint())));
+	}
+}
+
+void BioXASMAXvMotorMoveToLimitAction::onMotorMoveSucceeded()
```
iainworkman: If this and the above onMotorMoveFailed() contain the exact same code could you not hook both signals up to one slot?
onMotorMoveFinished()?

iainworkman:
@helfrij Just wondering what the use-case for this is? Are there many cases where BIOXAS wishes to drive motors to their limits?

source/actions3/BioXAS/BioXASMAXvMotorMoveToLimitAction.cpp
helfrij: Yes, that's a good idea.

helfrij:
This functionality will be immediately useful in the updated crystal change procedure being developed in #2315. Several parts of the procedure involve moving motors from CW limits to CCW, and reverse. Previously this was accomplished by approximating the limit position and trying to use `move(...)` to get there, but moves done this way fail if they can't reach the destination setpoint. That's part of the rationale behind making these changes anyways.

iainworkman:
The only other then might be that the use of CCW and CW is quite specific to this case for the enums, and that maybe HIGH_LIMIT and LOW_LIMIT might be better?

iainworkman:
Otherwise looks good. 
Time Estimate: 10 minutes

helfrij:
The LimitCW and LimitCCW enums are used here because they already existed in CLSMAXvMotor. We could switch to HIGH_LIMIT and LOW_LIMIT, but then we'd be using different terminology (vs CW/CCW) than the MAXvMotors use for these values. Or am I misunderstanding something?

iainworkman:
Oh, I see. That seems the most reasonable thing then.

source/actions3/BioXAS/BioXASMAXvMotorMoveToLimitActionInfo.cpp
```diff
@@ -0,0 +1,67 @@
+#include "BioXASMAXvMotorMoveToLimitActionInfo.h"
+
+BioXASMAXvMotorMoveToLimitActionInfo::BioXASMAXvMotorMoveToLimitActionInfo(const AMControlInfo &motor, CLSMAXvMotor::Limit limitSetpoint, QObject *parent) :
+	AMActionInfo3(QString(), QString(), QString(), parent)
```
dretrex: Just to be pedantic, do you want to have real values for these names and descriptions?

helfrij: Both of those field are updated in `updateDescriptionText()`, called below. Is that sufficient?

dretrex: Totally.

PR #2332, opened 2016-05-18
Pull Request for Issue2329: 2D maps last line overwriting first line

dretrex:
Fixed the overwriting data issue.  AMScanAxisRegion was truncating values by casting them to integers rather than rounding the result.


helfrij:
Looks good.

Time Estimate: 2 minutes.

PR #2344, opened 2016-05-20
Pull Request for Issue2343: Modifying limits functionality in BioXASControlEditor

helfrij:
Changes:
- Created new class BioXASValueSetpointEditor that's a combo spinbox/combobox as appropriate, depending on the InputType. The class also has an InputStatus property that's used to style the widget according to the whether or not the input is okay.
- Created new stylesheet for the setpoint editor that gives it a red border for bad input values.
- Modified AMAppController to apply stylesheets while setting up the user interface, instead of in the constructor. This allows for app controller subclasses to reimplement applyStylesheets() properly.
- Created new class BioXASValueSetpointEditorDialog, a dialog with the new setpoint editor and OK/Cancel buttons.
- Modified BioXASValueEditor to use the new dialog instead of the old default 'QInputDialog' option.
- Modified BioXASSSRLMonochromatorRegionControlEditor to inherit from BioXASControlEditor instead of AMExtendedControlEditor.

### DIFF
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -251,10 +251,12 @@ QString BioXASAppController::getStylesheet() const
 
 	// BioXASValueSetpointEditor.
 
-	QFile qss3(":/BioXAS/BioXASValueSetpointEditor.qss");
+	QFile qss(":/BioXAS/BioXASValueSetpointEditor.qss");
 
-	if (qss3.open(QFile::ReadOnly))
-		stylesheet.append(QString("\n\n%1").arg(QLatin1String(qss3.readAll())));
+	if (qss.open(QFile::ReadOnly))
+		stylesheet.append(QString("\n\n%1").arg(QLatin1String(qss.readAll())));
+
+	qss.close();
 
 	return stylesheet;
 }
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -67,45 +67,6 @@ void BioXASAppController::shutdown()
 	CLSAppController::shutdown();
 }
 
-void BioXASAppController::applyStylesheets()
-{
-	// Go through list of stylesheets to be applied,
-	// composing a 'master' sheet.
-
-	QString stylesheet;
-
-	// AMToolButton
-
-	QFile qss1(":/AMToolButton.qss");
-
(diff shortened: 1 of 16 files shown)


helfrij:

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -67,6 +67,45 @@ void BioXASAppController::shutdown()
...
+	QFile qss1(":/AMToolButton.qss");
+
+	if (qss1.open(QFile::ReadOnly))
+		stylesheet.append(QString("\n\n%1").arg(QLatin1String(qss1.readAll())));
+
+	qss1.close();
+
+	// AMDeadTimeButton
+
+	QFile qss2(":/AMDeadTimeButton.qss");
```
XuetaoLiu-CLS: shouldn't we put those stylesheet into super class?

source/ui/BioXAS/BioXASPersistentView.cpp
```diff
@@ -28,6 +28,7 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 #include "ui/BioXAS/BioXASCryostatView.h"
 #include "ui/BioXAS/BioXASSIS3820ScalerChannelsView.h"
 #include "ui/AMToolButton.h"
+#include "ui/BioXAS/BioXASValueSetpointEditor.h"
```
XuetaoLiu-CLS: is this used anywhere in the BioXASPersistentView?

XuetaoLiu-CLS:
Looks good. other than a question.

Time Estimate: 10 minutes

source/application/BioXAS/BioXASAppController.cpp
helfrij: Unfortunately, the stylesheet needs to be applied all at once at the AppController level; subsequent calls to `qApp->setStylesheet(...)` overwrite previous entries. We could do something like a general `AMDatamanAppController::getStylesheet(...)` that could be reimplemented for each application as desired--that's an idea.

source/ui/BioXAS/BioXASPersistentView.cpp
helfrij: It shouldn't be. Probably just leftover from testing. 😃

dretrex:
This looks really good!
Time Estimate: 10 minutes

PR #2347, opened 2016-05-20
Pull Request for Issue1399: Scanning backwards in 1D

dretrex:
Implemented scanning backwards in one dimension.  This _should_ have no effect on anyone since no one scans backwards.

### DIFF
--- a/source/acquaman/AMGenericScanActionControllerAssembler.cpp
+++ b/source/acquaman/AMGenericScanActionControllerAssembler.cpp
@@ -139,8 +139,13 @@ AMAction3* AMGenericScanActionControllerAssembler::generateActionTreeForStepAxis
 
 	AMListAction3 *allRegionsList = new AMListAction3(new AMListActionInfo3(QString("%1 Regions for %2 Axis").arg(stepScanAxis->regionCount()).arg(stepScanAxis->name()), QString("%1 Regions for %2 Axis").arg(stepScanAxis->regionCount()).arg(stepScanAxis->name())), AMListAction3::Sequential);
 
-	for(int x = 0; x < stepScanAxis->regionCount(); x++)
-		allRegionsList->addSubAction(generateActionTreeForStepAxisRegion(axisControl, stepScanAxis->regionAt(x), (x == stepScanAxis->regionCount()-1) ));
+	if (direction_ == AMScanConfiguration::Increase)
+		for(int x = 0; x < stepScanAxis->regionCount(); x++)
+			allRegionsList->addSubAction(generateActionTreeForStepAxisRegion(axisControl, stepScanAxis->regionAt(x), (x == stepScanAxis->regionCount()-1) ));
+
+	else
+		for (int x = stepScanAxis->regionCount()-1; x >= 0; x--)
+			allRegionsList->addSubAction(generateActionTreeForStepAxisRegion(axisControl, stepScanAxis->regionAt(x), (x == 0) ));
 
 	axisActions->addSubAction(allRegionsList);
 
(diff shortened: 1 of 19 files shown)


source/acquaman/AMGenericScanActionControllerAssembler.cpp
```diff
@@ -124,16 +124,28 @@ AMAction3* AMGenericScanActionControllerAssembler::generateActionTreeForStepAxis
 		// generate axis initialization list
 		AMListAction3 *initializationActions = new AMListAction3(new AMListActionInfo3(QString("Initializing %1").arg(axisControl->name()), QString("Initializing Axis with Control %1").arg(axisControl->name())), AMListAction3::Sequential);
 
-		AMAction3 *initializeControlPosition = AMActionSupport::buildControlMoveAction(axisControl, stepScanAxis->axisStart());
+		AMAction3 *initializeControlPosition = 0;
+
+		if (direction_ == AMScanConfiguration::Increase)
+			initializeControlPosition = AMActionSupport::buildControlMoveAction(axisControl, stepScanAxis->axisStart());
```
helfrij: Just curious, because I'm not sure exactly how directions work for scan axes. Does the `axisStart()` method take into account the desired scan direction, or is it the initial position of the first region?

source/acquaman/AMGenericScanActionControllerAssembler.cpp
```diff
@@ -156,7 +168,13 @@ AMAction3* AMGenericScanActionControllerAssembler::generateActionTreeForStepAxis
 
 	if (axisControl){
 
-		AMAction3 *regionStart = AMActionSupport::buildControlMoveAction(axisControl, stepScanAxisRegion->regionStart());
+		AMAction3 *regionStart = 0;
+
+		if (direction_ == AMScanConfiguration::Increase)
```
helfrij: I wonder, is it possible for the `direction_` to be out of sync with the regions? I mean, could somebody manually enter regions that go backwards and `direction_ == AMScanConfiguration::Increase`?

source/acquaman/AMStepScanActionController.cpp
```diff
@@ -169,7 +169,7 @@ void AMStepScanActionController::buildScanController()
 			scan_->addRawDataSource(new AMRawDataSource(scan_->rawData(), scan_->rawData()->measurementCount()-1), oneDetector->isVisible(), oneDetector->hiddenFromUsers());
 	}
 
-	if (scan_->rawData()->scanAxesCount() >= 2)
+	if (scan_->rawData()->scanAxesCount() >= 2 || (scan_->rawData()->scanAxesCount() == 1 && stepConfiguration_->direction() == AMScanConfiguration::Decrease))
```
helfrij: I'm guessing the case where `stepConfiguration_->direction() == AMScanConfiguration::Increase` is handled elsewhere?

helfrij:
Looks good @dretrex. I had a couple of questions, mostly because I'm not super familiar with the scan assembler and controller deeper down. You mentioned that no one wants to scan backwards yet, but I wonder if there is a way to test the code anyways, maybe with a 1D generic step scan?

Time Estimate: 15 minutes.

source/acquaman/AMGenericScanActionControllerAssembler.cpp
dretrex: No, `axisStart` and `axisEnd` should always (ideally) be the way they always have (i.e.: -10 to 10) and it's up to the direction to build the actions such that the direction constraint is satisfied.  I'm sure there is probably a better way but this is what we'll use for version 1.

source/acquaman/AMGenericScanActionControllerAssembler.cpp
dretrex: Yes, it is possible.  But it's always been possible (for instance you could put in 400 eV to 300 eV and just have the worst scan ever for the longest time.

source/acquaman/AMStepScanActionController.cpp
dretrex: No, increasing scans don't prefill their data.  Scanning backwards needs to have it's data prefilled due to a limitation of the CDF data store.

PR #2348, opened 2016-05-24
Pull Request for Issue2341: Create New Data Folder Dialog Gets in Bad State When Cancelled

### LINKED ISSUE #2341: Create New Data Folder Dialog Gets in Bad State When Cancelled
When setting up a new data folder, if the user inputs the new data folder then clicks cancel the semi created data folder gets in a bad state which means it cannot be used without being manually removed first.

mButlerDM:
First run sets the userDataFolder to the remote location on first run, and the local location for subsequent runs. The local location is not made since the wizard did not run with the first settings and the wizard tries to run at a non-existent location on the second. I adjusted the conditional logic so that: It checks if it is the first run, if not, it checks that the local location was made successfully, otherwise treat it like a first run.

I fixed another issue I noticed in testing where if a data folder is made with only the folder name in the dialog, then the user switches to a full path project, or specifies a previous data folder by the full path, the remoteDataFolder in the .ini is not updated and it attempts to sync the selected data folder with the previous remote data folder. 

Edited logic so that if the data folder name does not exist in the remote data path, delete the remoteDataFolder config entry. This avoids attempts to sync to the wrong folder. For full path data folders, remote is set by the network synchronizer. In the worst case, remoteDataFolder is deleted in error, and the user will be prompted if they want to use local storage with their project and it will be setup to the right location by the synchronizer again.

Resolves #2341

### DIFF
--- a/source/ui/util/AMChooseDataFolderDialog.cpp
+++ b/source/ui/util/AMChooseDataFolderDialog.cpp
@@ -16,222 +16,224 @@
 
 bool AMChooseDataFolderDialog::getDataFolder(const QString &localRootDirectory, const QString &remoteRootDirectory, const QString &dataDirectory, const QStringList &extraDataDirectory, QWidget *parent)
 {
-	AMUserSettings::load();
+        AMUserSettings::load();
 
-	AMChooseDataFolderDialog dialog(AMUserSettings::userDataFolder, localRootDirectory, remoteRootDirectory, dataDirectory, parent);
-	dialog.exec();
+        AMChooseDataFolderDialog dialog(AMUserSettings::userDataFolder, localRootDirectory, remoteRootDirectory, dataDirectory, parent);
+        dialog.exec();
 
-	if (dialog.result() == QDialog::Accepted){
+        if (dialog.result() == QDialog::Accepted){
 
-		// when we are using this dialog, we are not user-account based storage
-		AMUserSettings::userBasedDataStorage = false;
+                // when we are using this dialog, we are not user-account based storage
+                AMUserSettings::userBasedDataStorage = false;
 
-		QString dialogInput = dialog.filePath();
+                QString dialogInput = dialog.filePath();
 
                 if (!dialog.isFullPath()){
 
-			QFileInfo remoteFullPath(QString("%1/%2/%3").arg(remoteRootDirectory).arg(dataDirectory).arg(dialogInput));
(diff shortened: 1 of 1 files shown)


dretrex:
Hey @mButlerDM, if this is all ready to go then place the "Ready for Review" label, assign an estimate (best guess) and assign it to a person.

mButlerDM:
Time Estimate: 1 Hour

mButlerDM:
@XuetaoLiu-CLS Could you review when you have a chance? Thanks.

dretrex:
Hey @mButlerDM, sorry, I meant a ZenHub estimate.  The time estimate should go with the issue itself.

To see the estimate you need to get the ZenHub plugin (for chrome or firefox).

mButlerDM:
@dretrex Okay thanks, I think I have it sorted now.

source/ui/util/AMChooseDataFolderDialog.cpp
```diff
@@ -46,12 +48,19 @@ bool AMChooseDataFolderDialog::getDataFolder(const QString &localRootDirectory,
...
-				AMUserSettings::save();
-			}
+                        else if(successfulFirstTimeStartup){
+
+                                AMUserSettings::userDataFolder = QString("%1/%2/%3/userData/").arg(localRootDirectory).arg(dataDirectory).arg(dialogInput);
+                                AMUserSettings::remoteDataFolder = QString("%1/%2/%3/userData/").arg(remoteRootDirectory).arg(dataDirectory).arg(dialogInput);
+                                AMUserSettings::save();
+                        }
+                        else{
+                                //If this is not the first startup and first startup was not successful:
```
XuetaoLiu-CLS: do we need to check the `extradataDirectory` as what we did on Ln43

XuetaoLiu-CLS:
Looks good other than a question on `extraDataDirectory`

Time Estimate: 5 minutes

source/ui/util/AMChooseDataFolderDialog.cpp
dretrex: Good catch.

mButlerDM: The main bug was caused by bad .ini configurations, so I only checked folders referenced in the config file. Ideally `extraDataDirectory` entries are always made at Ln43 before we reach the cases for a good or bad startup wizard run. `extraDataDirectory` entries were not validated before, maybe they should be? I'm more then happy to change it but that was my reasoning.

dretrex:
Looks good.
Time Estimate: 1 minute

PR #2357, opened 2016-05-27
Pull Request for Issue2356: Show/hide sidebar panes

helfrij:
Created a new class AMWindowPaneProxyModel, a subclass of QSortFilterProxyModel. This class basically will reassign QModelIndexes based on given sort or filter rules. In this case, the proxy model will ignore window pane items that have the AMWindowPaneModel::IsVisibleRole set to false and include all others. The proxy model replaces AMWindowPaneModel as the model for the sidebar. By default, all panes are visible, but can use new methods in AMMainWindow to show or hide panes as appropriate. Modified BioXASSideAppController to only show the Ge detector view if the detector is connected.

### DIFF
--- a/source/application/BioXAS/BioXASSideAppController.cpp
+++ b/source/application/BioXAS/BioXASSideAppController.cpp
@@ -44,10 +44,31 @@ bool BioXASSideAppController::startup()
 
 	return result;
 }
+#include <QDebug>
+void BioXASSideAppController::onGeDetectorConnectedChanged()
+{
+    BioXAS32ElementGeDetector *detector = BioXASSideBeamline::bioXAS()->ge32ElementDetector();
+    QWidget *detectorView = componentViewMapping_.value(detector, 0);
+    QWidget *detectorPane = viewPaneMapping_.value(detectorView, 0);
+
+    if (detector && detectorView && detectorPane) {
+	qDebug() << "\n\nSide Ge detector connected changed.";
+	qDebug() << "Detector connected:" << (detector->isConnected() ? "Yes" : "No");
+
+	if (detector->isConnected()) {
+	    mw_->showPane(detectorPane);
+	} else {
+	    mw_->hidePane(detectorPane);
+	}
+    }
+}
 
 void BioXASSideAppController::initializeBeamline()
 {
 	BioXASSideBeamline::bioXAS();
+
+	connect( BioXASSideBeamline::bioXAS()->ge32ElementDetector(), SIGNAL(connected(bool)), this, SLOT(onGeDetectorConnectedChanged()) );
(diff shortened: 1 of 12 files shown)


helfrij:

source/application/BioXAS/BioXASSideAppController.cpp
```diff
@@ -45,6 +59,10 @@ bool BioXASSideAppController::setupDataFolder()
 void BioXASSideAppController::initializeBeamline()
 {
 	BioXASSideBeamline::bioXAS();
+	connect( BioXASSideBeamline::bioXAS()->ge32ElementDetector(), SIGNAL(connected(bool)), this, SLOT(onGeDetectorConnectedChanged()) );
+	onGeDetectorConnectedChanged();
```
XuetaoLiu-CLS: This might cause a crash since `initializationBeamline()` is called in the early stage of `startup()` and `setupUserInterface()` is called after that.

`onGeDtectorConnectedChanged()` will try to update the mw_.

The best place for this call (Ln62, Ln63) is in `setupUserInterfaceImplementation()`.

source/application/BioXAS/BioXASSideAppController.cpp
```diff
@@ -45,6 +59,10 @@ bool BioXASSideAppController::setupDataFolder()
...
+
+	setupScanConfigurations();
```
XuetaoLiu-CLS: `setupScanConfigurations()` is called in `CLSAppController::startup()`, you don't need to call it here.

If you think this should happen earlier, you might want to change it in `CLSAppController::startup()`

source/application/BioXAS/BioXASSideAppController.cpp
XuetaoLiu-CLS: All UI related operations should only happen in the virtual functions called by `setupUserInterface()`, and beamline specific stuff should go to `setupUserInterfaceImplementation()`.

XuetaoLiu-CLS:
Looks good other than a flaw in initializing the beamline instance, which might cause a crash.

Time Estimate: 10 minutes

source/application/BioXAS/BioXASSideAppController.cpp
helfrij: This is probably a mistake on my part in trying to merge in the recent changes from master. I'll remove this call.

source/application/BioXAS/BioXASSideAppController.cpp
helfrij: It doesn't cause a crash, because there is no entry for the detector in the component-view mapping until the view is created.

dretrex:
Looks good.
Time Estimate: 10 minutes

PR #2363, opened 2016-05-29
Pull Request for Issue2362: BioXAS scan crashes in AMUtility/AM1DDerivativeAB

helfrij:
Made some small modifications to AM1DDerivativeAB that seem to help solve the problem.
- Modified the limit of one for-loop to be `totalSize-1` instead of `totalSize`.
- Added if-check to see if the `totalSize` is greater than 0 before updating cached values.

The main problem seems to be that the `computeCachedValues()` is being called when the size of the data and axis values appears to be 0, and this can happen midway through a scan. Sometimes this condition causes a crash, sometimes it doesn't, but when the crash does happen, this seems to be the situation.

### DIFF
--- a/source/analysis/AM1DDerivativeAB.cpp
+++ b/source/analysis/AM1DDerivativeAB.cpp
@@ -168,18 +168,28 @@ void AM1DDerivativeAB::setInputSource()
     emitAxisInfoChanged();
     emitInfoChanged();
 }
-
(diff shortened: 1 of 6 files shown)


helfrij:

dretrex:
Looks good to me.  Sorry about that @helfrij :(
Time Estimate: 1 minute

Issue #1628, opened 2015-09-25
Disable scan state controls when they aren't able to be changed.


helfrij:
Brian B. attempted to pause a scan on BioXASMain, but the scan didn't actually pause. The application continued with the scan and it collected and exported the scan data, and then froze when the scan should have ended. The scan data can be viewed as normal when the application is closed and reopened.

helfrij:
Talked to Brian B. more about this, and may likely be caused by one of two things:
- 'Pause' button was pressed during scan initialization
- scan was paused to change low-level bragg motor settings on EDM screen.

I was able to successfully reproduce the error by pressing 'Pause' during scan initialization. It looks like the application doesn't actually freeze, but in the workflow the scan still appears to be running and can't be stopped/cancelled. Application must be restarted to start other scans.

@dretrex, is pausing scans that are initializing something worth looking into?

helfrij:
This was discussed at the BioXAS meeting this morning. A better solution right now is to disable pause/resume/skip buttons when they can't be used: during scan initialization and cleanup.

XuetaoLiu-CLS:
Time Estimate: 3 days

PR #2366, opened 2016-05-31
PullRequest for Issue1628: Disable the pause buttons when doing scan initialization

XuetaoLiu-CLS:
#1628 
- when doing scanAction initialization, we will not allow pausing for action runner
- some code refactors


XuetaoLiu-CLS:
Tested with the MockAcquaman, pause button is disabled when initialization. Need more testing on real beamline, and also need to test the auto pause feature

XuetaoLiu-CLS:
@helfrij can you review this?

source/acquaman/BioXAS/BioXASGenericStepScanConfiguration.cpp
```diff
@@ -32,3 +34,63 @@ AMScanController* BioXASGenericStepScanConfiguration::createController()
 
 	return controller;
 }
+
+bool BioXASGenericStepScanConfiguration::usingAnyGeDetector() const
```
helfrij: Just curious, why are these methods added here?

source/ui/actions3/AMActionRunnerCurrentViewBase.cpp
```diff
@@ -43,8 +44,8 @@ void AMActionRunnerCurrentViewBase::onCurrentActionChanged(AMAction3 *action)
 	cancelButton_->setDisabled((action == 0));
 	skipButton_->setDisabled(true);
 
-	if (action)
-		pauseButton_->setEnabled(action->canPause());
+//	if (action)
```
helfrij: Do you need this commented out code?

source/application/AMAppController.cpp
```diff
@@ -139,12 +139,28 @@ bool AMAppController::startup(){
...
-	AMDatamanAppControllerForActions3::shutdown();
+	AMActionRunner3::releaseScanActionRunner();
+	AMActionRunner3::releaseWorkflow();
+
 	AMBeamline::releaseBl();
 	AMStorageRing::releaseStorageRing();
 	AMProcessVariableSupport::shutdownChannelAccess();
 }
 
+bool AMAppController::startupBeforeUserInterface()
```
helfrij: Did this code used to live in AMDatamanAppController? Why has it been moved here?

source/actions3/AMActionRunner3.h
```diff
@@ -83,6 +83,10 @@ class AMActionRunner3 : public QObject
 	void addActionToQueue(AMAction3* action) { insertActionInQueue(action, -1); }
 	/// Insert an action at a position \c index in the queue.  If \c index is -1 or > queuedActionCount(), will append to the end of the queue. If \c index is 0, will insert at the beginning of the queue. This class takes ownership of the \c action and will log and delete it after running it.
 	void insertActionInQueue(AMAction3* action, int index);
+	/// Insert an action at a position \c index in the queue.  If \c index is -1 or > queuedActionCount(), will append to the end of the queue. If \c index is 0, will insert at the beginning of the queue. This class takes ownership of the \c action and will log and delete it after running it.
+	bool InsertActionToQueue(AMAction3 *action, int index);
```
helfrij: It looks like the only difference in naming between this new method and the one above it is that this one starts with a capital letter. Is this intentional? Do they do different things?

source/ui/actions3/AMActionRunnerCurrentViewBase.cpp
XuetaoLiu-CLS: no. don't need it anymore

source/actions3/AMActionRunner3.h
XuetaoLiu-CLS: the name is slightly different. The original function called `insertAction**In**Queue`. The added function has the common codes to do the real "insert" into queue, so I named it `insertActionToQueue`. It should be protected.

source/application/AMAppController.cpp
XuetaoLiu-CLS: No. This is newly added codes. Try to initialize the workflow action runner and the scan action runner before we initialize UI.

This is also one of the main changes for this issue: synchronize the scanAction status between the workflow action runner and the scan action runner.

source/acquaman/BioXAS/BioXASGenericStepScanConfiguration.cpp
XuetaoLiu-CLS: I remembered these functions were living in the BioXASBeamlineSupport.h. The purpose of those functions are to check whether a given scan configuration is using a detector/control/scaler or not. Technically, the scan configuration knows whether they are using what. Therefore, this should be better to be a property of the scan configuration class instead of a static function.

helfrij:
Hey @XuetaoLiu-CLS, looks good! Thanks for going through it with me. 😄

Time Estimate: 20 minutes.

source/acquaman/AMGenericStepScanConfiguration.h
```diff
@@ -57,6 +57,11 @@ class AMGenericStepScanConfiguration : public AMStepScanConfiguration
 	/// Returns the regions of interest.
 	QList<AMRegionOfInterest *> regionsOfInterest() const { return regionsOfInterest_; }
 
+	/// Returns whether the scan configuration has axis of the given control
+	bool usingControl(AMControlInfo controlInfo) const;
```
dretrex: This should be a `const AMControlInfo &`.

source/actions3/AMActionRunner3.cpp
```diff
@@ -320,78 +275,27 @@ void AMActionRunner3::internalDoNextAction()
...
-			box.setDefaultButton(cancelRunningButton);
-
-			box.exec();
-			if(box.clickedButton() == cancelRunningButton){
-				cancelCurrentAction();
-				return;
-			}
-		}
+		AMAction3* newAction = removeActionFromQueue(0);
+		setCurrentAction(newAction);
```
dretrex: This part is a great refactor 💯

source/actions3/AMActionRunner3.cpp
```diff
@@ -476,6 +372,150 @@ void AMActionRunner3::onImmediateActionStateChanged(int state, int previousState
...
+			return false;
+		}
+	}
+
+	return true;
+}
+
+void AMActionRunner3::onCurrentActionStarting()
+{
+	if(loggingDatabase_)
```
dretrex: I'm sure that this was in the old code and I've just forgotten, but I just wanted to double check: is this essentially saying that logging is being initialized for this action?

source/actions3/AMActionRunner3.cpp
```diff
@@ -476,6 +372,150 @@ void AMActionRunner3::onImmediateActionStateChanged(int state, int previousState
...
+			parentLogId = parentAction->logActionId();
+
+		AMActionHistoryModel3 *historyModel = AMAppControllerSupport::actionHistoryModelFromDatabaseName(loggingDatabase()->connectionName());
+		if(!historyModel || !historyModel->logUncompletedAction(currentAction_, loggingDatabase_, parentLogId)){
+			AMErrorMon::report(AMErrorReport(this, AMErrorReport::Alert, -200, "There was a problem logging the uncompleted action to your database.  Please report this problem to the Acquaman developers."));
+		}
+	}
+
+	if (isScanAction())
+		emit scanActionCreated((AMScanAction *)currentAction());
```
dretrex: This should be a qobject_cast.

source/actions3/AMActionRunner3.cpp
```diff
@@ -1045,4 +1085,30 @@ bool AMActionRunner3::runActionImmediatelyInQueue(AMAction3 *action)
...
 }
 
+bool AMActionRunner3::isActionRunnerPausable() const
+{
+	bool isPausable = isActionRunnerPauseEnabled_;
+
+	if (currentAction_) {
+		AMAction3::State currentActionState = currentAction_->state();
+
+		isPausable = isPausable && currentAction_->canPause() && (currentActionState == AMAction3::Running || currentActionState == AMAction3::Paused);
```
dretrex: The case where `currentActionState == AMAction3::Paused` meant to catch if the action was paused programmatically rather than through the action runner?

source/ui/actions3/AMActionRunnerCurrentViewBase.cpp
```diff
@@ -176,6 +178,12 @@ void AMActionRunnerCurrentViewBase::onCancelButtonClicked()
 	}
 }
 
+void AMActionRunnerCurrentViewBase::onActionRunnerPausableChanged(bool pausable)
+{
+	pauseButton_->setEnabled(pausable);
+	pauseButton_->repaint(); // force repaint;
```
dretrex: If memory serves me correctly, I think you want to call `update()` rather than repaint.  Conversely, why do you need to force it to repaint anyway?  Usually widgets are pretty responsive to this kind of change.

dretrex:
Just a couple questions.  Overall, this is a great refactor!  Awesome work, @XuetaoLiu-CLS!
Time Estimate: 30 minutes

source/actions3/AMActionRunner3.cpp
XuetaoLiu-CLS: yes. When instantiating AMActionRunner3, we need to pass the instance of a loggingDatabase_. We should not need to do this check. Otherwise, we should add protection everywhere we are using `loggingDatabase_`

source/actions3/AMActionRunner3.cpp
XuetaoLiu-CLS: the function `isScanAction()` already did a `qobject_cast` check. It is safe to direct cast here.

source/actions3/AMActionRunner3.cpp
XuetaoLiu-CLS: I think so. This code is originally from the `void AMActionRunnerCurrentViewBase::onStateChanged(int state, int previousState)`. 

I was thinking that it is not necessary. But I keep it just in case I missed something

dretrex: 👍

source/actions3/AMActionRunner3.cpp
```diff
@@ -434,8 +434,7 @@ bool AMActionRunner3::validateAction(AMAction3 *newAction, const QString &okButt
 
 void AMActionRunner3::onCurrentActionStarting()
 {
-	if(loggingDatabase_)
-		currentAction_->setIsLoggingFinished(false);
+	currentAction_->setIsLoggingFinished(false);
```
dretrex: You might want to keep the if check.  I wasn't sure, but wanted to bring it up.

PR #2369, opened 2016-06-02
Pull Request for Issue2178: Adding basic user configuration

helfrij:
- Created new class for the PGM user configuration.
- Added instance to PGMAppController, with support methods.

### DIFF
--- a/source/application/PGM/PGMAppController.h
+++ b/source/application/PGM/PGMAppController.h
@@ -22,6 +22,7 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 #ifndef PGMAPPCONTROLLER_H
 #define PGMAPPCONTROLLER_H
 
+#include "dataman/PGM/PGMUserConfiguration.h"
 #include "application/CLS/CLSAppController.h"
 
 class PGMAppController : public CLSAppController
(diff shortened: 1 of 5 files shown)


helfrij:

source/application/PGM/PGMAppController.h
```diff
@@ -58,6 +70,13 @@ class PGMAppController : public CLSAppController
 	virtual void createDetectorPanes();
 	/// create pane for the scan configuration views
 	virtual void createScanConfigurationPanes();
+
+	/// Returns true if the list of regions of interest contains the given ROI.
+	bool containsRegionOfInterest(QList<AMRegionOfInterest*> roiList, AMRegionOfInterest *regionOfInterest) const;
```
XuetaoLiu-CLS: There are two usages for this function now (PGM and BioXAS). Then I think it would be nicer to move this function to both AMXRFDetector and AMUserConfiguration, and call this function in is  addRegion function, instead of here, with two reasons:
- from OO design, it would be nicer that the classes can be encrypted well. The interfaces/functions provided will let callers DO the job without worrying about how. For this case, the caller will only care about addRegion() and don't need to worry about whether I should add it. It's up to the AMXRFDetector and the UserConfiguration to worry about it;
- from the functionality view, if we add the check in AMXRFDetector and UserConfiguration, all the beamlines will benefit from this without code changes (this is also the intention of the OO design. As long as the interfaces are the same, the classes can change the implementation without affecting the callers)

XuetaoLiu-CLS: We can live with this now though. and an issue for code refactor will be good too.

XuetaoLiu-CLS:
Looks good other than a code refactor suggestion.

Time Estimate: 10 minutes

helfrij:
Created new issue to handle @XuetaoLiu-CLS's suggested refactor (#2375).

Issue #2374, opened 2016-06-03
Build an AMScanAxisView


dretrex:
For the EXAFS beamlines we have a nice multi-region AMEXAFSScanAxisView and we should have one for PGM's XAS scan.  I created the class already but not implemented any details yet.  I imagine it should have a _very_ familiar look to the AMEXAFSScanAxisView and provide similar functionality.

I imagine once we finish it we will see some ways of having the EXAFS view to subclass the regular one, but that isn't the priority for this issue.

dretrex:
Turns out there is already an AMStepScanAxisView and I couldn't find it...
Time Estimate: 1.5 hours

dretrex:
Done in #2382

PR #2376, opened 2016-06-03
Pull Request for Issue2171: PGM Scanning - XAS

dretrex:
Implemented XAS scanning.  I created the scan configuration, controller, and a shell of a view.  I had to make a PGMPicoAmmeter which uses the synchronized dwell time.

The scan configuration view still needs to be done though as well as some finer polish.


source/application/PGM/PGMAppController.cpp
```diff
@@ -84,12 +88,27 @@ void PGMAppController::registerExporterOptions()
 
 void PGMAppController::setupScanConfigurations()
 {
-
+	xasScanConfiguration_ = new PGMXASScanConfiguration;
+	xasScanConfiguration_->setControl(0, AMBeamline::bl()->exposedControlByName("Energy")->toInfo());
+//	xasScanConfiguration_->addDetector(AMBeamline::bl()->exposedDetectorByName("teyBladeCurrentDetector")->toInfo());
```
iainworkman: I assume these are being left in as we're not yet certain which detector's Lucia is going to require for the XAS scans?

source/ui/dataman/AMScanAxisView.cpp
```diff
@@ -0,0 +1,8 @@
+#include "AMScanAxisView.h"
+
+
+AMScanAxisView::AMScanAxisView(const QString &title, AMStepScanConfiguration *configuration, QWidget *parent)
+	: QWidget(parent)
+{
+
+}
```
iainworkman: To be done in #2374

iainworkman:
One comment. Given the trivial nature of the comment, and the fact that everything else looks good I will merge.
Time Estimate: 30 minutes.

source/application/PGM/PGMAppController.cpp
dretrex: Yup.

PR #2380, opened 2016-06-03
Pull Request for Issue2373: Show window items by default.

helfrij:
Made a small modification to AMWindowPaneModel. When queried for the visibility of a particular window item, the model will return true if there is no visibility setting specified for the item, instead of returning an invalid QVariant. This should ensure that items with no visibility set should show up, but the setting is obeyed for others.


helfrij:

XuetaoLiu-CLS:
Looks good.

Time Estimate: 2 minutes

Issue #2375, opened 2016-06-03
Minor ROIs refactor -- checking for duplicates


helfrij:
BioXAS and PGM applications both check whether the user configurations and the XRF detectors have an ROI before adding them to each class, in the app controller. This was added to BioXAS because there was a problem there where duplicate ROIs were being added to the detector, and PGM got that change as well. It would probably be a good idea to check whether this is something we can add to the user configuration/app controller and XRF detector at a more general level, just in case this happens in other beamline applications.

XuetaoLiu-CLS:
Time Estimate: 1 hour

PR #2390, opened 2016-06-08
PullRequest for Issue2375 code refactor: create CLSUserConfiguration to generalize ROI usages…

XuetaoLiu-CLS:
- create CLSUserConfiguration to generalize the usage of ROIs
- check whether the ROI exists or not before adding it to the userConfiguration and AMXRFDetectors

#2375


helfrij:
Looks good, @XuetaoLiu-CLS! I was wondering, because the user configurations are AMDbObjects, do these changes require a database upgrade?

XuetaoLiu-CLS:
@helfrij I don't think so. We didn't modify any db schema here. It should be good

dretrex:
I agree with @helfrij. I'm not confident that this is going to work without a database upgrade.  The fact that the inheritance changes might cause issues (i.e.: AMDbObjectTypes_table).  Unfortunately, there is more to the database system than just looking at the end result table.

XuetaoLiu-CLS:
@helfrij @dretrex 
Just tested on SXRMB:
1) it can open exising db, with ROI configured, without issue;
2) it can create a new database, with the same DB table as the original ones (SXRMBUserConfiguration_table and SXRMBUserConfiguration_table_regionOfInterest). 

The `CLSUserConfiguration` class here is only acting as an intermediate class, which is not acting as  a physical table. Therefore, I think we should be fine.

I'll do more testing when I can use beamline to do some scans.

XuetaoLiu-CLS:
@dretrex @helfrij I added the DB upgrade to add CLSUserConfiguration_table and CLSUserConfigration_table_regionsOfInterest; also updated the inheritence of beamline UserConfiguration information

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
```diff
@@ -0,0 +1,134 @@
...
+			databaseToUpgrade_->rollbackTransaction();
+			AMErrorMon::alert(this, CLSDBUPGRADE1PT3_COULD_NOT_CREATE_TABLE, QString("Could not create %1.").arg(clsConfigurationTableName));
+			return false;
+		}
+
+		QString configurationROITableName = QString("%1_regionsOfInterest").arg(clsConfigurationTableName);
+		QStringList ROITableColumnNames = QStringList() << "id1" << "table1" << "id2" << "table2";
+		QStringList ROITableColumnTypes = QStringList() << "INTEGER" << "TEXT" << "INTEGER" << "TEXT";
+
+		if (!databaseToUpgrade_->ensureTable(configurationROITableName, ROITableColumnNames, ROITableColumnTypes)) {
```
dretrex: This will likely need to use AMDbUpgradeSupport::addTable like you used above to make sure all the auxiliary information in other tables are properly handled.

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
```diff
@@ -0,0 +1,134 @@
...
+			databaseToUpgrade_->rollbackTransaction();
+			AMErrorMon::alert(this, CLSDBUPGRADE1PT3_COULD_NOT_CREATE_TABLE, QString("Could not create %1.").arg(configurationROITableName));
+			return false;
+		}
+
+		QStringList indexColumns = QStringList() << "id1" << "id2" ;
+		foreach (QString column, indexColumns) {
+			if (!databaseToUpgrade_->createIndex(configurationROITableName, column)) {
```
dretrex: Does this work as expected?  I've never used createIndex and I usually hate using things I'm not familiar with when it comes to database upgrades :(

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+		QVariantList dboClassIds = databaseToUpgrade_->retrieve("AMDbObjectTypes_table", "id");
+		QVariantList dboClassNames = databaseToUpgrade_->retrieve("AMDbObjectTypes_table", "AMDbObjectType");
+		int index = dboClassNames.indexOf(beamlineUserConfigurationClassName);
+		if ( index >= 0) {
+
+			dboValues = QVariantList() << beamlineUserConfigurationClassName << QString("%1_table").arg(beamlineUserConfigurationClassName) << tableDescriptionTemplate.arg(beamlineName)
+									   << 1 << QString("%1;%2").arg(beamlineUserConfigurationClassName).arg(clsUserConfigurationClassInheritence);
+
+			int dbResult = databaseToUpgrade_->insertOrUpdate(dboClassIds.at(index).toInt(), "AMDbObjectTypes_table", dboColumnNames, dboValues);
```
dretrex: If this is inserting to the AMDbObjectTypes_table then you need to make sure that the AMDbObjectTypes_allColumns, AMDbObjectTypes_visibleColumns, and AMDbObjectTypes_loadColumns are properly populated.

dretrex:
Time Estimate: 20 minutes

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
XuetaoLiu-CLS: This will only update the inheritance information, which won't change any thing related to the columns

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
XuetaoLiu-CLS: Yes it works as expected. it created the create index statements.

```
CREATE TABLE CLSUserConfiguration_table_regionsOfInterest
(
    id INTEGER PRIMARY KEY,
    id1 INTEGER,
    table1 TEXT,
    id2 INTEGER,
    table2 TEXT
);
CREATE UNIQUE INDEX idx_CLSUserConfiguration_table_regionsOfInterest_id2 ON CLSUserConfiguration_table_regionsOfInterest (id2);
CREATE UNIQUE INDEX idx_CLSUserConfiguration_table_regionsOfInterest_id1 ON CLSUserConfiguration_table_regionsOfInterest (id1);
```

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
XuetaoLiu-CLS: @dretrex this is to create the `CLSUserConfiguration_table_regionsOfInterest` table, which is an auxiliary table itself. the `addTable()` function will do the same as the `ensureTable()`

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
dretrex: Okay.

source/dataman/CLS/CLSDbUpgrade1Pt3.cpp
dretrex: Oh yes.  That's my bad.  Actually, now that I think about it, would addTable actually be a bad idea since it auto-adds things to the other tables?  You might have been more correct with your original code.

dretrex:
@XuetaoLiu-CLS, can you revert that last commit you did (#dfff7d5)?  Getting you to use that method was a mistake on my part and your original solution is the preferred method.

XuetaoLiu-CLS:
@dretrex the changes is reverted.

dretrex:
Great!  Thanks @XuetaoLiu-CLS.

Issue #2377, opened 2016-06-03
Polish up the XAS scan configuration


dretrex:
Now that scanning is working and once we implement the scan axis view we will need to add some polish to ensure that selections and configurations are possible.  Such selecting whether you are using the Ocean Optics spectrometer or other standard options.  We also need to incorporate the Ocean Optics Spectrometer into the scan configuration and controller in a similar way as is used on other beamlines.

mButlerDM:
Time Estimate: 1 day

PR #2391, opened 2016-06-09
Pull Request for Issue 2377: Add additional functionality to PGM scan configuration view.

mButlerDM:
Added in a detector selection component similar to the one used by BioXAS.

Setup the export options and added a check box to enable export, default is enabled.

Adjusted visual components to be cleaner. 

Corrected double entry in pri file.

Corrected component load order in PGMBeamline to load setupControlSets later in the list to resolve a segfault on launch.

#2377

### DIFF
--- a/source/application/PGM/PGMAppController.cpp
+++ b/source/application/PGM/PGMAppController.cpp
@@ -222,9 +222,9 @@ void PGMAppController::createDetectorPanes()
 
 void PGMAppController::createScanConfigurationPanes()
 {
-        xasScanConfigurationView_ = new PGMXASScanConfigurationView(xasScanConfiguration_);
-        xasScanConfigurationViewHolder3_ = new AMScanConfigurationViewHolder3(xasScanConfigurationView_, true);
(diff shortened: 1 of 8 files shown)


mButlerDM:
@helfrij Could you review when you have a chance? Thanks!

source/application/PGM/PGMAppController.cpp
```diff
@@ -137,7 +141,9 @@ void PGMAppController::registerDBClasses()
 
 void PGMAppController::registerExporterOptions()
 {
-
+        AMExporterOptionXDIFormat *pgmXASExportOptions = PGM::buildXDIFormatExporterOption("PGMXASDefault", true);
```
helfrij: In some of the scan configurations on other beamlines, we make an additional call to build exporter options in the scan configuration constructor, pretty much just repeating these calls made here. Just wanted to check whether that was needed on PGM as well.

helfrij:
Looks good, @mButlerDM. One minor question.

Time Estimate: 10 minutes.

dretrex:
Generally, this looks good.  @helfrij brings up a good point, which should be addressed.  Also, there are some indent issues.

Time Estimate: 1 minute

Issue #2354, opened 2016-05-26
Refactor BioXAS to use AMControl child control methods with bool returns.


mButlerDM:
Refactor code logic to take advantage of changes to addChildControl() removeChildControl() and clearChildControls() in AMControl and inherited classes in AM.

mButlerDM:
Time Estimate: 2 Days

Issue #2351, opened 2016-05-26
Refactor REIXS/CLS to use AMControl child control methods with bool returns.


mButlerDM:
Refactor code logic to take advantage of changes to addChildControl() removeChildControl() and clearChildControls() in AMControl and inherited classes in AM.

mButlerDM:
Time Estimate: 1 hour

XuetaoLiu-CLS:
@mButlerDM remember to change the AM related classes as well.

PR #2313, opened 2016-05-16
Pull Request for Issue 2271: Modify AMControl child methods to return success bool

mButlerDM:
Changed the definitions within AMControl to return bool types and modified inherited classes and usages throughout the code base to simplify the code.

Issue #2271, opened 2016-04-26
Modify AMControl child methods to return success bool


helfrij:
It would be handy if `AMControl::addChildControl(...)`, `removeChildControl(...)`, and `clearChildControls()` returned a success bool. This issue would address implementing this new return value in these methods as well as reimplementations, and looking for ways the bool can simplify code logic.

mButlerDM:
Changes AM done in #2313

PR #2396, opened 2016-06-09
Pull request for Issue 2354: Refactor BioXAS classes for new AMControl childcontrol methods.

mButlerDM:
Adjust logic and control checking patterns to reduce inline overhead and take advantage of the new AMControl childControl method refactor.

In progress while I handle conflicts with master after BioXAS class were generalized for use with PGM.

#2354


mButlerDM:
@XuetaoLiu-CLS Would you mind reviewing when you have a chance?

Based on changes done in Issue #2271 and pull request #2313. There are a lot of changes but it is mostly refactoring the same code pattern.

Thanks!

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorPositionControl.cpp
```diff
@@ -71,38 +71,26 @@ void BioXASCarbonFilterFarmActuatorPositionControl::setMotor(CLSMAXvMotor *newCo
...
-			addChildControl(status_);
-
-			connect( status_, SIGNAL(valueChanged(double)), this, SLOT(onStatusControlValueChanged()) );
-		}
-
-		updateStates();
+            removeChildControl(status_); // disconnects all signals.
+            status_ = newControl;
+            addChildControl(status_);
+            connect( status_, SIGNAL(valueChanged(double)), this, SLOT(onStatusControlValueChanged()) );
```
XuetaoLiu-CLS: what will happen if the `status_` is NULL?

XuetaoLiu-CLS: A better solution is that change the signature of the `addChildControl()` function to return a boolean value. Therefore, here it will be like:
```
if (addChildControl(status_)
    connect(status_, SIGNAL(valueChanged(double)), this, SLOT(onStatusControlValueChanged()) );
```

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
```diff
@@ -116,195 +116,167 @@ void BioXASSSRLMonochromatorRegionControl::setUpperSlitBladeControl(AMControl *n
...
 	}
 }
 
 void BioXASSSRLMonochromatorRegionControl::setCrystalChangeCWLimitStatusControl(AMControl *limitStatus)
 {
 	if (crystalChangeCWLimitStatus_ != limitStatus) {
 
-		if (crystalChangeCWLimitStatus_)
-			removeChildControl(crystalChangeCCWLimitStatus_);
+                removeChildControl(crystalChangeCCWLimitStatus_);
```
XuetaoLiu-CLS: can u confirm here. I think we should remove `crystalChange**_CW_**LimitStatus` instead of `crystalChange**_CCW_**LimitStatus`

XuetaoLiu-CLS:
A formatting issue:
  @mButlerDM all the new added/changed lines are indented with spaces. We were using "tabs" instead of "spaces". Can you fix this? Otherwise, it makes the codes looks bad. You should also change your settings of QT creator: "Tools" --> "Options" --> "Text Editor" --> "Behavior" tab: uncheck the "insert spaces instead of tabs"

XuetaoLiu-CLS:
the following codes should be changed as well:

- [ ] `void AMBeamlineControl::setStatusControl() Ln71`
- [ ] `bool AMBeamlineControlGroup::clearControls() Ln56, we should check the returns of clearChildContorls() as well`
- [ ] 'bool AMSlit::removeChildControl(AMControl *control)' should changed to 
```
if (AMControl::removeChildControl(control)){
        return !allControls_->removeControl(control);
    }
```
- [ ] same changes required for `bool AMSlit::addChildControl(AMControl *control)`
- [ ] same changes required for `bool AMSlits::removeChildControl(AMControl *control)` 
- [ ] same changes required for `bool AMSlits::removeChildControl(AMControl *control)` )
- [ ] void CLSSIS3820ScalerAcquisitionMode::setScanCountControl(AMControl *newControl)
- [ ] void CLSSIS3820ScalerAcquisitionMode::setNumberOfScansPerBufferControl(AMControl *newControl)
- [ ] void CLSSIS3820ScalerAcquisitionMode::setStartScanControl(AMControl *newControl)
- [ ] void SGMVATValveState::setSpeed(AMControl *newControl)
- [ ] bool AMCompositeControl::addChildControl(AMControl *control) doesn't have a return value
- [ ] `AMSlit::AMSlit(const QString &name, QObject *parent) :` although the `addChildControl()` used here will likely always return true, but still it might be better to check the result before the `connect()`
- [ ] void AMSlit::setFirstBlade(AMControl *newControl, BladeOrientation orientation), should check the return value of addChildControl() before adding to the map
- [ ] void AMSlit::removeFirstBlade() / void AMSlit::removeSecondBlade(), might be better to return bool as well, and change the correspoing usages
- [ ] void AMSlit::setFirstBlade / void AMSlit::setSecondBlade()
- [ ] void AMSlitControl::setFirstBlade() / setSecondBlade
- [ ] void AMSlitControl::removeBlade(AMControl *control)

XuetaoLiu-CLS:
@mButlerDM open the `acquaman_suit.pro` and search for the usages of `addChildControl()`/`removeChildControl()`/`celarChildControl()`. 
There might be more changes required than I listed.

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorPositionControl.cpp
mButlerDM: @XuetaoLiu-CLS My concern here was changing behavior, since status would be status_ would be set to newControl regardless. If you think this is okay I'll go ahead and make the changes.

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
mButlerDM: @XuetaoLiu-CLS This is the case in master, maybe it is a typo?

mButlerDM:
CLS classes are handled in #2351 
SGM classes are to be ignored.

I can go through and fix my spacing. I lost my config at one point and I likely did the bulk of this during that time.

AM classes are handled in #2313  so I will do the changes there and back merge them here. 

@XuetaoLiu-CLS @dretrex My concerns for this code pattern were a change in behavior

```
		removeChildControl(status_); // disconnects all signals.

		status_ = newControl;

        	if (status_) {
 			addChildControl(status_);

		connect( status_, SIGNAL(valueChanged(double)), this, SLOT(onStatusControlValueChanged()) );
```

The childControl methods can handle a null status_ value but the assignment of status_ happened regardless. Should this change?

In the case that a child control is added twice, addChildControl will fail, and the connection will never be created. Is this concern for making connect dependent on addChildControl return values?

XuetaoLiu-CLS:
@mButlerDM that's a fair concern. If the new `status_` is already added. It will be like the following calling sequence:
```
- setStatus(control) // the first time call, it will set status_ to control, and connect the signal
- setStatus(control) // call it again. This time, the status_ equals to control, we won't do any chagnes
```
or
```
- setStatus(control) // the first time call, it will set status_ to control, and connect the signal
- setStatus(anotherControl) // status_ doesn't equal to anotherControl, we will remove the current status_(control), add the anotherControl and set status_ to anotherControl;
- setStatus(control) // call it again. This time, the status_ doesn't equal to control, we will do the same as the step above.
```

XuetaoLiu-CLS:
As another assumption: assume we can reach `addChildControl` with a duplicated control, then the codes:
```
if (addChildContorl(status_)
    connect(status, ....);
```
is still valid. The reason is that if the current `status_` is already added (`addChildControl()` will return false) and we won't need to connect the signal/slot again since it was already connected when the `status_` was added for the first time.

XuetaoLiu-CLS:
Since the CLS/AM/REXIS refactor will be handled in #2351, this PR looks good for BioXAS changes other than some small code changes.

Time Estimate: 30 minutes.

source/beamline/BioXAS/BioXASCarbonFilterFarmActuatorPositionControl.cpp
XuetaoLiu-CLS: @mButlerDM as I mentioned in the other comments, this won't change the behavior. We only need to connect the signal/slot when the new status_ can be added as a child. If it can't, then either status_ is null or it is already added (and the signal/slot is connected already).

source/beamline/BioXAS/BioXASSSRLMonochromatorRegionControl.cpp
XuetaoLiu-CLS: @mButlerDM I believe it is a typo. This method is handling *CW* control and there is another method handles the *CCW* control.

XuetaoLiu-CLS:
Looks good with the new changes.

dretrex:
Looks good.
Time Estimate: 5 minutes

Issue #2394, opened 2016-06-09
Use CLSBeamlineStatus/CLSShutter/CLSValves for SXRMB


XuetaoLiu-CLS:
try to use the above general class for SXRMB, so that we can try to generalize the usage of Beam on/off feature (#2355)

XuetaoLiu-CLS:
Time Estimate: 4 hours

PR #2397, opened 2016-06-09
PullRequest for Issue2394: using CLSBeamlineStatus in SXRMB

XuetaoLiu-CLS:
#2394 
- fix a bug to display the current selected main window pane (a missing mapping issue)
- use CLSBeamlineStatus and CLSBeamlineStatus view for SXRMB
  * added a new icon beamlineStatus.png to show the general beamline status
  * added a beamline status pv to the CLSBeamlineStatus, since SXRMB has a general beamline status pv. If it is not assigned, the default beamlineStatus PV of CLSBeamlineStatus is itself
 * added a flag to display the beamline status in the compact view (this might be removed in the future once the use is used to to the beamline status icon)


XuetaoLiu-CLS:
@iainworkman can you review this? Thx

iainworkman:
Looks good to me.
Time Estimate: 20 minutes.

source/beamline/CLS/CLSBeamlineStatus.h
```diff
@@ -42,6 +42,8 @@ class CLSBeamlineStatus : public CLSBiStateGroup
 	/// Returns the beam on value for the given control.
 	double componentBeamOnValue(AMControl *control) const { return controlState1ValueMap_.value(control, -1); }
 
+	/// Returns the beamline status PV
+	AMControl * beamlineStatusPVControl() { return beamlineStatusPVControl_ ? beamlineStatusPVControl_ : this; }
```
dretrex: What is this line?

source/ui/CLS/CLSBeamlineStatusView.cpp
```diff
@@ -72,7 +75,7 @@ void CLSBeamlineStatusView::setBeamlineStatus(CLSBeamlineStatus *newStatus)
 
 void CLSBeamlineStatusView::setSelectedComponent(AMControl *newControl)
 {
-	if (selectedComponent_ != newControl && beamlineStatus_->components().contains(newControl)) {
+	if (selectedComponent_ != newControl && (beamlineStatus_->components().contains(newControl) || beamlineStatus_ == newControl)) {
```
dretrex: What does the extra OR'ed comparison help with?

dretrex:
Couple of questions, otherwise looks good.
Time Estimate: 8 minutes

source/beamline/CLS/CLSBeamlineStatus.h
XuetaoLiu-CLS: SXRMB has a beamline status PV, which can tell us whether beam is ready or not. I think Vespers might have one as well.

If a beamline doesn't have the beam status PV, then we will use CLSBeamlineStatus as the PV, which will mark beam ready or not via checking the status of the shutters and valves

source/ui/CLS/CLSBeamlineStatusView.cpp
XuetaoLiu-CLS: I added a button in the BeamlineStatusView to show the overall beam status (whether beam is ready or not).

source/beamline/CLS/CLSBeamlineStatus.h
dretrex: I think this should return the control (which is initialized to 0) and returns 0 if not set.  Then the things that use it just have a piece of logic that is:

```c++
if (control->beamlineStatusPVControl())
    control->beamlineStatusPVControl()->stuff();
else
    control->stuff();
```

XuetaoLiu-CLS: @dretrex the usage of function `beamlineStatusPVControl()` is in `CLSBeamlneStatusView`:
```
	beamStatusEditor_->setControl(beamlineStatus_->beamlineStatusPVControl());
```
We need to return AMControl to do what we want. The original codes in CLSBeamlineStatusView is
```
	beamStatusEditor_->setControl(beamlineStatus_);
```
which is the case when there is no beam status PV.

source/ui/CLS/CLSBeamlineStatusView.cpp
XuetaoLiu-CLS: When that overall beam status button is clicked, the selected control is the beamlineStatus itself. That's why I added a OR comparison

source/beamline/CLS/CLSBeamlineStatus.cpp
```diff
@@ -40,6 +40,14 @@ bool CLSBeamlineStatus::isOff() const
 	}
 }
 
+AMControl * CLSBeamlineStatus::beamlineStatusPVControl()
+{
+	if (beamlineStatusPVControl_ )
+		return beamlineStatusPVControl_ ;
+	else
+		return this;
```
dretrex: I would prefer that you don't return `this`.  This should `return 0;` and have this logic outside the class.

dretrex: Return 0 as in:

```c++
if (beamlineStatusPVControl_)
    return beamlineStatusPVControl_;

return 0;
```

dretrex: Or just `return beamlineStatusPVControl_;` and have it initialized to 0 in the constructor.

XuetaoLiu-CLS: @dretrex I didn't agree with this changes. The purpose for this function is to return a signal control, which can be used to check the overall status of the beam. 

For that purpose, if there is a general PV control defined for beam status (`beamlineStatusPVControl_`), we should return that one; otherwise, the instance of `CLSBeamlineStatus` will behave for that purpose. The caller of `beamlineStatusPVControl()` should not be aware of whether they should use a general PV control or the CLSBeamlineStatus instance.

Btw, I will rename the function `beamlineStatusPVControl(()` to `beamlineStatusControl()`

dretrex: But that breaks the paradigm used everywhere else in the framework.  To be consistent, there should be a flag like `hasExternalStatusControl()` or something similar and then you know to ask for the control explicitly through `beamlineStatusControl()` or use the control itself.

The change by itself isn't really a big deal, but it's context in relation to the rest of the code is important.

XuetaoLiu-CLS: the usage exists in ` void CLSBeamlineStatusView::refresh()' and the changes  are

{ 	
// Update the beam status editor.
 - if (editor_) 	
-     editor_->setControl(beamlineStatus_); 	
 + if (beamStatusEditor_) {
+     beamStatusEditor_->setControl(beamlineStatus_->beamlineStatusControl());
+ }
```
I think my changes fit the usage here.

dretrex: I still don't like it because the behaviour that comes from calling it this way doesn't necessarily match the expected control flow.

I am prepared to merge this code in, but I want to emphasize that programming in this way is not clear and will likely cause misleading behaviour when debugging it in the future.  Personally, I prefer code clarity than simplicity - which should be paramount considering that we spend most of our time debugging old, hard to read code.

XuetaoLiu-CLS: @dretrex I changed this usage based on ur comments.

PR #2398, opened 2016-06-09
Pull Request for Issue2389: Investigating QSortFilterProxyModel crashes

helfrij:
Modified `AMMainWindow::expandAllHeadings()` to use `AMMainWindow::expandHeadingIndex(...)` instead of `QTreeView::expandAll()`. This seemed to be the cause of the startup crashes.

### DIFF
--- a/source/application/AMAppController.cpp
+++ b/source/application/AMAppController.cpp
@@ -149,12 +149,8 @@ bool AMAppController::startupCreateUserInterface() {
 
 	if (AMDatamanAppControllerForActions3::startupCreateUserInterface()){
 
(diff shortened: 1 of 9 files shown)


helfrij:

iainworkman:
Interesting that a crash should be caused by calling a method in a standard Qt class. I wonder if it's pointing to some underlying problem in the model?
Otherwise looks good.
Time Estimate: 20 minutes.

helfrij:
Yeah. These are the changes that seem to make the crashes stop--honestly, I just can't say much more than that at this point.

dretrex:
Looks good.
Time Estimate: 3 minutes

Issue #1476, opened 2015-07-11
Default print-to-file directory causes sync conflict


epengr:
AM crashed and there was a sync conflict when it was re-opened. It turned out to be a "print.pdf" in the network folder.  

It seems that a use wanted to print the screen and left the default option which was print to file into the network data folder.   When it didn't actually print they tried again and chose the actual printer instead of pdf, but didn't think to find a remove the unwanted pdf (who would?).  

The default print-to-file folder should be the exoprtData folder (I think), to avoid this (at 5 am on a Saturday...).

davidChevrier:
Solution: Find a way to ensure the default path does not contain _userData_.

mButlerDM:
Time Estimate: 4 Hours

PR #2404, opened 2016-06-13
Pull request for Issue 1476: Change default print location to exportData

mButlerDM:
The default print and export locations will some times be userData instead of exportData. Added in-method checks to insure paths do not contain userData and to use a default path option if that is the case.

Adjusted file naming so that print-to-file scans are set to a default name scheme, based on date and time, to avoid name conflicts or overwrites.

Fixed a bug in the AMGenericScanEditor that would save an exported file with the parent's directory name in the parent's folder. This occurs when the export window is opened, no name is specified and the user selected cancel. QFileDialog with no name entry or selection would return the window directory path by default and AMGenericScanEditor would tag on an extension.

#1476

### DIFF
--- a/source/ui/dataman/AM2DScanView.cpp
+++ b/source/ui/dataman/AM2DScanView.cpp
@@ -556,10 +556,7 @@ void AM2DScanView::exportGraphicsFile(const QString& fileName)
 		if(fileName.contains("userData"))
 				printer.setOutputFileName(AMUserSettings::defaultAbsoluteExportFolder() + "/" + QDateTime::currentDateTime().toString("yyyy-MM-dd--hh-mm-ss"));
 		else
-			if(fileName.endsWith("/"))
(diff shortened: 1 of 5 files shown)


mButlerDM:
@iainworkman Could you review when you have a moment? Thanks!

iainworkman:
I do wonder if a more long term solution for this might be in a kind of AMDataSafeFileDialog, which detects the UserData folder at the input layer, and prevents it being selected (with a message to the users so they known what's going on). I think that is best saved for another issue, however.

This seems like a perfectly good solution without getting into the above complexities.
Time Estimate: 10 minutes.

dretrex:
@iainworkman, I think that should just be AMFileDialog unless we want to have an AMLiveWithTheConsequencesFileDialog or AMDestructoFileDialog! ;)

Looks good.
Time Estimate: 2 minutes

PR #2405, opened 2016-06-13
Pull Request for Issue2333: Adding move progress reporting to AMControl

helfrij:


helfrij:

source/beamline/AMControl.h
```diff
@@ -633,6 +663,26 @@ protected slots:
 	/*! If isEnum() returns true and this is not specified by a subclass implementation, the regular enumNames() will be assumed to apply for both move() and value(). moveEnumNames() will return enumNames(). */
 	void setMoveEnumStates(const QStringList& enumStateNames) { if(moveEnumNames_ == enumStateNames) return; moveEnumNames_ = enumStateNames; emit enumChanged(); }
 
+	/// Sets the move start value.
+	void setMoveStart(double newValue);
+	/// Updates the move start value.
+	virtual void updateMoveStart();
```
XuetaoLiu-CLS: I didn't get the purpose of `updateMoveState()`, `updateMoveValue()` and `updateMoveEnd()`. those functions are called in the `move()` function of the PVControl. I think we should only need to `updateMoveValue()`. The other two are set before we perform the move and won't be changed when we are moving.

XuetaoLiu-CLS: never mind. I think I got it...

source/beamline/BioXAS/BioXASSSRLMonochromatorEnergyControl.cpp
```diff
@@ -194,6 +194,25 @@ AMAction3* BioXASSSRLMonochromatorEnergyControl::createCalibrateAction(double ol
...
+	// The bragg motor is what's used to change the energy.
+	// We calculate the move progress based on the bragg motor positions.
+
+	if (bragg_) {
+		double braggMoveStart = bragg_->moveStart();
+		double braggMoveValue = bragg_->moveValue();
+		double braggMoveEnd = bragg_->moveEnd();
+
+		if (braggMoveStart != braggMoveEnd)
+			result = qAbs(braggMoveValue - braggMoveStart) / (braggMoveEnd - braggMoveStart);
```
XuetaoLiu-CLS: is there sth like `brag_->calculateMoveProgressPercent()`?

XuetaoLiu-CLS:
Looks good other than a question on code optimization.

Time Estimate: 5 minutes

source/beamline/AMControl.h
helfrij: All three values should be updated before the move starts, so that we use _this_ move's information. The move value is what is updated to report the move progress (ie it could change every time the control value changes). Does that answer your question?

source/beamline/BioXAS/BioXASSSRLMonochromatorEnergyControl.cpp
helfrij: There is, as the bragg motor control inherits from AMControl. But this function is protected.

source/beamline/AMPVControl.cpp
```diff
@@ -249,6 +250,13 @@ AMControl::FailureExplanation AMPVControl::move(double setpoint) {
 			return NotConnectedFailure;
 		}
 		setpoint_ = setpoint;
+
+		// Update the move progress values.
+
+		updateMoveStart();
```
dretrex: Since all three of these functions are typically called together this way, maybe make a convenience method that calls all three when it is appropriate to do so?

source/ui/CLS/CLSControlEditor.cpp
```diff
@@ -67,18 +69,8 @@ void CLSControlEditor::setControl(AMControl *newControl)
 			connect( control_, SIGNAL(minimumValueChanged(double)), this, SLOT(updateMinimumValue()) );
 			connect( control_, SIGNAL(maximumValueChanged(double)), this, SLOT(updateMaximumValue()) );
 
-			connect( control_, SIGNAL(movingChanged(bool)), this, SLOT(refresh()) );
```
dretrex: Wow, that's a lot of connects that were removed.  Nice!

dretrex:
One question.  Looks great!

Time Estimate: 5 minutes

source/beamline/AMPVControl.cpp
helfrij: That's a good idea. I was wondering, though, does it make sense for the three values (start, value, end) to be implemented in AMControl? Or do they more logically belong to a subclass?

dretrex: Well, it might be that AMControl only has a percentage and everyone else has the responsibility of turning start and end points to percentages.

Issue #2355, opened 2016-05-26
add beam on/off actions and buttons to the CLSBeamlineStatusView


XuetaoLiu-CLS:

dretrex:
I'm taking this out of the milestone as Lucia will get back to us on what/if she wants this.

XuetaoLiu-CLS:
Time Estimate: 1 days

PR #2406, opened 2016-06-14
PullRequest for Issue2355: add generic beam on/off action to the CLSBeamlineStatusView and CLSShutters/CLSValves

XuetaoLiu-CLS:
#2355


source/beamline/CLS/CLSShutters.cpp
```diff
@@ -123,20 +123,36 @@ bool CLSShutters::isClosed() const
 	return result;
 }
 
+bool CLSShutters::isSafetyShutterOpen() const
+{
+	if (safetyShutter_) {
+		return safetyShutter_->isConnected() && isChildState1(safetyShutter_);
+	}
+
+	return true;
```
dretrex: I think that it is probably better to return false as a default for whether this kind of equipment is open.

source/beamline/CLS/CLSShutters.cpp
```diff
@@ -123,20 +123,36 @@ bool CLSShutters::isClosed() const
...
+		if (safetyShutter_) {
+			addShutter(safetyShutter_);
+		}
+
+		emit safetyShutterChanged(safetyShutter_);
 	}
 }
 
-bool CLSShutters::addShutter(AMControl *newShutter, double openValue, double closedValue, int beamOnOrder)
+bool CLSShutters::addShutter(AMControl *newShutter, int beamOnOrder, double openValue, double closedValue)
```
dretrex: Have many people used this function?  If yes, you'll want to make sure you've gone over every implementation to ensure that it's being added correctly.

source/beamline/CLS/CLSValves.cpp
```diff
@@ -93,7 +93,7 @@ QList<AMControl*> CLSValves::closedValvesList() const
 	return childrenInState2();
 }
 
-bool CLSValves::addValve(AMControl *newValve, double openValue, double closedValue, int beamOnOrder)
+bool CLSValves::addValve(AMControl *newValve, int beamOnOrder, double openValue, double closedValue)
```
dretrex: Same with this.

dretrex:
Oh, woops.  I didn't bother looking at the fact that this wasn't ready for review.  Sorry about that @XuetaoLiu-CLS!

XuetaoLiu-CLS:
I just tested on SXRMB, the beam on/off is working as expected. By far, the generic beam on/off is only used by SXRMB. I will try to bring it to the other beamlines (Vespers and Ideas) later

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -204,22 +205,6 @@ void BioXASAppController::onRegionOfInterestBoundingRangeChanged(AMRegionOfInter
 		genericConfiguration_->setRegionOfInterestBoundingRange(region);
 }
 
-void BioXASAppController::goToBeamlineStatusView(AMControl *control)
```
helfrij: Just curious--why is this functionality removed?

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -389,7 +374,7 @@ void BioXASAppController::createGeneralPanes()
 	QWidget* beamlineConfigurationView = createComponentView(BioXASBeamline::bioXAS());
 	addMainWindowView( beamlineConfigurationView, "Configuration", generalPaneCategeryName_, generalPaneIcon_);
 
-	beamlineStatusView_ = new CLSBeamlineStatusView(BioXASBeamline::bioXAS()->beamStatus(), false);
+	beamlineStatusView_ = new CLSBeamlineStatusView(BioXASBeamline::bioXAS(), false);
```
helfrij: It looks like the declaration of the `beamlineStatusView_` was removed from the .h file, but we still create an instance of the view here. Does this cause a crash?

source/application/CLS/CLSAppController.h
```diff
@@ -30,6 +32,9 @@ protected slots:
 	/// implementation for slot that connects generic scan editors that use the 2D scan view to the app controller so that it can enable quick configuration of scans.
 	virtual void onScanEditorCreatedImplementation(AMGenericScanEditor *editor);
 
+	/// Sets the beam status view as the current view, with the given control as the selected control.
```
helfrij: Ah, I see. The method was moved up to CLSAppController.

source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -100,8 +100,7 @@ bool AMPseudoMotorControl::validSetpoint(double value) const
 
 bool AMPseudoMotorControl::addChildControl(AMControl *control)
 {
-	if(AMControl::addChildControl(control)){
-		connect( control, SIGNAL(connected(bool)), this, SLOT(updateStates()) );
```
helfrij: Does this connect statement appear elsewhere, so it doesn't need to be here?

source/beamline/BioXAS/BioXASBeamline.cpp
```diff
@@ -513,19 +513,6 @@ bool BioXASBeamline::clearGe32Detectors()
 	return true;
 }
 
-void BioXASBeamline::setConnected(bool isConnected)
```
helfrij: Is this functionality replaced elsewhere?

source/beamline/BioXAS/BioXASBeamline.h
```diff
@@ -418,14 +408,9 @@ protected slots:
 	void addControlAsDetector(const QString &name, const QString &description, AMControl *control, bool hiddenFromUsers = false, bool isVisible = true);
 
 	/// Protected constructor.
-	BioXASBeamline(const QString &controlName);
+	BioXASBeamline(const QString &beamlineName, const QString &controlName);
```
helfrij: What is the difference between the beamline name and the control name?

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
```diff
@@ -12,13 +12,12 @@ BioXASFrontEndShutters:: BioXASFrontEndShutters(const QString &name, QObject *pa
 
 	upstreamPhotonShutter_ = 0;
 	downstreamPhotonShutter_ = 0;
-	safetyShutter_ = 0;
 
 	// Current settings.
 
 	setUpstreamPhotonShutter(new AMReadOnlyPVControl("IPSH1407-I00-01", "IPSH1407-I00-01:state", this));
 	setDownstreamPhotonShutter(new CLSExclusiveStatesControl("IPSH1407-I00-02", "IPSH1407-I00-02:state", "IPSH1407-I00-02:opr:open", "IPSH1407-I00-02:opr:close", this));
-	setSafetyShutter(new CLSExclusiveStatesControl("SSH1407-I00-01", "SSH1407-I00-01:state", "SSH1407-I00-01:opr:open", "SSH1407-I00-01:opr:close", this));
```
helfrij: Do we manage the safety shutter elsewhere, instead of with the other front-end shutters?

source/beamline/CLS/CLSBeamline.cpp
```diff
@@ -1,12 +1,227 @@
...
+void CLSBeamline::onBeamOffActionFinished(){
+	disconnect( beamOffAction_, 0, this, 0 );
+
+	beamOffAction_->deleteLater();
+	beamOffAction_ = 0; //NULL
+}
+
+/// ==================== protected methods =======================
+
+void CLSBeamline::createBeamlineStatus(CLSShutters *shutters, CLSValves *valves)
```
helfrij: Just curious--will the beamline status only ever consider shutters and valves?

source/beamline/CLS/CLSBeamlineStatus.cpp
```diff
@@ -10,18 +13,20 @@ CLSBeamlineStatus::CLSBeamlineStatus(const QString &name, QObject *parent) :
 	monoMaskControlSet_ = new AMControlSet(this);
 
 	// Setup the basic value options.
-	addOption(On, "On", true);
-	addOption(Off, "Off", true);
+	addOption(On, "Beam is ON", true);
+	addOption(Off, "Beam is OFF", true);
+
+	connect(this, SIGNAL(connected(bool)), this, SLOT(updateBeamlineStatus()));
```
helfrij: I'm always a little leery about connecting our own signals to our own slots. Is it possible to make the calls to `updateBeamlineStatus()` in the places where these signals are originally emitted?

source/beamline/CLS/CLSBeamlineStatus.cpp
```diff
@@ -10,18 +13,20 @@ CLSBeamlineStatus::CLSBeamlineStatus(const QString &name, QObject *parent) :
...
+	connect(this, SIGNAL(componentsChanged()), this, SLOT(updateBeamlineStatus()));
 }
 
 CLSBeamlineStatus::~CLSBeamlineStatus()
 {
-	shuttersControlSet_->clear();
-	valvesControlSet_->clear();
-	mirrorMaskControlSet_->clear();
-	monoMaskControlSet_->clear();
+	clearComponents();
```
helfrij: Does `clearComponents()` handle deleting components?

source/beamline/CLS/CLSShutters.cpp
```diff
@@ -1,11 +1,16 @@
...
 #include "beamline/AMControl.h"
+#include "actions3/AMActionSupport.h"
 #include "actions3/AMListAction3.h"
 
+#include "util/AMErrorMonitor.h"
+
 CLSShutters::CLSShutters(const QString &name, QObject *parent) :
 	CLSTriStateGroup(name, parent)
 {
+	safetyShutter_ = 0;
```
helfrij: Does the safety shutter belong here, conceptually? Must all groups of shutters have a safety shutter?

source/beamline/CLS/CLSShutters.h
```diff
@@ -60,6 +82,13 @@ public slots:
 
 	/// Returns the index for the current value.
 	virtual int currentIndex() const;
+
+protected:
+	/// the safety shutter of a beamline
+	AMControl *safetyShutter_;
+
+	/// the order of the shutters in the beam on action. If a shutter is not invovled in the beam on/off action, the order is -1
+	QMap<int, AMControl*> shuttersBeamOnOrderMap_;
```
helfrij: I like the inclusion of an order. Nice!

source/beamline/CLS/CLSShutters.cpp
```diff
@@ -14,7 +19,88 @@ CLSShutters::CLSShutters(const QString &name, QObject *parent) :
...
+AMListAction3* CLSShutters::createBeamOffActionList()
+{
+	// create the action list to move the Shutters (sequentially) and wait for the move done
+	AMListAction3 *closeShuttersActionList = new AMListAction3(new AMListActionInfo3("Close shutters action list", "Close Shutters"), AMListAction3::Sequential);
+	AMListAction3 *waitShuttersCloseActionList = new AMListAction3(new AMListActionInfo3("Wait shutters closed action list", "Wait Shutters Closed"), AMListAction3::Parallel);
+
+	// this is to make sure all the controls are checked
+	int currentBeamOffOrder = 1;
+	int checkedControlCount = 0;
+	while (checkedControlCount < shuttersBeamOnOrderMap_.count()) {
```
helfrij: Just checking: shutters are opened in order of map key, and closed in the order in which they were added to the map?

source/stylesheets/AMToolButton.qss
```diff
@@ -53,7 +53,7 @@ background-color: #800000;
 /* Styling for colorState = Neutral. */
 
 AMToolButton[colorState="3"] {
-background-color: qlineargradient(x1:0 y1:0, x2:0 y2:1, stop:0 #ffff00, stop:1 #808000);
+ background-color: qlineargradient(x1:0 y1:0, x2:0 y2:1, stop:0 #ffffff, stop:1 #808080);
```
helfrij: Could this change be handled in #2449 instead?

source/ui/CLS/CLSValueEditor.cpp
```diff
@@ -55,6 +55,11 @@ CLSValueEditor::~CLSValueEditor()
 
 }
 
+void CLSValueEditor::addLayout(QLayout *newLayout)
```
helfrij: Is having this functionality necessary? I can see it being a source of complexity/confusion down the road.

source/beamline/CLS/CLSBeamlineStatus.cpp
```diff
@@ -33,11 +38,7 @@ bool CLSBeamlineStatus::isOn() const
...
 {
-	if (beamlineStatusPVControl_) {
-		return (!isChildState1(beamlineStatusPVControl_));
-	} else {
-		return (isConnected() && !areAllChildrenState1());
-	}
+	return isConnected() && !isOn();
 }
 
 AMControl * CLSBeamlineStatus::beamlineStatusPVControl()
```
helfrij: I don't think BioXAS does, and I'm just curious--do many beamlines have a beamline status PV control like this?

source/beamline/CLS/CLSShutters.cpp
```diff
@@ -14,7 +19,88 @@ CLSShutters::CLSShutters(const QString &name, QObject *parent) :
 
 CLSShutters::~CLSShutters()
 {
+	shuttersBeamOnOrderMap_.clear();
+}
+
+AMListAction3* CLSShutters::createBeamOnActionList()
```
helfrij: Is there something beam on/off specific about these action lists? Or are they just opening/closing the shutters and valves? If it's the latter case, we could think about them instead in terms of a shutters move action (eg. `moveToOpen()`, `moveToClosed()`) and reimplement `AMPseudoMotorControl::createMoveAction(double)`.

helfrij:
A couple of questions.

Time Estimate: 1 hour.

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: it is moved to CLSAppController

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: it is defined in the CLSAppController. This instantiation here might be able to moved to CLSAppController when we are having more beamlines using the CLSBeamStatusView

source/beamline/AMPseudoMotorControl.cpp
XuetaoLiu-CLS: `AMPseudoMotorControl` is a subclass of `AMConnectedControl`. Here I changed to use `AMConnectedControl::addChildChontrol(control)`, which should be a proper usage and also did the connected() signal/slot connection

source/beamline/BioXAS/BioXASBeamline.cpp
XuetaoLiu-CLS: `setConnected()` was called by `updateConnected()` slot. And `updateConnected()` was replaced with `onBeamlineComponentConnected()` slot in the `CLSBeamline`

source/beamline/BioXAS/BioXASBeamline.h
XuetaoLiu-CLS: `beamlineName` is like a short version of the name, `controlName` is like a longer version, more detailed information.

This is primarily used for the messaging purpose

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
XuetaoLiu-CLS: it is still here. I added two parameters to the `setSafetyShutter()` function.

XuetaoLiu-CLS: @helfrij the implementation of `setSafetyShutter()` is moved to `CLSShutters`

source/beamline/CLS/CLSBeamline.cpp
XuetaoLiu-CLS: no. it has shutters/valves/morror mask/monoMask as BioXAS has before.

Shutters and Valves are two common components cross beamlines. For convenience, we pass in shutters and valves here. at Ln306/ln307 of BioXASSideBeamline.cpp, you can see we added the MirrorMask and MonoMask

source/beamline/CLS/CLSBeamlineStatus.cpp
XuetaoLiu-CLS: basically, CLSBeamlineStatus is a group of controls. The signals (`connected()`/`valueChanged()`) of any control managed by CLSBeamlineStatus will trigger CLSBeamlineStatus to emit the same signal.

In this way, connect to our own signals will avoid the mistake of missing a signal from the child controls.

source/beamline/CLS/CLSBeamlineStatus.cpp
XuetaoLiu-CLS: no. It only clears the control set.

source/beamline/CLS/CLSShutters.cpp
XuetaoLiu-CLS: No. The saftyShutter is not mandatory.

source/beamline/CLS/CLSShutters.cpp
XuetaoLiu-CLS: it is just opening/closing the shutters/valves. and yes, it will be better to reimplement the `AMPseudoMotorControl::createMoveAction(double)`

source/beamline/CLS/CLSShutters.cpp
XuetaoLiu-CLS: it is still in order of map keys. The next line of codes gets the shutter in the order. This `while` loop will make sure we go through all the ordered shutters

source/ui/CLS/CLSValueEditor.cpp
XuetaoLiu-CLS: nice catch. I forgot why I added this, but looks like it is never used. I will remove it

XuetaoLiu-CLS:
@helfrij thanks for going through this giant PR. Lots of codes are changed. I addressed most of your comments. Please take a look.

I will try to use the `createMoveAction()` to replace the `createMoveOnActions()`

dretrex:
Hey @XuetaoLiu-CLS, once you replace the createMoveOnActions let me know and I'll go over it then!

source/beamline/CLS/CLSBeamlineStatus.cpp
XuetaoLiu-CLS: SXRMB has one. BioXAS doesn't. This is optional.

XuetaoLiu-CLS:
@dretrex @helfrij I just checked the CLSShutters/CLSValves. There were an implementation of `createMoveAction()`, which will create a move action to open/close all the shutters/valves.

For beam on/off action list, we need to create the open/close actions in a given order, which might NOT include all the shutters/valves. 

I think it might be better to keep the current solution so far. What do you think?

source/application/PGM/PGMAppController.cpp
```diff
@@ -186,14 +171,14 @@ void PGMAppController::setupUserInterfaceImplementation()
 void PGMAppController::createPersistentView()
 {
 	PGMPersistentView *pgmPersistentView = new PGMPersistentView;
-	connect( pgmPersistentView, SIGNAL(beamlineStatusSelectedComponentChanged(AMControl*)), this, SLOT(goToBeamlineStatusView(AMControl*)) );
+	connect( pgmPersistentView, SIGNAL(beamlineStatusControlClicked(AMControl*)), this, SLOT(goToBeamlineStatusView(AMControl*)) );
```
dretrex: Was this connect statement not altered in BioXASAppController or was it done in a different pull request?

source/application/SXRMB/SXRMBAppController.cpp
```diff
@@ -195,9 +189,9 @@ void SXRMBAppController::onBeamAvailabilityChanged(bool beamAvailable)
 bool SXRMBAppController::setupDataFolder()
 {
 	// Get a destination folder.
-	return AMChooseDataFolderDialog::getDataFolder("/AcquamanLocalData/sxrmb",  //local directory
-												   "/home/sxrmb",               //remote directory
-												   "acquamanData",              //data directory
+	return AMChooseDataFolderDialog::getDataFolder("/AcquamanLocalData",  //local directory
```
dretrex: This seems like a random change.  What was the reason for it?

source/beamline/BioXAS/BioXASBeamline.h
```diff
@@ -66,7 +61,7 @@ class BioXASBeamline : public CLSBeamline
 	static BioXASBeamline *bioXAS()
 	{
 		if (!instance_) {
-			instance_ = new BioXASBeamline("BioXAS Beamline");
+			instance_ = new BioXASBeamline("BioXASBeamline", "BioXAS Beamline");
```
dretrex: Based on your answer to @helfrij just below, I'd suggest the name be `BioXAS` (other beamlines being the same).  Otherwise, this short name will only ever be one space shorter than the long name which adds no value.

source/beamline/AMPseudoMotorControl.cpp
helfrij: Sounds good.

source/application/BioXAS/BioXASAppController.cpp
helfrij: Ah okay.

source/beamline/BioXAS/BioXASBeamline.h
helfrij: Oh okay. What's messaging?

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
helfrij: Ah, okay.

source/beamline/CLS/CLSBeamline.cpp
helfrij: It looks like the ring current could also be incorporated eventually, based on the update beam status method above.

source/beamline/CLS/CLSBeamlineStatus.cpp
helfrij: That probably makes sense, as the CLSBeamlineStatus likely isn't the parent of each component.

source/beamline/CLS/CLSShutters.cpp
helfrij: Oh yes, I see. I misinterpreted what was going on.

source/beamline/CLS/CLSBeamline.cpp
```diff
@@ -1,12 +1,227 @@
...
+		AMErrorMon::error(this, CLSBEAMLINE_ERR_BEAM_ON_ALREADY_ON, QString("Failed to create the beam on actions because the beam is already ready."), true);
+		return 0;
+	}
+
+	if(!beamlineShutters_->isConnected() || !beamlineValves_->isConnected()) {
+		AMErrorMon::error(this, CLSBEAMLINE_ERR_BEAM_ON_UNCONNECTED_PV, QString("Failed to create the beam on actions due to unconnected shutter/valve PVs."), true);
+		return 0;
+	}
+
+	if (!beamlineShutters_->isSafetyShutterOpen()) {
```
dretrex: Using VESPERS as an example, beam on would include opening a safety shutter and would therefore never make it past this statement.  Is there a means of bypassing that?

source/beamline/CLS/CLSBeamlineStatus.cpp
dretrex: @helfrij is there something in particular you don't like?  As long as there is no mechanism for signal-slot loops then this should be fine.

source/beamline/CLS/CLSShutters.cpp
helfrij: I think your solution should probably replace the existing `moveToOpen/ClosedAction()` functionality.

source/beamline/CLS/CLSBeamlineStatus.cpp
dretrex: VESPERS kind of does.  It's all over the map.

source/beamline/CLS/CLSBeamlineStatus.cpp
```diff
@@ -122,8 +134,15 @@ bool CLSBeamlineStatus::clearComponents()
 {
 	bool result = clearBiStateControls();
 
-	if (result)
+	if (result) {
+		shuttersControlSet_->clear();
```
dretrex: Use `clearComponents()` here?

source/beamline/CLS/CLSShutters.cpp
dretrex: To put this in context, is this meant to be the shutter that is not controlled by beamlines or the one that is?

source/beamline/CLS/CLSShutters.cpp
dretrex: Just to confirm: @helfrij, are you suggesting this code live inside `CLSShutters`?

source/beamline/CLS/CLSShutters.cpp
```diff
@@ -14,7 +19,88 @@ CLSShutters::CLSShutters(const QString &name, QObject *parent) :
...
+{
+	// create the action list to move the Shutters (sequentially) and wait for the move done
+	AMListAction3 *openShuttersActionList = new AMListAction3(new AMListActionInfo3("Open shutters action list", "Open Shutters"), AMListAction3::Sequential);
+	AMListAction3 *waitShuttersOpenActionList = new AMListAction3(new AMListActionInfo3("Wait shutters open action list", "Wait Shutters Open"), AMListAction3::Parallel);
+
+	// this is to make sure all the controls are checked
+	int currentBeamOnOrder = 1;
+	int checkedControlCount = 0;
+	while (checkedControlCount < shuttersBeamOnOrderMap_.count()) {
```
dretrex: Why don't you just grab all the keys from the map and then use a for loop to iterate through each of them?

source/beamline/CLS/CLSShutters.cpp
dretrex: Shouldn't the turn off order be the inverse of the turn on order?

source/beamline/CLS/CLSValves.h
```diff
@@ -16,6 +23,9 @@ class CLSValves : public CLSTriStateGroup
 	/// Destructor.
 	virtual ~CLSValves();
 
+	/// helper function to create a beam on action list
+	AMListAction3* createBeamOnActionList();
+
```
dretrex: No off?

source/beamline/CLS/CLSBeamlineStatus.cpp
helfrij: Given the option, I would prefer to reimplement the `AMEnumeratedControl::updateStates()` or `updateValue()` or something, and add whatever additional functionality is needed there (emitting the new `beamlineStatusChanged(bool)` signal).

source/beamline/CLS/CLSValves.h
helfrij: That might be a safety thing. I'm not sure we'd want to turn off beam by attempting to close all the valves.

source/beamline/SXRMB/SXRMBBeamline.cpp
```diff
@@ -670,32 +579,26 @@ void SXRMBBeamline::setupComponents()
...
 
-	beamlineShutters_->addShutter(SSH1406B1001Shutter_, CLSBEAMLINE_VALVE_OPEN, CLSBEAMLINE_VALVE_CLOSED);
-	beamlineShutters_->addShutter(PSH1406B1002Shutter_, CLSBEAMLINE_VALVE_OPEN, CLSBEAMLINE_VALVE_CLOSED);
-
-	beamlineValves_->addValve(VVR16064B1003Valve_, CLSBEAMLINE_VALVE_OPEN, CLSBEAMLINE_VALVE_CLOSED);
-	beamlineValves_->addValve(VVR16064B1004Valve_, CLSBEAMLINE_VALVE_OPEN, CLSBEAMLINE_VALVE_CLOSED);
-	beamlineValves_->addValve(VVR16064B1006Valve_, CLSBEAMLINE_VALVE_OPEN, CLSBEAMLINE_VALVE_CLOSED);
-	beamlineValves_->addValve(VVR16064B1007Valve_, CLSBEAMLINE_VALVE_OPEN, CLSBEAMLINE_VALVE_CLOSED);
-	beamlineValves_->addValve(VVR16065B1001Valve_, CLSBEAMLINE_VALVE_OPEN, CLSBEAMLINE_VALVE_CLOSED);
+	beamlineValves_->addValve(VVR16064B1003Valve, CLSExclusiveStatesControl::Open, CLSExclusiveStatesControl::Closed, 1);
```
dretrex: Quick question about SXRMB automatic valve opening.  Were there any waits between the opening of consecutive valves?  If not, is this a feature they use often?  I know from past experience that valves are typically opened with a bit of time between valve to minimize pressure waves inside the vacuum chambers.

source/ui/CLS/CLSBeamlineStatusView.h
```diff
@@ -54,26 +62,43 @@ protected slots:
...
 	QLayout* createBeamlineStatusButtonBarLayout();
+
+	/// creates and returns a layout with the beam on/off buttons
+	QLayout* createBeamOnOffButtons();
+
 	/// Creates and returns a component view for the given control.
 	virtual QWidget* createComponentView(AMControl *control);
 
+	/// helper functions to create a push button
+	QPushButton *createPushButton(QString text);
```
dretrex: `const QString &`.  Also, technically, a `const` function too.

source/beamline/CLS/CLSShutters.cpp
helfrij: Yes, I don't think it makes sense to move it out of CLSShutters. There is an existing implementation of `CLSShutters::createMoveToOpen/ClosedActions()` in this class that doesn't take the newly implemented order into account; I think @XuetaoLiu-CLS's code in `CLSShutters::createBeamOn/OffActionList()` could replace the older code.

source/beamline/CLS/CLSBeamlineStatus.cpp
dretrex: Ah, gotcha.

source/beamline/BioXAS/BioXASBeamline.h
dretrex: I think he means when being used in warning message boxes or in the status view.

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
```diff
@@ -46,7 +45,7 @@ void BioXASFrontEndShutters::setUpstreamPhotonShutter(AMReadOnlyPVControl *newCo
 		upstreamPhotonShutter_ = newControl;
 
 		if (upstreamPhotonShutter_)
-			addShutter(upstreamPhotonShutter_, 1, 4);
+			addShutter(upstreamPhotonShutter_, CLS_SHUTTER_OPEN, CLS_SHUTTER_CLOSED);
```
helfrij: Quick question. Do these `CLS_SHUTTER_OPEN/CLOSED` values represent the PV open/closed, or the pseudomotor CLSShutters open/closed? Remember that those will be different values.

dretrex:
Lots of code and a bunch of questions, but overall looking alright.
Time Estimate: 50 minutes

source/beamline/BioXAS/BioXASBeamline.h
helfrij: 👍

source/application/PGM/PGMAppController.cpp
XuetaoLiu-CLS: in PR for Issue2346 (#2428), @helfrij altered it to  `beamStatusControlClicked()`

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
XuetaoLiu-CLS: @helfrij yes, that confused me before. the Open/Close value we used in the addShutter function should be the one we used for the control.

That is, if the control is a regular PV, the open/close should be 1/4 as the case here. If the control is a pseudomotor, the open/close should be 0/1 (or 1/0) (this depends on the open/close definition of that pseudomotor).

XuetaoLiu-CLS: @helfrij and also in this case, we were using the magic number (1, 4). I simply replaced it with the state definitions,

source/beamline/CLS/CLSBeamline.cpp
XuetaoLiu-CLS: @helfrij maybe. My purpose here is that for the CLSBeamline, it will try to have some CLS level generalization. Eventually, all the beamline status will be determined by the ring current and the status of the  shutters/valves. 

(at this moment, each beamline has its own implementation, but similar to each other, for that)

source/beamline/CLS/CLSBeamline.cpp
XuetaoLiu-CLS: @dretrex good to know about that. This version of generic beam on/off only works for SXRMB. If the safety shutter can be turned on programmatically, we will only make this check if the safety shutter is NOT in the beam on mapping.

source/beamline/CLS/CLSBeamlineStatus.cpp
XuetaoLiu-CLS: this is the implementation of `clearComponents()`.

source/beamline/CLS/CLSShutters.cpp
XuetaoLiu-CLS: @dretrex my understanding for this is that it is up to the user to make the decision how to use the safety shutter for CLSShutters.
For SXRMB:
The safety shutter is a Readonly PV control, which can't be turned on grammatically. The users complained several times that they can't turn beam on with acquaman. Acutally, the safety shutter is closed, but they don't know. With this changes, we will visually display the status of the safety shutter and also pop a message to tell them if the safety shutter is closed when they are trying to turn beam on.

For VESPERS, since we are able to open the safety shutter, we don't have to set the safety shutter, which can be identified as a regular shutter. If we still want to distinguish it with the other shutters, it will still works as I mentioned before.

To address @helfrij's question, safety shutter is not mandantory to the `CLSShutters`. You can use `CLSShutters` as what you did with the original `BioXASShutters`. It can be simply a group of shutters you want.

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
helfrij: Thanks @XuetaoLiu-CLS!

source/beamline/CLS/CLSShutters.cpp
dretrex: In that case, we should change that to not be safety shutter, but rather something like `addReadOnlyComponent` where the readonly components are important for the action to take place but cannot be done programmatically.

source/beamline/CLS/CLSShutters.cpp
XuetaoLiu-CLS: @dretrex I want to make sure they are ordered. And yes, I can use that. Just checked the documents:
```
QList<Key> QMap::keys() const
Returns a list containing all the keys in the map in ascending order.
```

source/beamline/CLS/CLSValves.h
XuetaoLiu-CLS: @dretrex we never turn Valves off when we turning beam off.
Also in the following implementation, it also mentioned that we don't support close valves
```
AMAction3* CLSValves::createMoveAction(double setpoint)
{
	AMAction3 *result = 0;

	if (setpoint == Open)
		result = createMoveToOpenAction();

	// Closing valves is not supported at this time.
	// This is because a user accidentally closing valves
	// could trigger machine protection protocols.

	return result;
}
```

source/beamline/CLS/CLSBeamlineStatus.cpp
dretrex: lol, so it is.

source/beamline/CLS/CLSValves.h
dretrex: I was thinking more as a general CLSValves aspect rather than being a coupled part of beam status.  Generally, the use case for closing valves is minimal but I didn't want us to not think about it.

source/beamline/SXRMB/SXRMBBeamline.cpp
XuetaoLiu-CLS: @dretrex there are waits actions. This is not a frequently used feature. Usually, the valves are all open. To put valves in the beam on actions is to make sure the beam will be ON if they want.

XuetaoLiu-CLS:
@dretrex @helfrij when going through the comments, I had a thought. It might be less confusing if I keep the `CLShutters/CLSValves` as it was. There should be a subclass, called `CLSBeamStatusShutters/CLSBeamStatusValves`, of CLShutters/CLSValves, which will take care of the safety shutters and beam on/off actions.

helfrij:
Lots of the changes, introducing the ordering and updating the move actions in particular, enhance the existing functionality of CLSShutters and CLSValves. If there are SXRMB-specific things you'd like to keep, perhaps they could go in an SXRMBShutters subclass?

dretrex:
To combine both ideas, I think it would make a lot of sense having the upgraded CLSValves/CLSShutters, but then making a CLSBeamStatus that has both of them.  There is no reason to make a beam status version of them.  Then we could still subclass the CLSValves/CLSShutters for specific beamline needs without compromising the beam status part.

XuetaoLiu-CLS:
@dretrex @helfrij I made the new changes to have a CLSBeamStatusShutters/CLSBeamStatusValves as a subclass of CLSShutters/CLSValves. Please take a look at the changes.

@dretrex I saw you comment. If you still think it will be better not to have the specified beamstatus shutters/valves after you reviewed the new changes, I can revert the changes.

source/beamline/AMConnectedControl.h
```diff
@@ -13,6 +13,11 @@ class AMConnectedControl : public AMControl
 	/// Destructor.
 	virtual ~AMConnectedControl();
 
+	/// Adds a given control to the list of child controls.
```
helfrij: What's the reason for moving these methods from public to protected?

dretrex: It's actually the opposite.  But a reason would still be appreciated.

source/beamline/CLS/CLSValves.h
```diff
@@ -13,8 +16,6 @@ class CLSValves : public CLSTriStateGroup
 
 	/// Constructor.
 	explicit CLSValves(const QString &name, QObject *parent = 0);
-	/// Destructor.
-	virtual ~CLSValves();
```
dretrex: Why was this removed?

source/ui/CLS/CLSBeamlineStatusView.h
```diff
@@ -54,26 +62,43 @@ protected slots:
...
 
+	/// helper functions to create a push button
+	QPushButton *createPushButton(const QString &text);
+
 protected:
 	/// flag to identify whether this is the compact status view or full status view
 	bool compactView_;
+
+	/// The beamline being operated
+	CLSBeamline *beamline_;
```
dretrex: This is unnecessary because there will only ever be one beamline which can always be accessed by `CLSBeamline::bl()`.

dretrex:
Hey @XuetaoLiu-CLS, after looking over it, I think extra stage of beam status shutters and valves obfuscates what you're trying to achieve.  I do still think my suggestion is the better way to go on this issue.

source/beamline/AMPseudoMotorControl.cpp
```diff
@@ -112,7 +112,7 @@ bool AMPseudoMotorControl::addChildControl(AMControl *control)
 
 bool AMPseudoMotorControl::removeChildControl(AMControl *control)
 {
-	if (AMConnectedControl::removeChildControl(control)){
+	if (AMControl::removeChildControl(control)){
```
helfrij: Why the change from AMConnectedControl to AMControl here?

source/beamline/BioXAS/BioXASFrontEndShutters.cpp
```diff
@@ -70,9 +70,9 @@ void BioXASFrontEndShutters::setDownstreamPhotonShutter(CLSExclusiveStatesContro
 AMAction3* BioXASFrontEndShutters::createMoveToOpenAction()
 {
 	AMListAction3 *actionList = new AMListAction3(new AMListActionInfo3("Opening front-end shutters", "Opening front-end shutters"), AMListAction3::Sequential);
-	actionList->addSubAction(createCheckChildIsOpenAction(upstreamPhotonShutter_, 10));
-	actionList->addSubAction(createMoveChildToOpenAction(safetyShutter_));
-	actionList->addSubAction(createMoveChildToOpenAction(downstreamPhotonShutter_));
+	actionList->addSubAction(createCheckChildIsOpen(upstreamPhotonShutter_, 10));
```
helfrij: Why the rename here? The 'action' part seems important, as that's what's actually returned.

XuetaoLiu-CLS:
@dretrex the changes for CLSBeamStatusShutters/CLSBeamStatusValves are reverted. The  CLSShutters/Valves was upgraded with the beam on/off feature; the CLSBeamlineStatus contains both CLSShutters/CLSValves and supports the beam on/off feature.

Issue #2345, opened 2016-05-20
Facilities combobox is blank


helfrij:
When starting a new run using BioXAS Main application, there is an option to enter a run name and select a facility. The facility combobox appears to be empty.

There may not be facility options to display anymore, given the recent refactor in #2225. If that's the case, the box should be removed.

XuetaoLiu-CLS:
Time estimate: 2 hours

helfrij:
Just tried creating a new run again on a branch up-to-date with master, and I see the same problem.

XuetaoLiu-CLS:
Time Estimate: 4 hours

PR #2408, opened 2016-06-17
PullRequest for Issue2345: fix the bug in AMDbObjectThumbnail table of the AMFacility data

XuetaoLiu-CLS:
- add CLSDbUpgrade1.2 to fix  a bug in CLSDbUpgrade1.1, which doesn't take care of the upgrade of AMDbObjectThumbnail_table for AMFacility thumbnails
- fix the bug of CLSDbUpgrade1.1 to generate the thumbnail for the new
facility is required

#2345

### DIFF
--- a/source/application/AMDatamanAppController.cpp
+++ b/source/application/AMDatamanAppController.cpp
@@ -242,9 +242,11 @@ bool AMDatamanAppController::startup() {
 			return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_HANDING_NON_FIRST_TIME_USER, "Problem with Acquaman startup: handling non-first-time user.");
 	}
 
-
 	if(!startupRegisterDatabases())
 		return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_REGISTERING_DATABASES, "Problem with Acquaman startup: registering databases.");
+
+	if(!startupDatabaseUpgrades())
+		return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_UPGRADING_DATABASES, "Problem with Acquaman startup: upgrading database.");
 
 	// Now that we have a database: populate initial settings, or just load user settings
 	if(isFirstTimeRun_) {
@@ -429,10 +431,6 @@ bool AMDatamanAppController::startupOnFirstTime()
 		if(!startupCreateDatabases())
 			return false;
 
-		// check for and run any database upgrades we require...
-		if(!startupDatabaseUpgrades())
-			return false;
-
 		if(userLocalStorage){
 			QDir userDataFolder(AMUserSettings::userDataFolder);
 			QStringList dataFolderFilters;
@@ -485,10 +483,6 @@ bool AMDatamanAppController::startupOnEveryTime()
 		return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_DATA_DIR_BACKUP_ERROR, "Problem with Acquaman startup: backing up data directory.");
 
 	if(!startupCreateDatabases())
-		return false;
-
-	// check for and run any database upgrades we require...
-	if(!startupDatabaseUpgrades())
 		return false;
 
 	if(usingLocalStorage()) {
--- a/source/application/AMDatamanAppController.cpp
+++ b/source/application/AMDatamanAppController.cpp
@@ -242,11 +242,11 @@ bool AMDatamanAppController::startup() {
 			return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_HANDING_NON_FIRST_TIME_USER, "Problem with Acquaman startup: handling non-first-time user.");
 	}
 
-	if(!startupRegisterDatabases())
-		return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_REGISTERING_DATABASES, "Problem with Acquaman startup: registering databases.");
-
 	if(!startupDatabaseUpgrades())
 		return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_UPGRADING_DATABASES, "Problem with Acquaman startup: upgrading database.");
+
+	if(!startupRegisterDatabases())
+		return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_REGISTERING_DATABASES, "Problem with Acquaman startup: registering databases.");
 
 	// Now that we have a database: populate initial settings, or just load user settings
 	if(isFirstTimeRun_) {
--- a/source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
+++ b/source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
@@ -2,6 +2,7 @@
 
 #include <QStringBuilder>
 
+#include "dataman/database/AMDbObjectSupport.h"
 #include "util/AMErrorMonitor.h"
 
 CLSDbUpgrade1Pt1::CLSDbUpgrade1Pt1(const QString &facilityName, const QString &databaseNameToUpgrade, QObject *parent)
@@ -24,15 +25,18 @@ bool CLSDbUpgrade1Pt1::upgradeNecessary() const
 {
 	// run update when:
 	//
-	//   - the AMRun table has the "facilityId" column
+	//   - The AMRun table has the "facilityId" column
 	//   - The AMFacility table has more than one record, or
-	//   - the name of the only facility doesn't match with the given beamline name, or
+	//   - The name of the only facility doesn't match with the given beamline name, or
 
+	//   - The AMRun table has the "facilityId" column
 	bool hasFacilityId = databaseToUpgrade_->tableExists("AMRun_table") && databaseToUpgrade_->columnExists("AMRun_table", "facilityId");
 
+	//   - The AMFacility table has more than one record, or
 	QVariantList tableFacilityNames = databaseToUpgrade_->retrieve("AMFacility_table", "name");
 	bool singleFacility = (tableFacilityNames.count() == 1);
 
+	//   - The name of the only facility doesn't match with the given beamline name, or
 	bool correctFacilityName = false;
 	if (singleFacility)
 		correctFacilityName = (tableFacilityNames.at(0).toString() == targetFacilityName_);
@@ -62,21 +66,21 @@ bool CLSDbUpgrade1Pt1::upgradeImplementation()
 		for (int rowId = tableFacilityNames.count(); rowId > 0; rowId--) {
 			dbResult = databaseToUpgrade_->deleteRow(rowId, "AMFacility_table");
 			if ( dbResult == 0 ){
-				AMErrorMon::alert(this, CLSDBUPGRADE1PT1_COULD_NOT_DELETE_FACILITY, QString("Could not delete the facility (%1) in AMFacility_table").arg(rowId));
+				AMErrorMon::alert(this, CLSDBUPGRADE1PT1_COULD_NOT_DELETE_FACILITY, QString("Could not delete the facility (%1) in %2").arg(rowId).arg("AMFacility_table"));
 				databaseToUpgrade_->rollbackTransaction();
 				return false;
 			}
 		}
 
-		// add the new facility
-		QStringList columns = QStringList() << "AMDbObjectType" << "thumbnailCount" << "thumbnailFirstid" << "name" << "description" << "iconFileName";
-		QVariantList values = QVariantList() << "AMFacility" << 1 << 1 << targetFacilityName_ <<  QString("CLS %1 Beamline").arg(targetFacilityName_) << ":/clsIcon.png";
-		dbResult = databaseToUpgrade_->insertOrUpdate(0, "AMFacility_table", columns, values);
-		if (dbResult == 0) {
-			AMErrorMon::alert(this, CLSDBUPGRADE1PT1_COULD_NOT_INSERT_NEW_FACILITY, QString("Could not insert the facility (%1) in AMFacility_table").arg(targetFacilityName_));
+		// add the new facility and create the thumbnail
+		AMFacility *facility = new AMFacility(targetFacilityName_, QString("CLS %1 Beamline").arg(targetFacilityName_), ":/clsIcon.png");
+		if (!facility->storeToDb(databaseToUpgrade_)) { // this will create Thumbnail as well
+			AMErrorMon::alert(this, CLSDBUPGRADE1PT1_COULD_NOT_INSERT_NEW_FACILITY, QString("Could not insert the facility (%1) in %2").arg(targetFacilityName_).arg("AMFacility_table"));
 			databaseToUpgrade_->rollbackTransaction();
+			facility->deleteLater();
 			return false;
 		}
+		facility->deleteLater();
 	}
 
 	// Stage 2:  update the AMRun table by deleting the facilityId column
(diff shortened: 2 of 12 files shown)


XuetaoLiu-CLS:
@helfrij  can you review this? Thx

source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
```diff
@@ -62,29 +66,29 @@ bool CLSDbUpgrade1Pt1::upgradeImplementation()
 		for (int rowId = tableFacilityNames.count(); rowId > 0; rowId--) {
 			dbResult = databaseToUpgrade_->deleteRow(rowId, "AMFacility_table");
 			if ( dbResult == 0 ){
-				AMErrorMon::alert(this, CLSDBUPGRADE1PT1_COULD_NOT_DELETE_FACILITY, QString("Could not delete the facility (%1) in AMFacility_table").arg(rowId));
+				AMErrorMon::alert(this, CLSDBUPGRADE1PT1_COULD_NOT_DELETE_FACILITY, QString("Could not delete the facility (%1) in %2").arg(rowId).arg("AMFacility_table"));
```
helfrij: I think there might need to be another '%'  in this QString.

helfrij: Oh no wait, it's just here in white text and hard to see. 👍

helfrij:
Looks good, @XuetaoLiu-CLS!

Time Estimate: 10 minutes.

source/application/AMDatamanAppController.cpp
```diff
@@ -242,10 +242,12 @@ bool AMDatamanAppController::startup() {
 			return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_HANDING_NON_FIRST_TIME_USER, "Problem with Acquaman startup: handling non-first-time user.");
 	}
 
-
 	if(!startupRegisterDatabases())
 		return AMErrorMon::errorAndReturn(this, AMDATAMANAPPCONTROLLER_STARTUP_ERROR_REGISTERING_DATABASES, "Problem with Acquaman startup: registering databases.");
 
+	if(!startupDatabaseUpgrades())
```
dretrex: I'm assuming placing this method here covers both first time and every time cases?  Either way, it looks a little weird to be called here rather than after `startupCreateDatabases()`.

source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
```diff
@@ -62,29 +66,29 @@ bool CLSDbUpgrade1Pt1::upgradeImplementation()
...
 
-		// add the new facility
-		QStringList columns = QStringList() << "AMDbObjectType" << "thumbnailCount" << "thumbnailFirstid" << "name" << "description" << "iconFileName";
-		QVariantList values = QVariantList() << "AMFacility" << 1 << 1 << targetFacilityName_ <<  QString("CLS %1 Beamline").arg(targetFacilityName_) << ":/clsIcon.png";
-		dbResult = databaseToUpgrade_->insertOrUpdate(0, "AMFacility_table", columns, values);
-		if (dbResult == 0) {
-			AMErrorMon::alert(this, CLSDBUPGRADE1PT1_COULD_NOT_INSERT_NEW_FACILITY, QString("Could not insert the facility (%1) in AMFacility_table").arg(targetFacilityName_));
+		// add the new facility and create the thumbnail
+		AMFacility *facility = new AMFacility(targetFacilityName_, QString("CLS %1 Beamline").arg(targetFacilityName_), ":/clsIcon.png");
+		if (!facility->storeToDb(databaseToUpgrade_)) { // this will create Thumbnail as well
```
dretrex: Have you tested this?  In general, you can't use AMDbObject inside of AMDbUpgrades because the upgrade has to be done before all the mappings and relations are established.  You should always need to do the inserts and updates by hand (unfortunately).

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -0,0 +1,133 @@
...
+	//   - The AMDbObjectThumbnails_table doesn't have thumbnail for the given beamline name or has more than one thumbnails for AMFacilityTable
+
+	QString thumbnailTableName = AMDbObjectSupport::thumbnailTableName();
+	QMap<int, QString> facilityTableThumbnails;
+
+	QSqlQuery thumbnailTableQuery = databaseToUpgrade_->select(thumbnailTableName, "id, title", QString("objectTableName=%1").arg("AMFacility_table") );
+	if (thumbnailTableQuery.exec()) {
+		while (thumbnailTableQuery.next()) {
+			facilityTableThumbnails.insert(thumbnailTableQuery.value(0).toInt(), thumbnailTableQuery.value(1).toString()); //id, title
+//			qDebug() << "==== " << thumbnailTableQuery.value(0).toInt() << thumbnailTableQuery.value(1).toString();
```
dretrex: There are some debugs that got missed here.

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -0,0 +1,133 @@
...
+	////////////////////////////////////////////////
+
+	databaseToUpgrade_->startTransaction();
+
+	QString thumbnailTableName = AMDbObjectSupport::thumbnailTableName();
+
+	// Stage 1: update the thumbnail table if required
+	//   - The AMDbObjectThumbnails_table doesn't have thumbnail for the given beamline name
+	QMap<int, QString> facilityTableThumbnails;
+	QSqlQuery thumbnailTableQuery = databaseToUpgrade_->select(thumbnailTableName, "id, title", QString("objectTableName=%1").arg("AMFacility_table") );
```
dretrex: This select statement doesn't match `AMDatabase::select()` declaration and there is no QSqlDatabase version either.

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -0,0 +1,133 @@
...
+	if (thumbnailTableQuery.exec()) {
+		while (thumbnailTableQuery.next()) {
+			facilityTableThumbnails.insert(thumbnailTableQuery.value(0).toInt(), thumbnailTableQuery.value(1).toString()); //id, title
+		}
+	}
+	thumbnailTableQuery.finish();
+
+	bool correctThumbnailTableData = (facilityTableThumbnails.count() == 1) && (facilityTableThumbnails.values().at(0) == targetFacilityName_);
```
dretrex: Are we only expecting one here or is there an implicit assumption about the fact that only asking the first value is necessary?

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -0,0 +1,133 @@
...
+			// delete the existing facility and re-add it along with thumbnail
+			dbResult = databaseToUpgrade_->deleteRow(1, "AMFacility_table");
+			if ( dbResult == 0 ){
+				AMErrorMon::alert(this, CLSDbUpgrade1Pt2_COULD_NOT_DELETE_FACILITY, QString("Could not delete the facility in %1").arg("AMFacility_table"));
+							databaseToUpgrade_->rollbackTransaction();
+							return false;
+			}
+
+			// add the new facility and create a new thumbnail
+			AMFacility *facility = new AMFacility(targetFacilityName_, QString("CLS %1 Beamline").arg(targetFacilityName_), ":/clsIcon.png");
```
dretrex: This is unsafe for the same reason as mentioned above.

dretrex:
I'm not confident this code works based on how the relationship of AMDbUpgrades work with respect to AMDatabase and AMDbObject.

Time Estimate: 15 minutes

source/application/AMDatamanAppController.cpp
XuetaoLiu-CLS: - yes, it covers both first time and every time cases.
- As you pointed below as well, the reason I put this is after `startUpRegisterDatabases()` is that I want to use method `if (!facility->storeToDb(databaseToUpgrade_))`, which can save the thumbnail at the same time.

source/dataman/CLS/CLSDbUpgrade1Pt1.cpp
XuetaoLiu-CLS: @dretrex 
I tested this, and it works, since I moved the `startupDatabaseUpgrades()` after the `startupRegisterDatabases()`.

yes, you are right. Being lazy to do the `inserts and updates by hand (unfortunately)`, I tried to call this function and  made the changes to AMDatamanAppController.

I can change to the way we usually do for sure.

For sure I can change it to the way we us

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
XuetaoLiu-CLS: We are expecting only one for the facility table thumbnail, and the the first value should match with the target name

XuetaoLiu-CLS:
@dretrex I rewrite the CLSDbUpgrade1Pt2, can you take a look? Thx

dretrex:
Sure thing.

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -27,84 +28,132 @@ bool CLSDbUpgrade1Pt2::upgradeNecessary() const
...
+	foreach (int id, facilityTableThumbnails.keys()) {
+		if (facilityTableThumbnails.value(id) != targetFacilityName_) {
+			dbResult = databaseToUpgrade_->deleteRow(id, thumbnailTableName_);
+			if ( dbResult == 0 ){
+				AMErrorMon::alert(this, CLSDbUpgrade1Pt2_FAIL_TO_DELETE_FACILITY_THUMBNAIL, QString("Could not delete the thumbnail (%1) in %2").arg(id).arg(thumbnailTableName_));
+				databaseToUpgrade_->rollbackTransaction();
+				return false;
+			}
+
+			facilityTableThumbnails.remove(id);
```
dretrex: I find that removing elements from a list that is being used in the for loop arguments can lead to some unpredictable behaviour.  I would suggest something like:

```c++
QList<int> idsToRemove;
foreach(){
    if(){
        // stuff
        idsToRemove << id;
    }
}
foreach(id, idsToRemove)
    facilityTableThumbnails.remove(id);
```

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -27,84 +28,132 @@ bool CLSDbUpgrade1Pt2::upgradeNecessary() const
...
+		int facilityId = -1;
+		AMFacility *facility = 0;
+
+		QSqlQuery queryFacility = databaseToUpgrade_->query();
+		queryFacility.prepare(QString("SELECT id, iconFileName FROM %1 where name='%2'").arg(facilityTableName_).arg(targetFacilityName_));
+
+		if (queryFacility.exec()) {
+
+			if (queryFacility.first()) {
+				facility = new AMFacility(targetFacilityName_, QString("CLS %1 Beamline").arg(targetFacilityName_), ":/clsIcon.png");
```
dretrex: What is the purpose of this?  Since you don't use `facility->storeToDb()` (and rightly so) this doesn't give you anything.  I'm also reluctant to have AMDbObjects instantiated since it is very tied to the AMDatabase system.  It just seems like one more place that something can go wrong.

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -131,3 +180,26 @@ QString CLSDbUpgrade1Pt2::description() const
...
+		while (query.next()) {
+			facilityTableThumbnails.insert(query.value(0).toInt(), query.value(1).toString());
+		}
+	} else {
+
+		AMErrorMon::alert(this, CLSDbUpgrade1Pt2_FAIL_TO_QUERY_FACILITY_THUMBNAIL, QString("%1").arg(query.lastQuery()));
+		AMErrorMon::alert(this, CLSDbUpgrade1Pt2_FAIL_TO_QUERY_FACILITY_THUMBNAIL, QString("Query to find objectTableName (%1) in table %2 is failed.").arg(facilityTableName_).arg(thumbnailTableName_));
+	}
+
+//	query.finish();
```
dretrex: Isn't this necessary?

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
XuetaoLiu-CLS: I want to use it for the thumbnail() function to generate the Thumbnail.

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
XuetaoLiu-CLS: I forgot to remove the commented code.

XuetaoLiu-CLS: I think I removed it. Maybe I forgot to push the last commit I made. I will double check it

dretrex: What I meant was I think that commented line is necessary.  You should always call query.finish() if you are done with an QSqlQuery.

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
dretrex: Just use 

```c++
facilityImage = QImage(queryFacility.value(1).toString());
QBuffer imageBuffer;
bout.open(QIODevice::WriteOnly);
if(image.save(&imageBuffer, "PNG")) 
	thumbnail = imageBuffer.buffer();

.......
values << thumbnail;
```

Like I said, I would generally feel better if we don't use or instantiate any AMDbObjects.  There are aspects to that system that are linked together and can create unexpected behaviour.  I am quite certain that this won't have unintended consequences, but I don't want us to get in the habit of using them because database corruption is a very real possibility - especially for our delicate object/table relationships.

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
XuetaoLiu-CLS: @dretrex Good catch.

XuetaoLiu-CLS:
@dretrex you can re-review this now. It is changed and tested

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
```diff
@@ -193,13 +158,9 @@ QMap<int, QString> CLSDbUpgrade1Pt2::queryAMFacilityTableThumbnails() const
 			facilityTableThumbnails.insert(query.value(0).toInt(), query.value(1).toString());
 		}
 	} else {
-
-		AMErrorMon::alert(this, CLSDbUpgrade1Pt2_FAIL_TO_QUERY_FACILITY_THUMBNAIL, QString("%1").arg(query.lastQuery()));
 		AMErrorMon::alert(this, CLSDbUpgrade1Pt2_FAIL_TO_QUERY_FACILITY_THUMBNAIL, QString("Query to find objectTableName (%1) in table %2 is failed.").arg(facilityTableName_).arg(thumbnailTableName_));
 	}
 
-//	query.finish();
```
dretrex: Rather than removing this, wouldn't this be better to have?

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
dretrex: Turns out I was being over-cautious.  From Qt docs:

> Qt automatically takes a copy of the container when it enters a foreach loop. If you modify the container as you are iterating, that won't affect the loop. (If you do not modify the container, the copy still takes place, but thanks to implicit sharing copying a container is very fast.)

source/dataman/CLS/CLSDbUpgrade1Pt2.cpp
XuetaoLiu-CLS: @dretrex we don't need to call this, as mentioned in the QtDoc:
```
Instruct the database driver that no more data will be fetched from this query until it is re-executed. There is normally no need to call this function, but it may be helpful in order to free resources such as locks or cursors if you intend to re-use the query at a later time.

Sets the query to inactive. Bound values retain their values.
```

dretrex: Ah alright.

dretrex:
This is done now, right @XuetaoLiu-CLS?

XuetaoLiu-CLS:
@dretrex yes, it is done.

PR #2410, opened 2016-06-17
PullRequest for Issue2161: save userConfiguration data to database before shutdown

XuetaoLiu-CLS:
- add userConfiguration to CLSAppController
- save userConfigfiguration to the user database before shutdown
- rename the beamline userConfiguration to beamline specific name

### DIFF
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -11,7 +11,7 @@ BioXASAppController::BioXASAppController(const QString &beamlineName, QObject *p
 {
 	// Initialize controller settings.
 
-	userConfiguration_ = new BioXASUserConfiguration(this);
+	bioxasUserConfiguration_ = new BioXASUserConfiguration(this);
 	setDefaultUseLocalStorage(true);
 
 	// Database upgrades.
@@ -129,7 +129,7 @@ void BioXASAppController::cleanMoveImmediatelyAction()
 
 void BioXASAppController::onUserConfigurationLoadedFromDb()
 {
-	if (userConfiguration_) {
+	if (bioxasUserConfiguration_) {
 
 		AMDetectorSet *geDetectors = BioXASBeamline::bioXAS()->ge32ElementDetectors();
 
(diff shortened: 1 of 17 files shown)


XuetaoLiu-CLS:
@helfrij can you reivew this? Thx

source/application/CLS/CLSAppController.cpp
```diff
@@ -60,6 +61,14 @@ bool CLSAppController::startup()
 	return false;
 }
 
+void CLSAppController::shutdown()
+{
+	if (userConfiguration_)
```
helfrij: It seems like the beamline-specific user configurations are all named something different (eg. bioxasUserConfiguration_, ideasUserConfiguration_, etc). Is there a case where a user configuration is just called userConfiguration_? It seems like this would be the only case where this code is executed.

helfrij:
A question about whether the beamline-specific user configurations are saved on shutdown.

Time Estimate: 10 minutes.

source/application/CLS/CLSAppController.cpp
XuetaoLiu-CLS: @helfrij yes. #2411 will handle the code refactor to generalize this

helfrij: We could implement a user configuration setter for the app controller at the CLS-level. Each beamline could set their particular configuration that way.

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -352,15 +354,17 @@ void BioXASAppController::setupScanConfigurations()
...
 
-		bool loaded = userConfiguration_->loadFromDb(AMDatabase::database("user"), 1);
+		bool loaded = bioxasUserConfiguration_->loadFromDb(AMDatabase::database("user"), 1);
 		if (!loaded) {
-			userConfiguration_->storeToDb(AMDatabase::database("user"));
+			bioxasUserConfiguration_->storeToDb(AMDatabase::database("user"));
 			onUserConfigurationLoadedFromDb();
 		}
+
+		userConfiguration_ = bioxasUserConfiguration_;
```
dretrex: Is setting this variable going to be handled by the refactor as well?  Seems like an easy place for something to get missed and then having random bugs crop up.

dretrex:
One question, otherwise looks alright.
Time Estimate: 5 minutes

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: @dretrex yes, in the other refactor, there will be no requirements for this assignment.

Issue #2411, opened 2016-06-17
code refactor: create generic userConfiguration in CLS level 


XuetaoLiu-CLS:
when working on #2161, #2410, I noticed that the beamline userConfigurations are very alike. Maybe it is possible to create a `CLSUserConfiguration` as a generic userConfiguration.

XuetaoLiu-CLS:
Time Estimate: 4 hours

Issue #2252, opened 2016-04-20
sometime the workflow will stuck


XuetaoLiu-CLS:
This happened twice in SXRMB:
The current scan of a workflow is done (finished the last point and the scan is exported). But it won't run the next scan. Acquaman still shows the current scan is running.

XuetaoLiu-CLS:
@dretrex 
I guess I know what happened this now, as a side product of #1628 .

The queue will be stuck when doing the following operations:
1) pause the scan action from the scan action view, (this will pause the scan action runner)
   (at this moment, however, the workflow doesn't know the scan action runner is paused, and still counting down. This is the real bug we should fix).
2) pause the work flow, either on the workflow view or on the main bottom bar
3) resume the scan from the scan action view, which will resume the scan action runner
4) scan action runner will run all the remaining scans and finish. But the workflow will stuck there for ever and we can't even cancel it since the current scan action is NULL already (the scan action is finished).

---- 

The fixes:
1) when scan action runner is paused/resumed, it should notify the workflow to pause/resume as well (as the workflow did. When we pause/resume workflow, the scan action runner will pause/resume as well)
3) maybe in workflow, add a protection. When the scan action is finished (exported), it will try to clean the current action (??? not sure whether it is proper to do things like this or not)

XuetaoLiu-CLS:
Time Estimate: 4 days

PR #2414, opened 2016-06-21
PullRequest for Issue2411: generalize the userconfiguration classes

XuetaoLiu-CLS:
#2411

### DIFF
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -11,7 +11,7 @@ BioXASAppController::BioXASAppController(const QString &beamlineName, QObject *p
 {
 	// Initialize controller settings.
 
-	userConfiguration_ = new BioXASUserConfiguration(this);
+	bioxasUserConfiguration_ = new BioXASUserConfiguration(this);
 	setDefaultUseLocalStorage(true);
 
 	// Database upgrades.
@@ -129,7 +129,7 @@ void BioXASAppController::cleanMoveImmediatelyAction()
 
 void BioXASAppController::onUserConfigurationLoadedFromDb()
 {
-	if (userConfiguration_) {
+	if (bioxasUserConfiguration_) {
 
 		AMDetectorSet *geDetectors = BioXASBeamline::bioXAS()->ge32ElementDetectors();
 
@@ -139,7 +139,7 @@ void BioXASAppController::onUserConfigurationLoadedFromDb()
 
 			if (geDetector) {
 
-				foreach (AMRegionOfInterest *region, userConfiguration_->regionsOfInterest()){
+				foreach (AMRegionOfInterest *region, bioxasUserConfiguration_->regionsOfInterest()){
 					if (!containsRegionOfInterest(geDetector->regionsOfInterest(), region)) {
 						geDetector->addRegionOfInterest(region);
 						onRegionOfInterestAdded(region);
@@ -154,14 +154,17 @@ void BioXASAppController::onUserConfigurationLoadedFromDb()
 	}
 }
 
+#include <QDebug>
 void BioXASAppController::onRegionOfInterestAdded(AMRegionOfInterest *region)
 {
 	if (region) {
 
 		// Add the region of interest to the user configuration, if it doesn't have it already.
 
-		if (userConfiguration_ && !containsRegionOfInterest(userConfiguration_->regionsOfInterest(), region))
-			userConfiguration_->addRegionOfInterest(region);
+		if (bioxasUserConfiguration_ && !containsRegionOfInterest(bioxasUserConfiguration_->regionsOfInterest(), region)) {
+			qDebug() << "==== add ROI to userConfiguration " << region->name() ;
+			bioxasUserConfiguration_->addRegionOfInterest(region);
+		}
 
 		// Add the region of interest to the XAS scan configuration, if it doesn't have it already.
 
@@ -179,8 +182,8 @@ void BioXASAppController::onRegionOfInterestRemoved(AMRegionOfInterest *region)
 {
 	// Remove region of interest from the user configuration.
 
-	if (userConfiguration_)
-		userConfiguration_->removeRegionOfInterest(region);
+	if (bioxasUserConfiguration_)
+		bioxasUserConfiguration_->removeRegionOfInterest(region);
 
 	// Remove region of interest from the XAS scan configuration.
 
@@ -197,8 +200,8 @@ void BioXASAppController::onRegionOfInterestBoundingRangeChanged(AMRegionOfInter
 {
 	// Update the bounding range for the region of interest in the user configuration.
 
-	if (userConfiguration_)
-		userConfiguration_->setRegionOfInterestBoundingRange(region);
+	if (bioxasUserConfiguration_)
+		bioxasUserConfiguration_->setRegionOfInterestBoundingRange(region);
 
 	// Update the bounding range for the region of interest in the XAS scan configuration.
 
@@ -255,16 +258,16 @@ void BioXASAppController::onCurrentScanActionStartedImplementation(AMScanAction 
 
 	// Save current configuration to the database.
 
-	if (userConfiguration_)
-		userConfiguration_->storeToDb(AMDatabase::database("user"));
+	if (bioxasUserConfiguration_)
+		bioxasUserConfiguration_->storeToDb(AMDatabase::database("user"));
 }
 
 void BioXASAppController::onCurrentScanActionFinishedImplementation(AMScanAction *action)
 {
 	// Save current user configuration to the database.
 
-	if (userConfiguration_)
-		userConfiguration_->storeToDb(AMDatabase::database("user"));
+	if (bioxasUserConfiguration_)
+		bioxasUserConfiguration_->storeToDb(AMDatabase::database("user"));
 
 	// If the scan was an energy calibration scan, set the calibration view's scan and make it the current pane.
 
@@ -352,15 +355,17 @@ void BioXASAppController::setupScanConfigurations()
 
 void BioXASAppController::setupUserConfiguration()
 {
-	if (userConfiguration_) {
+	if (bioxasUserConfiguration_) {
 
-		connect( userConfiguration_, SIGNAL(loadedFromDb()), this, SLOT(onUserConfigurationLoadedFromDb()) );
+		connect( bioxasUserConfiguration_, SIGNAL(loadedFromDb()), this, SLOT(onUserConfigurationLoadedFromDb()) );
 
-		bool loaded = userConfiguration_->loadFromDb(AMDatabase::database("user"), 1);
+		bool loaded = bioxasUserConfiguration_->loadFromDb(AMDatabase::database("user"), 1);
 		if (!loaded) {
-			userConfiguration_->storeToDb(AMDatabase::database("user"));
+			bioxasUserConfiguration_->storeToDb(AMDatabase::database("user"));
 			onUserConfigurationLoadedFromDb();
 		}
+
+		userConfiguration_ = bioxasUserConfiguration_;
 	}
 }
 
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -128,34 +128,6 @@ void BioXASAppController::cleanMoveImmediatelyAction()
 	moveImmediatelyAction_ = 0;
 }
 
-void BioXASAppController::onUserConfigurationLoadedFromDb()
-{
-	if (userConfiguration_) {
-
-		AMDetectorSet *geDetectors = BioXASBeamline::bioXAS()->ge32ElementDetectors();
-
-		for (int i = 0; i < geDetectors->count(); i++) {
-
-			AMXRFDetector *geDetector = qobject_cast<AMXRFDetector*>(geDetectors->at(i));
-
-			if (geDetector) {
-
-				foreach (AMRegionOfInterest *region, userConfiguration_->regionsOfInterest()){
-					if (!containsRegionOfInterest(geDetector->regionsOfInterest(), region)) {
-						geDetector->addRegionOfInterest(region);
-						onRegionOfInterestAdded(region);
-					}
-				}
-
-				connect(geDetector, SIGNAL(addedRegionOfInterest(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestAdded(AMRegionOfInterest*)));
-				connect(geDetector, SIGNAL(removedRegionOfInterest(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestRemoved(AMRegionOfInterest*)));
-				connect(geDetector, SIGNAL(regionOfInterestBoundingRangeChanged(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestBoundingRangeChanged(AMRegionOfInterest*)));
-			}
-		}
-
-	}
-}
-
 void BioXASAppController::onRegionOfInterestAdded(AMRegionOfInterest *region)
 {
 	if (region) {
--- a/source/application/BioXAS/BioXASAppController.cpp
+++ b/source/application/BioXAS/BioXASAppController.cpp
@@ -151,10 +151,10 @@ void BioXASAppController::onUserConfigurationLoadedFromDb()
 				connect(geDetector, SIGNAL(regionOfInterestBoundingRangeChanged(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestBoundingRangeChanged(AMRegionOfInterest*)));
 			}
 		}
+
 	}
 }
 
-#include <QDebug>
 void BioXASAppController::onRegionOfInterestAdded(AMRegionOfInterest *region)
 {
 	if (region) {
@@ -162,7 +162,6 @@ void BioXASAppController::onRegionOfInterestAdded(AMRegionOfInterest *region)
 		// Add the region of interest to the user configuration, if it doesn't have it already.
 
 		if (bioxasUserConfiguration_ && !containsRegionOfInterest(bioxasUserConfiguration_->regionsOfInterest(), region)) {
-			qDebug() << "==== add ROI to userConfiguration " << region->name() ;
 			bioxasUserConfiguration_->addRegionOfInterest(region);
 		}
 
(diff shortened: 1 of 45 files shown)


XuetaoLiu-CLS:
Testing:
- [ ] SXRMB
- [ ] BioXAS
- [ ] VESPERS
- [ ] IDEAS
- [ ] REIXS
- [ ] PGM

XuetaoLiu-CLS:
@helfrij can you review? Thx

source/application/CLS/CLSAppController.cpp
```diff
@@ -83,6 +85,24 @@ void CLSAppController::shutdown()
 }
 
 // ============== implementation of protected slots =====================
+void CLSAppController::onUserConfigurationLoadedFromDb()
+{
+	AMXRFDetector *xrfDetector = CLSBeamline::clsBeamline()->xrfDetector();
```
helfrij: What happens if there are multiple XRF detectors on the beamline?

source/application/IDEAS/IDEASAppController.cpp
```diff
@@ -356,48 +319,37 @@ void IDEASAppController::configureSingleSpectrumView(AMGenericScanEditor *editor
...
-	connect(IDEASBeamline::ideas()->ketek(), SIGNAL(addedRegionOfInterest(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestAdded(AMRegionOfInterest*)));
-	connect(IDEASBeamline::ideas()->ketek(), SIGNAL(removedRegionOfInterest(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestRemoved(AMRegionOfInterest*)));
-	connect(IDEASBeamline::ideas()->ketek(), SIGNAL(regionOfInterestBoundingRangeChanged(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestBoundingRangeChanged(AMRegionOfInterest*)));
 }
 
 void IDEASAppController::onRegionOfInterestAdded(AMRegionOfInterest *region)
 {
-	ideasUserConfiguration_->addRegionOfInterest(region);
+	userConfiguration_->addRegionOfInterest(region);
 	mapScanConfiguration_->addRegionOfInterest(region);
```
helfrij: The user configuration is now generalized at the CLS-level. Is it possible to generalize updating the user configuration and the scan configurations in response to changes in the regions of interest by having a CLS-level list of configurations?

source/application/PGM/PGMAppController.cpp
```diff
@@ -53,40 +54,20 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
...
-				oceanOpticsDetector->addRegionOfInterest(region);
-			}
-		}
-	}
+	detectorPaneCategoryName_ = "XRF Detectors";
 }
 
 void PGMAppController::onRegionOfInterestAdded(AMRegionOfInterest *region)
 {
-	if (region) {
```
helfrij: Do we need this check for whether the region is valid?

source/application/REIXS/REIXSAppController.cpp
```diff
@@ -177,11 +160,6 @@ void REIXSAppController::setupScanConfigurations()
 	genericScanConfiguration_ = new AMGenericStepScanConfiguration();
 }
 
-void REIXSAppController::setupUserConfiguration()
```
helfrij: Did the setup for the REIXS user configuration happen elsewhere? Is it the same as the new CLS-level setup?

source/beamline/BioXAS/BioXASBeamline.cpp
```diff
@@ -338,6 +338,19 @@ CLSValves* BioXASBeamline::valves() const
 	return result;
 }
 
+AMXRFDetector *BioXASBeamline::xrfDetector() const
```
helfrij: This implementation works for now, as there is only one Ge detector on Side and Main. But that won't always be the case. I'm not sure this implementation makes sense for this future case.

source/beamline/BioXAS/BioXASMainBeamline.h
```diff
@@ -47,7 +47,7 @@ class BioXASMainBeamline : public BioXASBeamline
 	{
 		if(!instance_){
 			instance_ = new BioXASMainBeamline();
-			instance_->initializeBeamlineSupport();
+			instance_->initializeBeamline();
```
helfrij: What is the reason for the name change? To me, 'initializing' the beamline sounds like what happens when we create all the beamline components.

source/beamline/CLS/CLSBeamline.h
```diff
@@ -27,13 +27,33 @@ class CLSBeamline : public AMBeamline
...
+protected slots:
+	/// Slot to handle the signals which will affect the beam availability
+	void updateBeamAvailabilityStatus();
 
 protected:
 	/// Protected constructor, for singleton pattern.
 	CLSBeamline(const QString &controlName);
 
+	/// helper function to check whether the beam of a beamline is available or not --- this usually is combined with beamline status PV and/or beamline shutters/valves stauts
+	virtual bool isBeamlineBeamAvailable() = 0;
```
helfrij: It doesn't look like BioXAS has a real implementation for this, `BioXASBeamline::isBeamlineBeamAvailable()` always returns true. If that's the desired default implementation, maybe that could be implemented here?

source/beamline/IDEAS/IDEASBeamline.cpp
```diff
@@ -296,6 +296,16 @@ void IDEASBeamline::onShutterValueChanged()
 	emit overallShutterStatus(safetyShutter_->isOpen() && photonShutter2_->isOpen() && safetyShutter2_->isOpen());
 }
 
+AMXRFDetector *IDEASBeamline::xrfDetector() const
+{
+	if (ketek_ )
```
helfrij: Just checking. It looks like this method will preferentially return the Ketek detector. Is there ever a case where there would be instances of both detectors, but the ge13Element is the desired detector?

source/beamline/CLS/CLSBeamline.h
helfrij: Same thing with IDEAS.

source/beamline/SXRMB/SXRMBBeamline.cpp
```diff
@@ -439,14 +438,14 @@ AMBasicControlDetectorEmulator* SXRMBBeamline::energyFeedbackDetector() const
 	return energyFeedbackDetector_;
 }
 
-SXRMBBrukerDetector* SXRMBBeamline::brukerDetector() const
+AMXRFDetector *SXRMBBeamline::xrfDetector() const
 {
-	return brukerDetector_;
-}
+	if (brukerDetector_)
```
helfrij: Same questions about returning the appropriate detector here.

source/beamline/VESPERS/VESPERSBeamline.h
```diff
@@ -530,8 +532,8 @@ class VESPERSBeamline : public CLSBeamline
 signals:
 	/// Notifier that the beam has been changed.
 	void currentBeamChanged(VESPERS::Beam);
-	/// Notifier that passes on that the beam has gone down.
-	void beamDumped();
+//	/// Notifier that passes on that the beam has gone down.
```
helfrij: This signal is commented out for now. Do we still need it?

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -127,30 +128,6 @@ void BioXASAppController::cleanMoveImmediatelyAction()
 	moveImmediatelyAction_ = 0;
 }
 
-void BioXASAppController::onUserConfigurationLoadedFromDb()
```
helfrij: Hmmm. The BioXAS implementation differs from the default implementation in CLSAppController by handling multiple detectors. Could this slot be virtual? That way different applications can reimplement as necessary.

source/application/BioXAS/BioXASAppController.cpp
```diff
@@ -242,30 +219,18 @@ void BioXASAppController::goToEnergyCalibrationView(AMScan *toView)
...
-		bioxasUserConfiguration_->storeToDb(AMDatabase::database("user"));
-
 	// If the scan was an energy calibration scan, set the calibration view's scan and make it the current pane.
 
 	if (action && action->controller() && action->controller()->scan()) {
 		BioXASXASScanConfiguration *xasScanConfiguration = qobject_cast<BioXASXASScanConfiguration*>(action->controller()->scan()->scanConfiguration());
 		if (xasScanConfiguration && xasScanConfiguration->name() == "Energy Calibration XAS Scan")
 			goToEnergyCalibrationView(action->controller()->scan());
 	}
+
```
helfrij: Just checking. Does execution order matter here? Previously, the store to db happened before any of the view changes.

source/application/BioXAS/BioXASImagingAppController.cpp
```diff
@@ -151,12 +145,8 @@ void BioXASImagingAppController::createComponentsPane()
...
-void BioXASImagingAppController::onCurrentScanActionStartedImplementation(AMScanAction *action)
-{
-	Q_UNUSED(action)
-}
-
 void BioXASImagingAppController::onCurrentScanActionFinishedImplementation(AMScanAction *action)
 {
-	Q_UNUSED(action)
+	// perfrom the general actions when scan action finished
+	CLSAppController::onCurrentScanActionFinishedImplementation(action);
```
helfrij: I'm not sure that Imaging has a user configuration yet. Will executing this method cause a crash?

source/application/CLS/CLSAppController.cpp
```diff
@@ -92,7 +112,38 @@ void CLSAppController::onScanEditorCreatedImplementation(AMGenericScanEditor *ed
...
+void CLSAppController::onRegionOfInterestBoundingRangeChanged(AMRegionOfInterest *region)
+{
+	Q_UNUSED(region)
+
+	AMErrorMon::debug(this, CLS_APPCONTROLLER_INFO_UNIMPLEMENTED_METHOD, "Looks like there is no beamline implementation for onRegionOfInterestBoundingRangeChanged(). ");
+}
+
+void CLSAppController::onBeamAvailabilityChanged(bool beamAvailable)
+{
+	if (!beamAvailable && !AMActionRunner3::workflow()->pauseCurrentAction())
```
helfrij: I have no idea how the workflow system works! But wouldn't we want to pause a scan action runner instead of the workflow? What if something is being executed in the workflow that isn't a scan and doesn't need to be paused in a beam dump?

source/application/SGM/SGMAppController.cpp
```diff
@@ -147,38 +148,31 @@ void SGMAppController::resizeToMinimum()
...
+	userConfiguration_->removeRegionOfInterest(region);
 	xasScanConfiguration_->removeRegionOfInterest(region);
 	lineScanConfiguration_->removeRegionOfInterest(region);
 	mapScanConfiguration_->removeRegionOfInterest(region);
 }
 
+void SGMAppController::onRegionOfInterestBoundingRangeChanged(AMRegionOfInterest *region)
+{
+	userConfiguration_->setRegionOfInterestBoundingRange(region);
+//	xasScanConfiguration_->setRegionOfInterestBoundingRange(region);
```
helfrij: It looks like these configurations are commented out here, but not in other ROI related slots. Do we still need them?

source/application/VESPERS/VESPERSAppController.cpp
```diff
@@ -306,22 +305,6 @@ void VESPERSAppController::setupScanConfigurations()
...
 
-void VESPERSAppController::setupUserConfiguration()
-{
-	connect(vespersUserConfiguration_, SIGNAL(loadedFromDb()), this, SLOT(onUserConfigurationLoadedFromDb()));
-
-	if (!vespersUserConfiguration_->loadFromDb(AMDatabase::database("user"), 1)){
-
-		vespersUserConfiguration_->storeToDb(AMDatabase::database("user"));
-		// This is connected here because our standard way for these signal connections is to load from db first, which clearly won't happen on the first time.
-		connect(VESPERSBeamline::vespers()->vespersSingleElementVortexDetector(), SIGNAL(addedRegionOfInterest(AMRegionOfInterest*)), this, SLOT(onRegionOfInterestAdded(AMRegionOfInterest*)));
```
helfrij: A number of detectors are connected to here. This code has been removed, and I'm not sure I see where the same connections are being made elsewhere.

source/beamline/CLS/CLSBeamline.cpp
```diff
@@ -1,12 +1,40 @@
...
+
+	connect(CLSStorageRing::sr1(), SIGNAL(beamAvaliability(bool)), this, SLOT(updateBeamAvailabilityStatus()));
+
+	updateBeamAvailabilityStatus();
+}
+
+/**
+ * ================= implementation of protected slots / functions =======================
+ **/
+void CLSBeamline::updateBeamAvailabilityStatus()
```
helfrij: Just checking. It looks like this method should be called every time the storage ring beam availability changes and when the beamline beam availability changes. I see a connect statement above for the storage ring, but I'm not sure I see something for the beamline availability?

source/application/BioXAS/BioXASAppController.cpp
helfrij: Yup, I think it's virtual already. Nevermind!

helfrij:
A couple of questions. Cool changes, @XuetaoLiu-CLS!

Time Estimate: 1 hour.

source/application/CLS/CLSAppController.cpp
XuetaoLiu-CLS: in AMBeamline, we have A function `addSynchronizedXRFDetector()`, which can help synchronize the settings among the XRF detectors.

source/application/IDEAS/IDEASAppController.cpp
XuetaoLiu-CLS: The reason I didn't update `userConfiguration_` in the CLS-level is that there is no generalized way to update the scan configuration in CLS-level at this moment.

Maybe in the future we will do that

source/application/PGM/PGMAppController.cpp
XuetaoLiu-CLS: I don't think so. The region should be validated when we emit the signal.

source/application/REIXS/REIXSAppController.cpp
XuetaoLiu-CLS: yes, it is generalized in the CLSAppController. This function has the same implementation in all the AppControllers

source/beamline/BioXAS/BioXASBeamline.cpp
XuetaoLiu-CLS: I agree this might not work in the future. This function should return a default XRF detector. There is another overloaded function in the other beamlines (SXRMB, IDEAS, etc) `AMXRFDetector *xrfDetector(SXRMB::FluorescenceDetectors detectorType) const;`. This will return the XRFDetector with given type (the one configured).

XuetaoLiu-CLS: the overloaded version should be the one you need to use when you want a specific detector. This function might also be potentially generalized.

dretrex: VESPERS already has a configuration where multiple XRF detectors can be used at once.

source/beamline/CLS/CLSBeamline.h
XuetaoLiu-CLS: Good catch. I forgot to change it. IDEAS and REIXS still return true at this moment. Their overall beam availability is determined by the storage beam status at this moment. In the future, when they are adapting the CLSBeamStatus, we can generalize this in CLS level

source/beamline/IDEAS/IDEASBeamline.cpp
XuetaoLiu-CLS: As mentioned above, this function was designed return a default XRF detector. The specific XRF detector is return in another function:
`AMXRFDetector *IDEASBeamline::xrfDetector(IDEAS::FluorescenceDetectors detectorType)`

source/beamline/VESPERS/VESPERSBeamline.h
XuetaoLiu-CLS: We don't need it any more.

source/application/BioXAS/BioXASAppController.cpp
XuetaoLiu-CLS: I don't think it matters. The reason I did that is to adopt the ideas of constructor/destructor of O-O design"
```
Constructor order:
    super class
    sub calss

Destructor order:
    sub class
     super class
```
for this case, it is the actions we going to do when scan finished. So I put it in that order

source/application/BioXAS/BioXASImagingAppController.cpp
XuetaoLiu-CLS: It won't. We checked `userConfiguration_` before using it. I put it here just to make sure we won't forgot to call it in the future.

source/application/CLS/CLSAppController.cpp
XuetaoLiu-CLS: Yes, we were suppose to pause the scan action runner. As #2252, this will make the workflow action runner stuck. I use the workflow action runner here until we found a good solution for this.

source/application/SGM/SGMAppController.cpp
XuetaoLiu-CLS: Good question! First this is the SGM codes, which we won't care about it at this moment; second, SGM doesn't have the slot for the ROI bounding range changed. I don't know whether they want it or not. That's why I added those codes but commented off. 

At least in the future, if SGM is using Acuqman again, and if they have questions on the ROI bounding range changed, we know we thought about this.

source/application/VESPERS/VESPERSAppController.cpp
XuetaoLiu-CLS: it is in `void CLSAppController::onUserConfigurationLoadedFromDb()`.

source/beamline/CLS/CLSBeamline.cpp
XuetaoLiu-CLS: for BioXAS, PGM and SXRMB, it is called when the beamlineStatus_ value changed or connection changed; for VESPERS, it is called when the POE status changed.

source/beamline/BioXAS/BioXASMainBeamline.h
XuetaoLiu-CLS: in CLSBeamline, it is more `initializeBeamline()` now than just `initializebeamlineSupport()`.

PR #2428, opened 2016-06-28
Pull Request for Issue2346 - Adding wiggler gap to beam status

helfrij:
Created some new classes to test out new beam status-related design patterns.
- AMControlToolButton, inherits from AMToolButton and maps control properties to tool button color states.
- Created BioXASBeamStatus, general enumerated control that reports whether the beam is on/off depending on the provided components and component states.
- Created BioXASBeamStatusButtonBar, button view for the beam status where each component is represented by an AMControlToolButton.
- Created BioXASBeamStatusView, a view for the beam status.

The key differences between these classes and the existing CLSBeamlineStatus classes are:
- most beamline-/component-specific view information (eg. tooltips, editor titles, icons) is provided by a stylesheet. Each beamline could have their own stylesheet to specify their own view preferences.
- the mapping between color states and beam status states happens through new auxiliary classes: AMControlToolButtonColorState and BioXASBeamStatusState. These classes provide additional flexibility when defining beam status states. Controls can have multiple beam off/on states, for example.

Some of these ideas could eventually work their way into the existing CLSBeamlineStatus family of classes, if there's interest.


helfrij:

source/ui/beamline/AMControlToolButton.h
```diff
@@ -0,0 +1,83 @@
...
+	AMControlToolButtonColorState(AMToolButton::ColorState colorState, double controlValue) {
+		colorState_ = colorState;
+		controlMinValue_ = controlValue;
+		controlMaxValue_ = controlValue;
+	}
+
+	/// Destructor.
+	virtual ~AMControlToolButtonColorState() {}
+
+	/// The color state.
```
XuetaoLiu-CLS: it will be better to put those properties protected.

source/ui/beamline/AMControlToolButton.cpp
```diff
@@ -0,0 +1,93 @@
...
+		double controlValue = control_->value();
+
+		QList<AMToolButton::ColorState> colorStateMatches;
+
+		// Iterate through the list of color states, identifying
+		// states where the control value falls between the state min
+		// and max values.
+
+		for (int i = 0, count = colorStates_.count(); i < count; i++) {
+			if (colorStates_.at(i).controlMinValue_ <= controlValue && controlValue <= colorStates_.at(i).controlMaxValue_)
```
XuetaoLiu-CLS: maybe we can have a helper function to check whether a given value matches the color. 
```
bool AMControlToolButtonColorState::matchingColorState(double value) 
{
     return controlMinValue_ <= controlValue && controlValue <= controlMaxValue_;
}
```

source/ui/beamline/AMControlToolButton.h
```diff
@@ -0,0 +1,83 @@
...
+#define AMCONTROLTOOLBUTTON_H
+
+#include "beamline/AMControl.h"
+#include "ui/AMToolButton.h"
+
+class AMControlToolButtonColorState
+{
+public:
+	/// Constructor.
+	AMControlToolButtonColorState(AMToolButton::ColorState colorState, double controlMinValue, double controlMaxValue) {
```
XuetaoLiu-CLS: :+1:

source/ui/CLS/CLSValueEditor.cpp
```diff
@@ -227,7 +227,7 @@ void CLSValueEditor::updateEditAction()
 	if (!readOnly_)
 		enabled = true;
 
-	editAction_->setEnabled(true);
+	editAction_->setEnabled(enabled);
 }
```
XuetaoLiu-CLS: :fireworks: good catch!

source/ui/CLS/CLSButtonBar.cpp
```diff
@@ -81,23 +81,31 @@ void CLSButtonBar::removeButton(QAbstractButton *button)
 	}
 }
 
+void CLSButtonBar::removeButtons()
```
XuetaoLiu-CLS: this function is alike the `clearButtons()` except the deleting the button instance.

I think they are doing the same thing :question:

source/ui/BioXAS/BioXASWigglerView.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+void BioXASWigglerView::setWiggler(BioXASWiggler *newWiggler)
+{
+	if (wiggler_ != newWiggler) {
+		wiggler_ = newWiggler;
+		emit wigglerChanged(wiggler_);
+	}
+
+	updateGapEditor();
+	updateGapStatusEditor();
```
XuetaoLiu-CLS: I think we only need to update the editors when the wiggler is changed. Therefore, those two functions might be better to be called inside the `if` branch. 

Also, do we need to listen to the signals (`connected, valueChanged`) of the wiggler and update the editor?

XuetaoLiu-CLS:
The `CLSBeamlineStatusButtonBar/CLSBeamlineStatus/CLSBeamlineStatusView` was derived from the original `BioXASBeamlineStatusButtonBar/BioXASBeamlineStatus/BioXASBeamlineStatusView` to try to have a generic UI/controls for the beam status.

The newer version of `BioXASBeamlineStatusButtonBar/BioXASBeamlineStatus/BioXASBeamlineStatusView` in this PR is similar to the original version and has some similar features with the CLS version. I still think it is a good practice to try to use the same features across beamlines.

XuetaoLiu-CLS:
other than some small questions, it looks good.

The usage of generic BeamStatus and BeamStatusView can be a future topic. This PR is doing what it said.

Time Estimate: 20 minutes

source/ui/CLS/CLSButtonBar.cpp
helfrij: It depends on how the buttons are used. Because another widget could add buttons to the button bar and still keep references to those buttons, there would be a problem when those buttons are deleted. In that case, the widget would want to remove buttons instead of clearing them. It doesn't hurt to provide both functionalities. That's what I was thinking, anyways.

source/ui/BioXAS/BioXASWigglerView.cpp
helfrij: The editors need to be updated as part of the first call to `setWiggler(...)` in the constructor. They need to be updated even if the wiggler instance hasn't changed.

source/icons/icons.qrc
```diff
@@ -221,6 +221,7 @@
         <file>mono-icon5.png</file>
         <file>beam-icon.png</file>
         <file>mirror-icon1.png</file>
+        <file>arrow_wiggle.png</file>
```
dretrex: A wiggle icon for a wiggler!

helfrij: Some of these icons are a bit of a stretch 😄

dretrex:
Overall, this looks good!  However, I feel that this is going to **majorly** conflict with what @XuetaoLiu-CLS has been working on.  We are going to need to be very careful about those two merges.

Time Estimate: 10 minutes

XuetaoLiu-CLS:
Looks good. Lots of codes added .

Time Estimate: 20 minutes

helfrij:
Merge conflicts are resolved, and the BioXAS beam status appears to do what it's supposed to. Thanks @dretrex @XuetaoLiu-CLS!

PR #2433, opened 2016-07-06
Pull Request for Issue 1886: Refactor periodic table and spectrum view classes.

mButlerDM:
Refactor AMScanViewSingleSpectrumView in AMScanViewUtilities with AMXRFBaseDetectorview and AMXRFDetailedDetectorView. Duplicate codes will merged into a new parent class AMSpectrumAndPeriodicTableView.

### DIFF
--- a/source/ui/dataman/AMScanViewUtilities.cpp
+++ b/source/ui/dataman/AMScanViewUtilities.cpp
@@ -368,16 +368,16 @@ AMScanViewSingleSpectrumView::AMScanViewSingleSpectrumView(QWidget *parent)
 
 	/// NEW FUNCTION
 
-	periodicTable_ = new AMSelectablePeriodicTable(this);
-	periodicTable_->buildPeriodicTable();
-	connect(periodicTable_, SIGNAL(elementSelected(AMElement*)), this, SLOT(onElementSelected(AMElement*)));
-	connect(periodicTable_, SIGNAL(elementDeselected(AMElement*)), this, SLOT(onElementDeselected(AMElement*)));
-	periodicTableView_ = new AMSelectablePeriodicTableView(periodicTable_);
-	periodicTableView_->buildPeriodicTableView();
-	connect(periodicTableView_, SIGNAL(elementSelected(AMElement*)), this, SLOT(onElementClicked(AMElement*)));
+//	periodicTable_ = new AMSelectablePeriodicTable(this);
+//	periodicTable_->buildPeriodicTable();
+//	connect(periodicTable_, SIGNAL(elementSelected(AMElement*)), this, SLOT(onElementSelected(AMElement*)));
+//	connect(periodicTable_, SIGNAL(elementDeselected(AMElement*)), this, SLOT(onElementDeselected(AMElement*)));
+//	periodicTableView_ = new AMSelectablePeriodicTableView(periodicTable_);
+//	periodicTableView_->buildPeriodicTableView();
+//	connect(periodicTableView_, SIGNAL(elementSelected(AMElement*)), this, SLOT(onElementClicked(AMElement*)));
 
-	currentElement_ = periodicTable_->elementBySymbol("Fe");
-	combinationElement_ = periodicTable_->elementBySymbol("Ca");
+//	currentElement_ = periodicTable_->elementBySymbol("Fe");
+//	combinationElement_ = periodicTable_->elementBySymbol("Ca");
 
 	////////////////////////////////////////////////////////////
 
@@ -397,9 +397,9 @@ AMScanViewSingleSpectrumView::AMScanViewSingleSpectrumView(QWidget *parent)
 	plotLayout->addLayout(rowAbovePeriodicTableLayout_);
 	plotLayout->addWidget(periodicTableView_, 0, Qt::AlignCenter);
 
-	emissionLineValidator_ = new AMNameAndRangeValidator(this);
-	pileUpPeakValidator_ = new AMNameAndRangeValidator(this);
-	combinationPileUpPeakValidator_ = new AMNameAndRangeValidator(this);
+//	emissionLineValidator_ = new AMNameAndRangeValidator(this);
+//	pileUpPeakValidator_ = new AMNameAndRangeValidator(this);
+//	combinationPileUpPeakValidator_ = new AMNameAndRangeValidator(this);
 
 	sourceButtonsLayout_ = new QVBoxLayout;
 	sourceButtonsLayout_->addWidget(new QLabel("Available Spectra"), 0, Qt::AlignLeft);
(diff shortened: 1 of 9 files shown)


mButlerDM:
@iainworkman Could you review when you have a moment? Since this one was a bit open ended I suspect there's improvements to be made! Thanks!

source/ui/dataman/AMScanViewUtilities.cpp
```diff
@@ -766,41 +432,39 @@ void AMScanViewSingleSpectrumView::onDataPositionChanged(const AMnDIndex &index)
...
 {
 	if (isVisible()){
-
 		addMultipleSpectra_ = true;
 		updatePlot(start, end);
 	}
 }
 
-void AMScanViewSingleSpectrumView::setAxisInfo(AMAxisInfo info, bool propogateToPlotRange)
+void AMScanViewSingleSpectrumView::onCheckBoxChanged(int id)
```
iainworkman: I feel like the name of this function could be more descriptive. What do the check boxes represent?

source/ui/dataman/AMSpectrumAndPeriodicTableView.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+ AMSpectrumAndPeriodicTableView::~AMSpectrumAndPeriodicTableView(){}
+AMSpectrumAndPeriodicTableView::AMSpectrumAndPeriodicTableView(QWidget *parent)
+	: QWidget(parent)
+{
+	emissionLineValidator_ = new AMNameAndRangeValidator(this);
+	pileUpPeakValidator_ = new AMNameAndRangeValidator(this);
+	combinationPileUpPeakValidator_ = new AMNameAndRangeValidator(this);
+
+	periodicTable_ = new AMSelectablePeriodicTable(this);
```
iainworkman: Is there a reason why the periodic table and periodic table view both follow this paradigm of call constructor then call a build function? 
Any idea @dretrex?

source/ui/dataman/AMSpectrumAndPeriodicTableView.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+void AMSpectrumAndPeriodicTableView::buildEnergyRangeSpinBoxView()
+{
+	minimum_ = new QDoubleSpinBox;
+	minimum_->setSuffix(" eV");
+	minimum_->setDecimals(0);
+	minimum_->setRange(0, 1000000);
+	connect(minimum_, SIGNAL(editingFinished()), this, SLOT(onMinimumEnergyChanged()));
+
+	maximum_ = new QDoubleSpinBox;
```
iainworkman: Could these be renamed to
maximumEnergySpinBox and minimumEnergySpinBox? Just makes it a little clearer that they are ui elements and not values.

iainworkman:
Couple of questions/comments. Otherwise looks great.
Time Estimate: 25 minutes.

mButlerDM:
Made the suggested renames.

source/ui/dataman/AMSpectrumAndPeriodicTableView.cpp
dretrex: Because there are virtual functions which build the objects and virtual functions are not properly referenced to their appropriate class until after the object has been constructed.

source/ui/dataman/AMScanViewUtilities.cpp
```diff
@@ -766,41 +432,39 @@ void AMScanViewSingleSpectrumView::onDataPositionChanged(const AMnDIndex &index)
...
-	}
+	AMAxisInfo info = sources_.first()->axisInfoAt(sources_.first()->rank()-1);
 
+	if (info.units.isEmpty())
+		plot_->axisBottom()->setAxisName(info.name);
 	else
-		plot_->plot()->removeItem(series_.at(id));
+		plot_->axisBottom()->setAxisName(info.name % ", " % info.units);
+
+	x_.resize(info.size);
```
dretrex: Not the biggest deal, but I'd recommend not using resize but just making a new vector of the appropriate size with `x_ = QVector<double>(info.size);`.

source/ui/dataman/AMSpectrumAndPeriodicTableView.cpp
```diff
@@ -0,0 +1,388 @@
...
+}
+
+void AMSpectrumAndPeriodicTableView::setAxisInfo(AMAxisInfo info, bool propogateToPlotRange)
+{
+	if (info.units.isEmpty())
+		plotView_->plot()->axisBottom()->setAxisName(info.name);
+	else
+		plotView_->plot()->axisBottom()->setAxisName(info.name % ", " % info.units);
+
+	x_.resize(info.size);
```
dretrex: Same here.

source/ui/dataman/AMSpectrumAndPeriodicTableView.h
```diff
@@ -0,0 +1,194 @@
...
+	Q_OBJECT
+
+public:
+	AMSpectrumAndPeriodicTableView(QWidget *parent = 0);
+
+	virtual ~AMSpectrumAndPeriodicTableView();
+
+	/// Energy Settings
+	/// Returns the energy range used for filtering.  If no range has been provided the range is null.
+	const AMRange &energyRange() const { return emissionLineValidator_->range(); }
```
dretrex: This should just be an AMRange and not a `const AMRange &`.  It's different when passing in classes than returning.  Passing in variables with it means your are passing a non-modifiable reference, which is very fast.  Returning one requires that reference being valid and if anything happens to that memory you will get a seg fault crash.  This is why return values will create a copy and return that (i.e.: `AMRange energyRange() const;`).

source/ui/dataman/AMSpectrumAndPeriodicTableView.h
```diff
@@ -0,0 +1,194 @@
...
+	void setMaximumEnergy(double newMaximum);
+
+	/// Sets the scale for each point along the x-axis. This also calls setPlotRange to make the ranges match. Set \param propogateToPlotRange to false if you don't want the information to propogate.
+	void setAxisInfo(AMAxisInfo info, bool propogateToPlotRange);
+
+	/// Emission Lines
+	/// Returns the list of emission line name filters that have provided to the element view.
+	virtual QList<QRegExp> emissionLineNameFilters() const { return emissionLineValidator_->nameFilters(); }
+	/// Returns the emission line name filter at a given index.
+	virtual const QRegExp &emissionLineNameFilterAt(int index) const { return emissionLineValidator_->nameFilterAt(index); }
```
dretrex: Same here.

dretrex: There are a couple places where this is the case.

dretrex:
Couple of comments.  Looks awesome otherwise!  Great refactor!
Time Estimate: 10 minutes

mButlerDM:
@dretrex Made the suggested changes.

source/ui/dataman/AMScanViewUtilities.cpp
```diff
@@ -358,7 +358,7 @@ AMScanViewSingleSpectrumView::AMScanViewSingleSpectrumView(QWidget *parent)
 	title_ = new QLabel;
 	title_->setFont(newFont);
 
-	x_.resize(0);
+	x_ = QVector<double>(0);
```
dretrex: This one could also be `QVector<double>();`.

mButlerDM:
Modified the empty vector assignment.

PR #2438, opened 2016-07-08
PullRequest for Issue2252: pause workflow action runnner no matter which view we are using now

XuetaoLiu-S:
#2252

issue: when pausing the scan action runner, the workflow action runner is till running, which will cause the workflow action runner stuck.

solution: we will always pause the workflow action runner no matter which action view we are using


XuetaoLiu-CLS:
waiting for review when Issue1628 is merged

XuetaoLiu-CLS:
We have same issue when cancelling/pausing from the current action runner, which will put the workflow action runner stuck forever.

XuetaoLiu-S:
close this branch since we have a better solutions

PR #2440, opened 2016-07-12
Pull Request for Issue2434: IDEAS: Change over to new sample stage motors 

epengr:
Changed two PVs.

Also removed an old, crude hack that is no longer required.

### DIFF
--- a/source/beamline/IDEAS/IDEASMonochromatorControl.cpp
+++ b/source/beamline/IDEAS/IDEASMonochromatorControl.cpp
@@ -42,8 +42,8 @@ void IDEASMonochromatorControl::onAllControlsConnected()
 AMControl::FailureExplanation IDEASMonochromatorControl::move(double setpoint)
 {
 
-	if (setpoint > IDEASBeamline::ideas()->monoHighEV()->value() + 824)  //HACK for max energy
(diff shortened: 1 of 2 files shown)


source/beamline/IDEAS/IDEASMonochromatorControl.cpp
```diff
@@ -42,8 +42,8 @@ void IDEASMonochromatorControl::onAllControlsConnected()
 AMControl::FailureExplanation IDEASMonochromatorControl::move(double setpoint)
 {
 
-	if (setpoint > IDEASBeamline::ideas()->monoHighEV()->value() + 824)  //HACK for max energy
-		setpoint = IDEASBeamline::ideas()->monoHighEV()->value() + 824;
+	if (setpoint > IDEASBeamline::ideas()->monoHighEV()->value())
```
dretrex: You don't need the hack anymore?

epengr: Nope!

Tonia made the HighEV and LowEV PVs writable, so I can change them to reflect reality instead of adding a correction to an arbitrary number.

dretrex: My god!

dretrex:
Time Estimate: 1 minute

Issue #2443, opened 2016-07-14
PGM XAS Scan configuration view


XuetaoLiu-CLS:
There are three issues with PGM XAS Scan configuration view:
- [x] There is no estimation information as the other beamlines 
- [x] There should be a label "Scan name" on the left of the scan name edit box.
- [x] The edit box is too big. Along with the estimation information label, they might be able to nest in the left layout.

mButlerDM:
Time Estimate: 1 Day

PR #2463, opened 2016-07-15
Pull Request for Issue 2443: Adjustments to PGM XAS scan configuration view.

mButlerDM:
Added more spacing to the sides of the view.

Adjusted the name filed so it was nested into the left layout.

Added a label specifying the name field was for the name.

Added an estimation per scan field underneath the name field as it is on other beamlines.

#2443

### DIFF
--- a/source/ui/PGM/PGMXASScanConfigurationView.cpp
+++ b/source/ui/PGM/PGMXASScanConfigurationView.cpp
@@ -13,14 +13,14 @@ PGMXASScanConfigurationView::PGMXASScanConfigurationView(PGMXASScanConfiguration
 
 	// Scan Name
 
-	QLabel *scanNameLable = new QLabel("Scan Name: ");
+	QLabel *scanNameLabel = new QLabel("Scan Name: ");
 
 	scanName_ = new QLineEdit();
 	scanName_->setText(configuration_->name());
 	scanName_->setAlignment(Qt::AlignCenter);
 
 	QHBoxLayout *scanNameLayout = new QHBoxLayout;
-	scanNameLayout->addWidget(scanNameLable);
+	scanNameLayout->addWidget(scanNameLabel);
 	scanNameLayout->addWidget(scanName_);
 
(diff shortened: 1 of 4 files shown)


mButlerDM:
@XuetaoLiu-CLS Could you review when you have a chance?

XuetaoLiu-CLS:
Looks good. Time Estimate: 5 minutes

source/ui/PGM/PGMXASScanConfigurationView.cpp
```diff
@@ -11,14 +11,37 @@ PGMXASScanConfigurationView::PGMXASScanConfigurationView(PGMXASScanConfiguration
 {
 	configuration_ = configuration;
 
+	// Scan Name
+
+	QLabel *scanNameLable = new QLabel("Scan Name: ");
```
dretrex: scanName**_Label_**

dretrex:
If you change the number of iterations, does the time in the configuration view holder update accordingly?

dretrex:
Only a couple comments/questions.  Looks good otherwise.

Time Estimate: 3 minutes

source/ui/PGM/PGMXASScanConfigurationView.cpp
XuetaoLiu-CLS: :bomb: good catch

XuetaoLiu-CLS:
@dretrex I think the estimate here is only about a single scan. We didn't do the estimation for the iterated scans in the other beamlines as well.

dretrex:
Since the scan configuration view holder displays the estimated time, we should connect it in this pull request as well.  At the moment, you should only need to emit a signal from AMScanConfiguration and it should do the rest.

mButlerDM:
Seems reasonable I will look into making that change.

mButlerDM:
@dretrex Got that change in, let me know if that is what you had in mind.

dretrex:
Looks good!

Issue #2464, opened 2016-07-18
Broken connect statement


helfrij:
There's a minor error related to AMSpectrumAndPeriodicTableView reported when I build BioXAS applications, that there is no slot called `onExportClicked()`. There is a connect statement on line 76 of that class that could be broken as a result. Not sure if this impairs any functionality, maybe the connect statement just needs to be removed?

mButlerDM:
Time Estimate: 4 Hours

PR #2465, opened 2016-07-18
Pull Request for Issue 2464: onExportClicked not defined in AMSpectrumAndPeriodicTableView

mButlerDM:
Fixed a few issues with connections in AMSpectrumAndPeriodicTableView

onExportClicked() was connected in parent when it needed to be connected in the child classes.
Renamed onExportClicked and onSaveButtonClicked to onExportButtonClicked so child classes are consistent with each other.

Moved some connections that were made in the parent to the child so they would properly use the child's version of the slot function.

#2464

### DIFF
--- a/source/ui/beamline/AMXRFDetailedDetectorView.cpp
+++ b/source/ui/beamline/AMXRFDetailedDetectorView.cpp
@@ -172,8 +172,6 @@ void AMXRFDetailedDetectorView::buildEnergyRangeSpinBoxView()
 {
 	AMSpectrumAndPeriodicTableView::buildEnergyRangeSpinBoxView();
 
-	connect(exportButton_, SIGNAL(clicked()), this, SLOT(onExportButtonClicked()));
-
 	showEnergyRangeSpinBoxes_ = new QPushButton(QIcon(":/system-run.png"), "Settings");
 	showEnergyRangeSpinBoxes_->setCheckable(true);
 
--- a/source/ui/beamline/AMXRFDetailedDetectorView.cpp
+++ b/source/ui/beamline/AMXRFDetailedDetectorView.cpp
@@ -63,7 +63,6 @@ AMXRFDetailedDetectorView::AMXRFDetailedDetectorView(AMXRFDetector *detector, QW
 	// this is to make sure we have more columns than rows, which will looks nicer
 	deadTimeViewFactor_ = (detector->elements() / deadTimeViewFactor_ ) > deadTimeViewFactor_ ? deadTimeViewFactor_ + 1 : deadTimeViewFactor_;
 
-
 	connect(periodicTable_, SIGNAL(elementSelected(AMElement*)), this, SLOT(onElementSelected(AMElement*)));
 	connect(periodicTable_, SIGNAL(elementDeselected(AMElement*)), this, SLOT(onElementDeselected(AMElement*)));
 	connect(periodicTableView_, SIGNAL(elementSelected(AMElement*)), this, SLOT(onElementClicked(AMElement*)));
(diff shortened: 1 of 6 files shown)


mButlerDM:
@helfrij Could you review when you get a moment? Thanks!

source/ui/beamline/AMXRFDetailedDetectorView.cpp
```diff
@@ -167,6 +172,8 @@ void AMXRFDetailedDetectorView::buildEnergyRangeSpinBoxView()
 {
 	AMSpectrumAndPeriodicTableView::buildEnergyRangeSpinBoxView();
 
+	connect(exportButton_, SIGNAL(clicked()), this, SLOT(onExportButtonClicked()));
```
helfrij: Should this connect statement be in the constructor with the others? Or is it better to keep it here?

helfrij:
Looks good @mButlerDM!

Time Estimate: 5 minutes.

source/ui/beamline/AMXRFDetailedDetectorView.cpp
XuetaoLiu-CLS: @mButlerDM @helfrij I think it might be better to be where we created the button. In this case it is the function `AMSpectrumAndPeriodicTableView::buildEnergyRangeSpinBoxView();`.

@mButlerDM if `onExportButtonClicked()` must be implemented in the Detailed detector view, you can make it a (pure) virtual slot in `AMSpectrumAndPeriodicTableView`.

mButlerDM:
@helfrij  I think it makes more sense to have it somewhere else. I wasn't too what the best way to do that was. I followed @XuetaoLiu-CLS suggestion and used a pure virtual function declaration to keep it in the parent.

source/ui/beamline/AMXRFDetailedDetectorView.h
```diff
@@ -48,13 +48,6 @@ class AMXRFDetailedDetectorView : public AMXRFBaseDetectorView
 	/// Re-implementing but still going to use the base class buildDetectorView since this view is merely adding to it.
 	virtual void buildDetectorView();
 
-	/// Returns the energy range used for filtering.  If no range has been provided the range is null.
-	const AMRange &energyRange() const { return emissionLineValidator_->range(); }
```
dretrex: Why were these removed?  Were they moved somewhere else and then not removed initially?

source/ui/dataman/AMSpectrumAndPeriodicTableView.h
```diff
@@ -156,6 +156,8 @@ protected slots:
 
 	/// The export button.
 	QPushButton *exportButton_;
+	///
```
dretrex: Why is this function buried deep within the protected members when it is supposed to be a slot?

dretrex: Also, you need to be careful using virtual functions in constructors (I'm not sure about using them in connect statements).  In general, the use of virtual functions in constructors should be avoided because the virtual function won't be properly resolved using polymorphism yet.

dretrex:
Time Estimate: 3 minutes

source/ui/beamline/AMXRFDetailedDetectorView.h
mButlerDM: Yes they were supposed to be removed in the initial refactor, as I went over all the functions again I noticed they were defined in both AMSpectrumAndPeriodicTableView and here.

source/ui/dataman/AMSpectrumAndPeriodicTableView.h
mButlerDM: @dretrex Probably because I saw `exportButton_` and got excited, oops!

The connection to the virtual functions occurs in `buildEnergyRangeSpinBoxView()` which is called outside the constructor and after initialization in `AMXRFDetailedDetectorView`, though it is called in the constructor of `AMScanViewUtilities`. It might be worth refactoring the `AMScanViewUtilities` constructor to be more in line with AMXRF?

Issue #2460, opened 2016-07-15
Plugin path for PGM is not correct


dretrex:
You cannot reopen scans on PGM.  This is likely due to the plugin path not being correct.  @XuetaoLiu-CLS, can you take a look at this since you have already setup other beamlines to run out of the acquaman account?

XuetaoLiu-CLS:
Time Estimate: 10 minutes

PR #2472, opened 2016-07-19
PullRequest for Issue2460: set the default plugin path based on the current running path

XuetaoLiu-CLS:
#2460

### DIFF
--- a/source/util/AMSettings.cpp
+++ b/source/util/AMSettings.cpp
@@ -172,12 +172,12 @@ void AMSettings::load() {
 	publicDataFolder_ = settings.value("publicDataFolder", "/home/acquaman/data/").toString();
 	publicDatabaseFilename_ = settings.value("publicDatabaseFilename", "publicdata.db").toString();
 
-	QString applicationExecutablePath(QCoreApplication::applicationDirPath());
-	QDir applicationDir(applicationExecutablePath);
+	QDir applicationDir(QCoreApplication::applicationDirPath());
 	if(!applicationDir.exists()) {
 		AMErrorMon::alert(0, AM_SETTINGS_NONE_EXIST_APPLICATION_PATH, QString("The application path %1 doesn't exist!").arg(applicationDir.absolutePath()));
 	} else {
 		applicationDir.cdUp();
+
 		QDir fileLoderPluginDir(applicationDir.absolutePath() % "/plugins/FileLoaders");
 		if (!fileLoderPluginDir.exists()) {
 			AMErrorMon::alert(0, AM_SETTINGS_NONE_EXIST_PLUGIN_PATH, QString("The plugin path %1 doesn't exist!").arg(fileLoderPluginDir.absolutePath()));
--- a/source/util/AMSettings.cpp
+++ b/source/util/AMSettings.cpp
@@ -21,6 +21,7 @@ along with Acquaman.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "AMSettings.h"
 
+#include <QCoreApplication>
 #include <QDir>
 #include <QDateTime>
 #include <QStringBuilder>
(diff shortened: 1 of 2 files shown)


XuetaoLiu-CLS:
@helfrij can you review? Thx

source/util/AMSettings.cpp
```diff
@@ -171,9 +170,9 @@ void AMSettings::load() {
 	publicDataFolder_ = settings.value("publicDataFolder", "/home/acquaman/data/").toString();
 	publicDatabaseFilename_ = settings.value("publicDatabaseFilename", "publicdata.db").toString();
 
-	QString defaultBasePath(QDir::homePath() % "/beamline/programming");
-	fileLoaderPluginsFolder_ = settings.value("fileLoaderPluginsFolder", QString(defaultBasePath % "/acquaman/plugins/FileLoaders")).toString();
-	analysisBlockPluginsFolder_ = settings.value("analysisBlockPluginsFolder", QString(defaultBasePath % "/acquaman/plugins/AnalysisBlocks")).toString();
+	QString defaultBasePath(QCoreApplication::applicationDirPath() % "/..");
```
helfrij: Does the `QCoreApplication::applicationDirPath()` include the /acquaman directory component used in the `fileLoaderPluginsFolder_` and the `analysisBlockPluginsFolder_` below?

helfrij:
Looks good, @XuetaoLiu-CLS. Just a clarification question.

Time Estimate: 5 minutes.

source/util/AMSettings.cpp
XuetaoLiu-CLS: @helfrij for SGM/PGM/SXRMB, acquaman is running from `acquaman/bin`; if there is no exception, the other beamlines should run acquaman from `acquaman/build`. Therefore, `QCoreApplication::applicationDirPath()` includes either `acquaman/bin` or `acquaman/build`. That's why we also need to add `/..` for the plugin folder

dretrex: Rather than just adding in `/..` we could throw the string from `QCoreApplication::applicationDirPath()` into a QFileInfo, then use `cdUp()` (synonymous with with `cd ..` then check the string is correct.  Thoughts?

dretrex:
Quick question.  Otherwise looks good.
Time Estimate: 2 minutes

source/util/AMSettings.cpp
XuetaoLiu-CLS: @dretrex Both solutions are based on the assumption that the plugin folder is in the same super folder of the executable ones. The only difference is that with QFileInfo, we can check whether it is a valid path or not; when we need to handle the case when the path is invalid.

I can use QFileInfo for sure.

dretrex: Agreed.

XuetaoLiu-CLS:
@dretrex it is changed. I use QDir now to check the directory.

source/util/AMSettings.cpp
```diff
@@ -170,10 +172,25 @@ void AMSettings::load() {
 	publicDataFolder_ = settings.value("publicDataFolder", "/home/acquaman/data/").toString();
 	publicDatabaseFilename_ = settings.value("publicDatabaseFilename", "publicdata.db").toString();
 
-	QString defaultBasePath(QCoreApplication::applicationDirPath() % "/..");
-	fileLoaderPluginsFolder_ = settings.value("fileLoaderPluginsFolder", QString(defaultBasePath % "/plugins/FileLoaders")).toString();
-	analysisBlockPluginsFolder_ = settings.value("analysisBlockPluginsFolder", QString(defaultBasePath % "/plugins/AnalysisBlocks")).toString();
+	QDir applicationDir(QCoreApplication::applicationDirPath());
+	if(!applicationDir.exists()) {
+		AMErrorMon::alert(0, AM_SETTINGS_NONE_EXIST_APPLICATION_PATH, QString("The application path %1 doesn't exist!").arg(applicationDir.absolutePath()));
+	} else {
```
dretrex: 👍

Issue #2457, opened 2016-07-15
Auto-export is pointing to local folder rather than remote


dretrex:
While showing Lucia how scans work I noticed that exported data is going to AcquamanLocalData rather than the remote location we setup.  This needs to be fixed.

mButlerDM:
@dretrex This may be due to a file system permission error.

dretrex:
Interesting.  Thanks, @mButlerDM!

XuetaoLiu-CLS:
This happens because we overwrite the remote data folder with the local data folder after the first time wizard. As a result, in the Acquaman.ini, both the remote and local path are the same. 

Time Estimate: 1 hours

PR #2478, opened 2016-07-19
PullRequest for issue2457 when using local data storage, we won't need to change the settting again

XuetaoLiu-CLS:
when using local data storage (non user based storage), the users are NOT allowed to choose the storage path in the first time wizard. Therefore, there is no need to rewrite the data folder settings after the wizard page is dismissed. 

#2457

### DIFF
--- a/source/application/AMDatamanAppController.cpp
+++ b/source/application/AMDatamanAppController.cpp
@@ -387,21 +387,24 @@ bool AMDatamanAppController::startupOnFirstTime()
 
 		firstTimeWizard->deleteLater();
 
-		if(userLocalStorage){
-			AMUserSettings::remoteDataFolder = userDataFolder;
-			if(!AMUserSettings::remoteDataFolder.endsWith('/'))
-				AMUserSettings::remoteDataFolder.append("/");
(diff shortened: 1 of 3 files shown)


XuetaoLiu-CLS:

source/application/AMDatamanAppController.cpp
```diff
@@ -387,22 +387,25 @@ bool AMDatamanAppController::startupOnFirstTime()
 
 		firstTimeWizard->deleteLater();
 
-		if(userLocalStorage){
-			AMUserSettings::remoteDataFolder = userDataFolder;
-			if(!AMUserSettings::remoteDataFolder.endsWith('/'))
-				AMUserSettings::remoteDataFolder.append("/");
+		// we only need to change the AMUserSettings if we are doing user based data storage
```
dretrex: What does user based storage mean? The kind of thing that SGM had?

dretrex:
Quick question.
Time Estimate: 2 minutes

source/application/AMDatamanAppController.cpp
XuetaoLiu-CLS: @dretrex yes, that's SGM specific stuff

PR #2499, opened 2016-07-26
Pull Request for Issue 2477: Removing scan region results in crash.

mButlerDM:
Crash resulted when a scan region was removed.

When adding the AMStepScanAxisView to the scan configuration setLayout was used which re-parents the layout, breaking references in the internal map.

Instead AMStepScanAxisView must be added to a view, and that layout set to the view container.


mButlerDM:
@helfrij could you review when you have a moment? Thanks!

helfrij:
Looks good, @mButlerDM!

Time Estimate: 2 minutes.

dretrex:
Great fix.
Time Estimate: 1 minute

PR #2511, opened 2016-07-26
Pull Request for Issue2501: Build AMHDF5File

dretrex:
This encapsulates many of the standard operations you'll need to interact with an HDF5 file.  It provides a couple small convenience functions as well.


mButlerDM:
Time Estimate: 5 Minutes

Looks good @dretrex

Issue #2502, opened 2016-07-26
Build AMHDF5Group


dretrex:

mButlerDM:
Time Estimate: 1 Day

dretrex:
Done in #2512.

PR #2512, opened 2016-07-27
Pull Request for Issue 2502: Build AMHDF5Group

mButlerDM:
Added the AMHDF5Group class which will handle interfacing the HDF5 group functions with AM. More may be required later, for now functionality includes:

isOpen
createHDF5Group
openHDF5Group
closeHDF5Group

#2502


mButlerDM:
@dretrex could you review when you have a moment?

source/dataman/HDF5/AMHDF5Group.cpp
```diff
@@ -0,0 +1,98 @@
...
+AMHDF5Group::AMHDF5Group(const QString &name, QObject *parent)
+	: QObject(parent)
+{
+	name_ = name;
+	groupID_ = 0;
+}
+
+AMHDF5Group::~AMHDF5Group()
+{
+	if(!isOpen())
```
dretrex: Mine has this bug too.  We only want to close it if we are open, therefore:

```c++
if (isOpen())
    closeHDF5Group();
```

source/dataman/HDF5/AMHDF5Group.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+bool AMHDF5Group::isOpen() const
+{
+	return groupID_ > 0;
+}
+
+bool AMHDF5Group::createHDF5Group(hid_t file)
+{
+	if (groupID_ != 0){
+		AMErrorMon::alert(this, 0, QString("Create: Group %1 is already open.").arg(name_));
```
dretrex: Lets add error codes for these error monitors.  Something similar to what is in AMHDF5File.h should be fine.

source/dataman/HDF5/AMHDF5Group.cpp
```diff
@@ -0,0 +1,98 @@
...
+		return false;
+	}
+
+	if(!file > 0){
+		AMErrorMon::alert(this, 0, QString("Create: Cannot create group %1 in a closed file.").arg(name_));
+		return false;
+	}
+
+	hid_t grpID = 0;
```
dretrex: Lets use full words:  `groupId`.

source/dataman/HDF5/AMHDF5Group.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+bool AMHDF5Group::openHDF5Group(hid_t file)
+{
+	if(groupID_ != 0){
+		AMErrorMon::alert(this, 0, QString("Open: Group %1 is already open.").arg(name_));
+		return false;
+	}
+
+	if(!file > 0){
```
dretrex: I don't think this does what you're expecting.

source/dataman/HDF5/AMHDF5Group.cpp
```diff
@@ -0,0 +1,98 @@
...
+	if(!isOpen())
+		closeHDF5Group();
+}
+
+bool AMHDF5Group::isOpen() const
+{
+	return groupID_ > 0;
+}
+
+bool AMHDF5Group::createHDF5Group(hid_t file)
```
dretrex: Since these are all IDs, lets make that explicit `bool AMHDF5Group::createHDF5Group(hid_t fileId);`

source/dataman/HDF5/AMHDF5Group.cpp
```diff
@@ -0,0 +1,98 @@
...
+		AMErrorMon::alert(this, 0, QString("Open: Group %1 is already open.").arg(name_));
+		return false;
+	}
+
+	if(!file > 0){
+		AMErrorMon::alert(this, 0, QString("Open: Group %1 cannot be opened from a closed file!").arg(name_));
+		return false;
+	}
+
+	hid_t grpID = 0;
```
dretrex: You can merge this line and the next line.

source/dataman/HDF5/AMHDF5Group.h
```diff
@@ -0,0 +1,46 @@
...
+
+signals:
+
+public slots:
+	/// Create the HDF5 group for a specified file and associates that group handle with this object.
+	bool createHDF5Group(hid_t file);
+	/// Opens an existing HDF5 group for access and associates the handle with this object.
+	bool openHDF5Group(hid_t file);
+	/// Closes a currently open HDF5 group and resets the group handle ID.
+	bool closeHDF5Group();
```
dretrex: You are missing flush.

dretrex:
You'll want to add to `AMHDF5File` as well.  Something like `bool addGroup(const QString &name);` and some other things like getters and what not.

dretrex:
Time Estimate: 15 minutes

source/dataman/HDF5/AMHDF5Group.h
mButlerDM: There's no support to flush a group in and of itself. To flush I would need to do it with the file reference which seemed a bit redundant?

dretrex: `herr_t H5Gflush(hid_t group_id)`

source/dataman/HDF5/AMHDF5File.cpp
```diff
@@ -169,3 +170,21 @@ bool AMHDF5File::flushHDF5File()
 	return true;
 }
 
+bool AMHDF5File::addGroup(const QString &groupName)
+{
+	AMHDF5Group* group = new AMHDF5Group(groupName);
+	if(group->createHDF5Group(id_))
```
dretrex: I'd redo these to be more like the following:

```c++
if (!group->createHDF5Group(id_)){
    group->deleteLater();
    AMErrorMon::alert();
    return false;
}

groups_ << group;

return true;
```

And the same would go with open.  You will also want to add a means of closing and getting those groups from the AMHDF5File class as well.

mButlerDM:
@dretrex Made the suggested changes. This should be more in line with what we needed.

source/dataman/HDF5/AMHDF5File.h
```diff
@@ -14,6 +17,8 @@
 #define AMHDF5FILE_INVALID_OPEN_OPTION 666004
 #define AMHDF5FILE_FLUSH_FILE_NOT_OPEN 666005
 
+class AMHDF5Group;
```
dretrex: This is unnecessary if you include it like you did.

source/dataman/HDF5/AMHDF5File.cpp
```diff
@@ -172,19 +171,59 @@ bool AMHDF5File::flushHDF5File()
...
+
+		groups_.insert(groupName, groupToOpen);
+
 		return true;
 	}
+
+	return false;
+}
+
+AMHDF5Group* AMHDF5File::findOpenGroup(const QString &groupName)
```
dretrex: This is a `const` function.

source/dataman/HDF5/AMHDF5File.h
```diff
@@ -63,16 +68,24 @@ public slots:
...
+
+	/// Adds a group to the current file. Returns reference to the group if successful.
 	bool addGroup(const QString &groupName);
-	/// Opens the associated group for the file.
+	/// Opens the associated group for the file. Returns reference to the group if successful.
 	bool openGroup(const QString &groupName);
+	/// Closes the associated group for this file.
+	bool closeGroup(const QString &groupName);
+	/// Returns an open group with a specified name if it exists.
+	AMHDF5Group* findOpenGroup(const QString &groupName);
```
dretrex: This should be a public function, not a public slot.

dretrex:
Still missing `H5Gflush`.

source/dataman/HDF5/AMHDF5File.h
mButlerDM: To compile `AMHDF5File` when it only handles references to `AMHDFGroup` required a forward declaration of the class.

mButlerDM:
@dretrex Made the other adjustments, still not sure on the forward declaration so left it as is for now.

source/dataman/HDF5/AMHDF5File.cpp
```diff
@@ -169,3 +169,75 @@ bool AMHDF5File::flushHDF5File()
...
+	}
+
+	return NULL;
+}
+
+bool AMHDF5File::closeGroup(const QString &groupName)
+{
+	if(isOpen() && groups_.contains(groupName)){
+
+		AMHDF5Group* groupToClose = groups_.value(groupName);
```
dretrex: This needs to be removed from the group map as well.

source/dataman/HDF5/AMHDF5File.cpp
```diff
@@ -169,3 +169,75 @@ bool AMHDF5File::flushHDF5File()
...
+
+		if(groupToClose->closeHDF5Group()){
+			groupToClose->deleteLater();
```
dretrex: In here.

mButlerDM:
@dretrex Pushed our changes from yesterday and changed `group_.value()` to `group_.take()` to insure a closed group is removed. 

Still working on that HDF5 file error we say yesterday.

dretrex:
Okay cool.

source/dataman/HDF5/AMHDF5Group.h
```diff
@@ -0,0 +1,55 @@
...
+public:
+
+	/// Constructor. Takes the name of the HDF5 Group. Does not create the group itself.
+	explicit AMHDF5Group(const QString &name, QObject *parent = 0);
+	/// Deconstructor for the group object.
+	virtual ~AMHDF5Group();
+	/// Checks if a group is currently open.
+	bool isOpen() const;
+	/// Returns the group ID currently held by AMHDF5Group.
+	hid_t groupID() const { return groupID_;}
```
dretrex: Not a big deal, but I think we could probably make this just `id()` since it is implicitly about an HDF5 group.

mButlerDM:
@dretrex Changed `groupID()` and you were right it was `>=` and HDF5 will return 0 on a successful close, sometimes.

Issue #2513, opened 2016-07-27
Add new run doesn't work


XuetaoLiu-CLS:
on the latest master branch, we can't add new run. The dialog can't popup

mButlerDM:
Time Estimate: 15 Minutes

PR #2519, opened 2016-07-28
Pull Request for Issue 2513: Add new run doesn't work

mButlerDM:
After changing the run selector to be modal to protect the main AM window, the new run selection did not inherent the modal property.

Due to the way that new run dialog is called it does not inherent the `modal(true)` flag from the parent. Manually set the new run dialog to be modal as well. This seems to resolve the issue.
#2513


mButlerDM:
@helfrij Could you review when you have a moment? Thanks!

helfrij:
Looks good, @mButlerDM.

Time Estimate: 2 minutes.

PR #2523, opened 2016-07-29
Pull Request for Issue2503: Build AMHDF5DataSet

dretrex:
Implemented HDF5 data set API as a class.  Also did some code refactor to AMHDF5Group as well as some general usage refactoring.


source/dataman/HDF5/AMHDF5DataSet.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+		if (!dataSpace->create()){
+
+			dataSpace->deleteLater();
+			return false;
+		}
+	}
+
+	hid_t dataSetId = H5Dcreate2(fileId_, name_.toUtf8().constData(), H5T_NATIVE_DOUBLE, dataSpace->id(), H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
+	dataSpace->close();
```
mButlerDM, line 123: Do we need to catch if close fails?

dretrex, line 123: I don't think so since it is created right above the `H5Dcreate2` method and it catches all the problems there.

source/dataman/HDF5/AMHDF5File.cpp
```diff
@@ -199,44 +220,167 @@ bool AMHDF5File::openGroup(const QString &groupName)
...
 
-	return false;
+	else {
+		AMErrorMon::alert(this, AMHDF5FILE_GROUP_ALREADY_ADDED, QString("Did not open %1 because %2 has already opened it.").arg(groupName).arg(name_));
+		return false;
+	}
 }
 
-AMHDF5Group* AMHDF5File::findOpenGroup(const QString &groupName) const
+AMHDF5Group* AMHDF5File::groupByName(const QString &groupName) const
```
mButlerDM, line 229: This may not matter but this change may not signify clearly enough that it only queries for open groups.

dretrex, line 229: If the group isn't open then it won't find it in here anyway.

source/dataman/HDF5/AMHDF5DataSet.cpp
mButlerDM, line 123: If the dataSpace fails to close at the H5 level and we delete the object is there a risk for data loss? This may be a non-issue for data spaces.

mButlerDM:
Time Estimate: 20 minutes

Looks good, a few minor questions.

source/dataman/HDF5/AMHDF5DataSet.cpp
dretrex, line 123: Potentially, but there is nothing we can really do about that.  We are essentially doing,

``` c++
AMHDF5DataSpace *dataSpace = new AMHDF5DataSpace();
H5Dcreate2();
dataSpace->deleteLater();
```

If something went wrong in H5Dcreate2 then we'd be screwed way worse than not catching the data space failure.

PR #2524, opened 2016-07-29
Pull Request for Issue 2505: Build AMHDF5Error

mButlerDM:
Implementation of AMHDF5Error for #2505.

### DIFF
--- a/source/dataman/HDF5/AMHDF5Error.cpp
+++ b/source/dataman/HDF5/AMHDF5Error.cpp
@@ -1,12 +1,35 @@
 #include "AMHDF5Error.h"
 
+#include <stdio.h>
+#include <wchar.h>
+
+/* This routine prints the information for an error on the error stack
+ * Note that it uses the following H5E_error_t struct, which is updated from
+ * the current H5E_error_t:
+ *
+ * typedef struct H5E_error_t {
+ *    hid_t 	  cls_id;	   Error class ID
+ *    hid_t 	  maj_id;	   Major error ID
+ *    hid_t       min_id;	   Minor error ID
+ *    const char *func_name;	Function in which error occurred
+ *    const char *file_name;	File in which error occurred
+ *    unsigned	  line;		 Line in file where error occurs
+ *    const char *desc;		 Optional supplied description
+ * } H5E_error_t;
+ *
+ */
+
+#include <QDebug>
+
 #define AMHDF5ERROR_ERROR_STACK_NOT_AQUIRED 700000
 #define AMHDF5ERROR_PRINT_FAILED 700001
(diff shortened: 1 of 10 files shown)


mButlerDM:
@dretrex I imagine this can be added `AMHDF5File` but its usefulness right now is limited.

source/dataman/HDF5/AMHDF5Error.cpp
```diff
@@ -0,0 +1,68 @@
...
+bool AMHDF5Error::printErrorStack()
+{
+	hid_t errorStack = H5Eget_current_stack();
+
+	if(errorStack < 0){
+		AMErrorMon::alert(this, AMHDF5ERROR_ERROR_STACK_NOT_AQUIRED, QString("Print: HDF5 error-stack could not be aquired."));
+		return false;
+	}
+
+	herr_t printStatus = H5Eprint2(errorStack, stdout);
```
dretrex, line 30: Is it all possible to be able to provide some kind of `FILE *` that would allow you to get this information into a QString to be added to an error monitor at all?

source/dataman/HDF5/AMHDF5Error.cpp
```diff
@@ -0,0 +1,68 @@
...
+/// This function may be removed down the road if a solution on error stack capturing is not found.
+herr_t captureErrorOutput(unsigned int n, const H5E_error2_t *err_desc, void *client_data)
+{
+//	char majorError[32] = "";
+//	const char minorError[32] = "";
+
+//	H5Eget_major(err_desc->maj_num);
+//	H5Eget_minor(err_desc->min_num);
+
+//	memcpy((char *)client_data, majorError, 32 * sizeof(char));
```
dretrex: This might be worth playing with.  If you got the client data copied into a `char *` buffer you should be able to get that into a string and return it.

source/dataman/HDF5/AMHDF5Error.cpp
mButlerDM, line 30: I think this is possible. A memstream would avoid having to have a file handle but it would take a bit of figuring. I'll keep looking at this option.

source/dataman/HDF5/AMHDF5Error.cpp
mButlerDM: I tried this for my first attempt but had trouble getting memcpy to move data over. I'm guessing it was memory protection for C++ classes. I'll look at this option a bit more closely.

mButlerDM:
@dretrex I got error output to `AMErrorMonitor` working with streams. There is likely a better way to reference the `AMHDF5Error` object/instance between the classes.

Still have a few small things to clean up.

PR #2532, opened 2016-08-02
Pull Request for Issue2530: Hiding Ge detector view

helfrij:
Added some methods to BioXASAppController to use for interacting with the componentView map, added similar methods to CLSAppController for the view-pane map. The Side Ge detector main window representation only uses a view (no pane), so adjusted the BioXASSideAppController code to look only for a view when updating the detector's visibility.

It would be nice to eventually have some way setting a 'configuration show/hide' bool where the Ge detector view can be forced to be shown or hidden, regardless of connected state. This could be useful for testing purposes. But for now, this pull request only deals with the requirement that the view be hidden when detector not connected.


helfrij:

